PR #5, opened 2024-01-18
Use multi-scan-id as env variable

orz25:
Analyzer Manager is expecting to get the multi_scan_id as env variable named JF_MSI. 

PR for creating this env variable right after we get the multi_scan_id from xsc (in jfrog-client-go):
https://github.com/jfrog/jfrog-client-go/pull/883

### DIFF
--- a/utils/analyzermanager.go
+++ b/utils/analyzermanager.go
@@ -91,12 +91,13 @@ func (am *AnalyzerManager) ExecWithOutputFile(configFile, scanCommand, workingDi
 		return
 	}
 	var cmd *exec.Cmd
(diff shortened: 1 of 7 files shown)


utils/analyzermanager.go
```diff
@@ -91,12 +91,13 @@ func (am *AnalyzerManager) ExecWithOutputFile(configFile, scanCommand, workingDi
 		return
 	}
 	var cmd *exec.Cmd
+	multiScanId := os.Getenv("JF_MSI")
```
attiasas, line 94: make sure "JF_MSI" is defined as const.

PR #11, opened 2024-01-25
Scan if config exists but resolver not defined

attiasas:
depends on: https://github.com/jfrog/jfrog-cli-core/pull/1117

image

When transfering the content there was a mistake, resulting that the missing repo error did not identified and an error was thrown that resolver is expected. reverting the content of the method to what it was supposed to be


commands/audit/scarunner.go
```diff
@@ -268,16 +268,27 @@ func SetResolutionRepoIfExists(params xrayutils.AuditParams, tech coreutils.Tech
...
-	if err != nil {
-		err = fmt.Errorf("failed getting server details: %s", err.Error())
-		return
+
+	// If the resolver repository doesn't exist and triggers a MissingResolverErr in ReadResolutionOnlyConfiguration, the repoConfig becomes nil. In this scenario, there is no depsRepo to set, nor is there a necessity to do so.
+	if repoConfig != nil {
+		log.Debug("Using resolver config from", configFilePath)
+		details, e := repoConfig.ServerDetails()
+		if e != nil {
+			err = fmt.Errorf("failed getting server details: %s", e.Error())
```
eranturgeman, line 286: do you need to return after checking this error?

eranturgeman (commented):
.

commands/audit/scarunner.go
attiasas: we are, there is no logic after that so `return` is redundant

PR #13, opened 2024-01-25
Support maven cli for curation cli cmd.

asafambar:
In this PR I added maven support for curation cli command.

In order to support maven, we first need to be able to get dependency graph from maven ecosystem without being blocked from curation service, this can not be done today as 'mvn dependency:graph' (or jfrog depTree plugin) download jars as part of building the dependency graph, this jars can be potentially blocked by curation service. 

In order to overcome this, Artifactory developed a new api that add a flag to curation service on each request, this flag symbol that requested package should not be blocked and also not cached by Artifactory, this will allow curation cli command to build the dep tree successfully, all need to be done is add the new api to the registry url so the package manager will download all the packages through it.

In addition to that because we bypass curation service while building the tree we also added a dedicated cache for curation command, so if any "blocked" package will be download it won't go to the default cache.

Also added to the node graph of maven the type of each node, so when building the download url of each node in the graph it will be clear which type of file it is.

Maven support added silently(depend on env variable), when Artifactory will release official support to the "pass through" api, we will also make it official.  

This PR depends on: https://github.com/jfrog/jfrog-cli-core/pull/1118


config/general.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/jfrog/jfrog-client-go/utils"
+	"os"
+	"path/filepath"
+)
+
+const (
+	JfrogCurationDirName = "curation"
+
+	CurationsDir         = "JFROG_CLI_CURATION_DIR"
+	CurationMavenSupport = "JFROG_CLI_CURATION_MAVEN"
```
attiasas, line 14:
```suggestion
         // #nosec G101 -- Not credentials.
	CurationMavenSupport = "JFROG_CLI_CURATION_MAVEN"
```
To fix the GO-Sec add the following, change the reason if needed.

attiasas (changes requested):
Please add a description to the PR related to the changes.
Don't forget to update the [Documentaion](https://github.com/jfrog/documentation) as well if needed.
Please add a link to the PR in core that is related to this.
make sure all the tests on both PR are passing before I will continue to review this.

commands/curation/curationaudit.go
```diff
@@ -253,11 +271,16 @@ func (ca *CurationAuditCommand) auditTree(tech coreutils.Technology, results map
...
+	rootNodes := map[string]struct{}{}
+	for _, tree := range fullDependenciesTrees {
+		rootNodes[tree.Id] = struct{}{}
+	}
 	// Fetch status for each node from a flatten graph which, has no duplicate nodes.
-	err = analyzer.fetchNodesStatus(flattenGraph, &packagesStatusMap, rootNode.Id)
-	analyzer.fillGraphRelations(rootNode, &packagesStatusMap,
-		&packagesStatus, "", "", datastructures.MakeSet[string](), true)
+	packagesStatusMap := sync.Map{}
+	err = analyzer.fetchNodesStatus(flattenGraph, &packagesStatusMap, rootNodes)
```
attiasas, line 281: if err exists after this line why are we not checking and continue to process until the end?
should we check for error here?

commands/curation/curationaudit.go
```diff
@@ -53,8 +54,11 @@ const (
 
 var CurationOutputFormats = []string{string(outFormat.Table), string(outFormat.Json)}
 
-var supportedTech = map[coreutils.Technology]struct{}{
-	coreutils.Npm: {},
+var supportedTech = map[coreutils.Technology]func() (bool, error){
+	coreutils.Npm: nil,
```
attiasas, line 58: I would always return true, and than remove the check for nil.

commands/audit/scarunner.go
```diff
@@ -292,18 +311,37 @@ func SetResolutionRepoIfExists(params xrayutils.AuditParams, tech coreutils.Tech
...
 		}
 		log.Debug("Unique dependencies list:\n" + clientutils.IndentJsonArray(jsonList))
 	}
-	uniqueNodes := []*xrayCmdUtils.GraphNode{}
-	for _, uniqueDep := range uniqueDeps {
-		uniqueNodes = append(uniqueNodes, &xrayCmdUtils.GraphNode{Id: uniqueDep})
-	}
-	return &xrayCmdUtils.GraphNode{Id: "root", Nodes: uniqueNodes}, nil
+	return nil
 }
```
attiasas, line 347:
```suggestion
func logDeps(uniqueDeps any) (err error) {
	if log.GetLogger().GetLogLevel() != log.DEBUG {
        // Avoid printing and marshaling if not on DEBUG mode.
        return
    }
	jsonList, err := json.Marshal(uniqueDeps)
	if errorutils.CheckError(err) != nil {
		return err
	}
	log.Debug("Unique dependencies list:\n" + clientutils.IndentJsonArray(jsonList))
	return
}
```

commands/audit/scarunner.go
```diff
@@ -220,17 +235,21 @@ func GetTechDependencyTree(params xrayutils.AuditParams, tech coreutils.Technolo
 	default:
 		err = errorutils.CheckErrorf("%s is currently not supported", string(tech))
 	}
-	if err != nil || len(uniqueDeps) == 0 {
+	if err != nil || (len(uniqueDeps) == 0 && uniqDepsWithTypes == nil) {
```
attiasas, line 238:
```suggestion
	if err != nil || (len(uniqueDeps) == 0 && len(uniqDepsWithTypes) == 0) {
```
len also check for nil, if its empty no need to continue as well right?

attiasas (changes requested):
Please add integration tests for curation maven at https://github.com/jfrog/jfrog-cli-security/blob/main/scans_test.go

commands/curation/curationaudit.go
asafambar: fetchNodeStatus does alot of requests and if some are failing, we still want to produce a report and prompt that there was an error. I'll add a comment

commands/curation/curationaudit.go
```diff
@@ -327,35 +350,46 @@ func (ca *CurationAuditCommand) CommandName() string {
...
 }
 
+func (ca *CurationAuditCommand) getRepoParams(projectType project.ProjectType) (*project.RepositoryConfig, error) {
+	configFilePath, exists, err := project.GetProjectConfFilePath(projectType)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, errorutils.CheckErrorf("no config file was found! Before running the " + projectType.String() + " command on a " +
+			"project for the first time, the project should be configured using the 'jf " + projectType.String() + "c' command")
```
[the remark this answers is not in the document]
asafambar: I think Eyal comment me last tome to change it to mvnc,npmc

PR #20, opened 2024-02-04
Add the logic of sending git info request to xsc

orz25:

-----

Add the logic of sending git info request to xsc (taken from client-go package) in order to have one multi scan id to multiple sca scans (we send a scan request per technology and we wish to associate all the scans to one multi scan id).

### DIFF
--- a/commands/audit/audit.go
+++ b/commands/audit/audit.go
@@ -2,10 +2,10 @@ package audit
 
 import (
 	"errors"
+	"github.com/jfrog/jfrog-cli-security/scangraph"
 	"os"
 
 	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
-	"github.com/jfrog/jfrog-cli-security/scangraph"
 	"github.com/jfrog/jfrog-cli-security/utils"
 	clientutils "github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/log"
@@ -170,6 +170,12 @@ func RunAudit(auditParams *AuditParams) (results *xrayutils.Results, err error) 
 	if results.ExtendedScanResults.EntitledForJas {
 		// Download (if needed) the analyzer manager in a background routine.
 		errGroup.Go(utils.DownloadAnalyzerManagerIfNeeded)
+	}
+
+	if auditParams.xrayGraphScanParams.XscGitInfoContext != nil {
+		if err = xrayutils.SendXscGitInfoRequestIfEnabled(auditParams.xrayGraphScanParams, xrayManager); err != nil {
+			return nil, err
+		}
 	}
 
 	// The sca scan doesn't require the analyzer manager, so it can run separately from the analyzer manager download routine.
--- a/go.mod
+++ b/go.mod
@@ -100,4 +100,4 @@ require (
 
 // replace github.com/jfrog/jfrog-cli-core/v2 => github.com/jfrog/jfrog-cli-core/v2 dev
 
-// replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go dev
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -100,4 +100,4 @@ require (
 
 // replace github.com/jfrog/jfrog-cli-core/v2 => github.com/jfrog/jfrog-cli-core/v2 dev
 
-// replace github.com/jfrog/jfrog-client-go => github.com/jfrog/jfrog-client-go dev
+replace github.com/jfrog/jfrog-client-go => github.com/orz25/jfrog-client-go v0.0.0-20240204100437-b823bf27a759
```
attiasas, line 103: don't forget to only merge this PR after pointing to client-go dev (i.e after merging the client-go PR and updating the go.mod)

utils/xraymanager.go
```diff
@@ -31,3 +35,22 @@ func CreateXrayServiceManagerAndGetVersion(serviceDetails *config.ServerDetails)
...
+		if err != nil {
+			return fmt.Errorf("failed sending Git Info request to XSC service, error: %s ", err.Error())
+		}
+		graphScanParams.MultiScanId = multiScanId
+		log.Debug(fmt.Sprintf("Created xsc git info successfully. multi_scan_id %s", multiScanId))
+		if err = os.Setenv("JF_MSI", multiScanId); err != nil {
+			// Not a fatal error, if not set the scan will not be shown at the XSC UI, should not fail the scan.
+			log.Debug(fmt.Sprintf("failed setting MSI as environment variable. Cause: %s", err.Error()))
+		}
+	}
```
attiasas, line 54:
```suggestion
    if graphScanParams.XscVersion == "" || graphScanParams.MultiScanId != "" {
        // No Xsc or multiScanId already provided.
        return
    }
    // Generate multiScanId and set it at the scan params.
    multiScanId, err := xrayManager.SendXscGitInfoRequest(graphScanParams.XscGitInfoContext)
	if err != nil {
		return fmt.Errorf("failed sending Git Info request to XSC service, error: %s ", err.Error())
	}
	graphScanParams.MultiScanId = multiScanId
	log.Debug(fmt.Sprintf("Created xsc git info successfully. multi_scan_id %s", multiScanId))
	if err = os.Setenv("JF_MSI", multiScanId); err != nil {
		// Not a fatal error, if not set the scan will not be shown at the XSC UI, should not fail the scan.
		log.Debug(fmt.Sprintf("failed setting MSI as environment variable. Cause: %s", err.Error()))
	}
```
keeping the code "flat" is more readable and easy to maintain and understand.

commands/audit/audit.go
```diff
@@ -172,6 +172,12 @@ func RunAudit(auditParams *AuditParams) (results *xrayutils.Results, err error)
 		errGroup.Go(utils.DownloadAnalyzerManagerIfNeeded)
 	}
 
+	if auditParams.xrayGraphScanParams.XscGitInfoContext != nil {
+		if err = xrayutils.SendXscGitInfoRequestIfEnabled(auditParams.xrayGraphScanParams, xrayManager); err != nil {
+			return nil, err
+		}
+	}
```
attiasas, line 179: I would suggest to add to `xrayutils.Results` attribute for multiScanId, also maybe to add it to the `SimpleJsonResults` so that it will be avaliable at the result to the user.
That will also allow you to add tests for checking if it was set / created (not empty)

attiasas (changes requested):
Please change the PR title to fit the changes, like adding option to set multiScanId for Xray scans....

commands/audit/audit.go
orz25: I added it to `xrayutils.Results`. Regarding `SimpleJsonResults` I'm not sure it's a significant value to present to the user. @hadarshjfrog What do you think?

hadarshjfrog: I think it's a good idea to add it there.
Will be very beneficial for e2e tests - we just need to make sure it doesn't break any schema to customers that rely on the results of simpleJson.

PR #21, opened 2024-02-13
Deleted setProxy logic from audit files and changed flow to use new function in cli-core

eranturgeman:

-----

Setting an Artifactory server as dependencies resolution registry logic got removed from 'audit' flow files and got moved to cli-core so we can use it in other flows as well.
Depends on: https://github.com/jfrog/jfrog-cli-core/pull/1128


attiasas (approved):
LGTM, Don't forget to only merge when replacing with core Jfrog dev branch

PR #31, opened 2024-02-25
Curation mvn enabled

asafambar:

-----

Depends on: https://github.com/jfrog/jfrog-cli-core/pull/1138


asafambar:
This PR enable curation support for maven according to xray and artifactory version, it also keep the option to enable it by env var

commands/curation/curationaudit.go
```diff
@@ -49,17 +50,50 @@ const (
...
 }
 
+func (ca *CurationAuditCommand) checkSupportByVersionOrEnv(tech coreutils.Technology, minRtVersion, minXrayVersion, envName string) (bool, error) {
+	flag, errResolveFlag := clientutils.GetBoolEnvValue(envName, false)
+	if errResolveFlag != nil {
+		log.Error(errResolveFlag)
+	}
+	if flag {
+		return true, nil
+	}
```
attiasas, line 74:
```suggestion
	if flag, errResolveFlag := clientutils.GetBoolEnvValue(envName, false); errResolveFlag != nil {
		log.Error(errResolveFlag)
	} else if flag {
		return true, nil
	}
```
Why are we forcing the use of this using env even if Xray and Artifactory are not supporting it? what is the use-case and how it will work?

commands/curation/curationaudit.go
```diff
@@ -49,17 +50,50 @@ const (
...
+	rtManager, serverDetails, err := ca.getRtManagerAndAuth(tech)
+	if err != nil {
+		return false, err
+	}
+	rtVersion, err := rtManager.GetVersion()
+	if err != nil {
+		return false, err
+	}
```
attiasas, line 82: I would extract it to a private method 
`func(ca *CurationAuditCommand) getArtifactoryVersion(tech coreutils.Technology)`
for readability.

commands/curation/curationaudit.go
```diff
@@ -49,17 +50,50 @@ const (
...
+	_, xrayVersion, err := utils.CreateXrayServiceManagerAndGetVersion(serverDetails)
+	if err != nil {
+		return false, err
+	}
+
+	xrayVersionErr := clientutils.ValidateMinimumVersion(clientutils.Xray, xrayVersion, minXrayVersion)
+	rtVersionErr := clientutils.ValidateMinimumVersion(clientutils.Artifactory, rtVersion, minRtVersion)
```
attiasas, line 89: after getting each version, we can validate the version right away, in case we can retrurn right away, no need to check and make more API requests because we got error.

commands/curation/curationaudit.go
```diff
@@ -210,6 +244,23 @@ func (ca *CurationAuditCommand) doCurateAudit(results map[string][]*PackageStatu
...
+	if ca.PackageManagerConfig == nil {
+		if err = ca.SetRepo(tech); err != nil {
+			return
+		}
+	}
+	serverDetails, err = ca.PackageManagerConfig.ServerDetails()
+	if err != nil {
+		return
+	}
+	rtManager, err = rtUtils.CreateServiceManager(serverDetails, 2, 0, false)
```
attiasas, line 257: Refactor values as consts.

commands/curation/curationaudit_test.go
```diff
@@ -398,30 +401,39 @@ func fillSyncedMap(pkgStatus []*PackageStatus) *sync.Map {
 	return &syncMap
 }
 
-func TestDoCurationAudit(t *testing.T) {
+func TestDoCurationAuditNpm(t *testing.T) {
```
attiasas, line 404: You removed the test for Maven, and made this name Npm (but the test cases also include the maven cases), is it mistake?...
if it is for all, make sure comments and var names also updated

attiasas (changes requested):

commands/curation/curationaudit.go
asafambar: this is for cases where you have milestones, snapshots version, mainly for development and demo environment

commands/curation/curationaudit.go
asafambar: this error is not server error but a "known" error which is "version not supported" I check both arti and xray in order to present the user both products version so he can act accordingly

commands/curation/curationaudit.go
asafambar: This is not a matter of consts in my opinion, thats what the function want, just read its signature

commands/curation/curationaudit_test.go
asafambar: will rename

PR #35, opened 2024-03-10
Curation - Generate a cache directory dedicated to each project

asafambar:

-----

Generate a cache directory dedicated to each project and inform the user that the initial setup might result in longer processing times due to an empty cache.

### DIFF
--- a/commands/audit/scarunner.go
+++ b/commands/audit/scarunner.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
 	"os"
 	"time"
 
@@ -174,12 +175,26 @@ func getDirectDependenciesFromTree(dependencyTrees []*xrayCmdUtils.GraphNode) []
 	return directDependencies.ToSlice()
 }
 
+func getCurationCacheByTech(tech coreutils.Technology) (string, error) {
+	if tech == coreutils.Maven {
+		return xrayutils.GetCurationMavenCacheFolder(true)
+	}
+	return "", nil
+}
+
 func GetTechDependencyTree(params xrayutils.AuditParams, tech coreutils.Technology) (flatTree *xrayCmdUtils.GraphNode, fullDependencyTrees []*xrayCmdUtils.GraphNode, err error) {
 	logMessage := fmt.Sprintf("Calculating %s dependencies", tech.ToFormal())
+	// in case it's not curation command these params will be empty
+	curationLogMsg, curationCacheFolder, err := getCurationCacheFolderAndLogMsg(params, tech)
+	if err != nil {
+		return
+	}
+	logMessage += curationLogMsg
 	log.Info(logMessage + "...")
 	if params.Progress() != nil {
 		params.Progress().SetHeadlineMsg(logMessage)
 	}
+
 	err = SetResolutionRepoIfExists(params, tech)
 	if err != nil {
 		return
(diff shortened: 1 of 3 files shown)


asafambar:
Added A dedicated message when curation cache folder observe its cache folder is empty for the specific project, this will prompt the user a message indicate that the current run can take longer as all dependencies and plugins need to be fetched.

utils/paths.go
```diff
@@ -35,10 +38,24 @@ func GetCurationCacheFolder() (string, error) {
...
+		}
+		// #nosec G401 -- Not a secret hash.
+		hasher := sha1.New()
+		_, err = hasher.Write([]byte(workingDir))
+		if err != nil {
+			return "", err
+		}
+		projectDir = hex.EncodeToString(hasher.Sum(nil))
+	}
+	return filepath.Join(curationFolder, "maven", projectDir), nil
```
attiasas, line 60:
```suggestion
        workingDir, err := os.Getwd()
	if err != nil {
		return
	}
	// #nosec G401 -- Not a secret hash.
	hasher := sha1.New()
	if _, err = hasher.Write([]byte(workingDir)); err != nil {
		return
	}
	projectDir = filepath.Join(curationFolder, "maven", hex.EncodeToString(hasher.Sum(nil)))
	return
```
change func return signature to `(projectDir string, err errror)`

utils/paths.go
```diff
@@ -35,10 +38,24 @@ func GetCurationCacheFolder() (string, error) {
 	return filepath.Join(curationFolder, "cache"), nil
 }
 
-func GetCurationMavenCacheFolder() (string, error) {
+func GetCurationMavenCacheFolder(withProjectDir bool) (string, error) {
```
attiasas, line 41: I can't see anywhere that we pass this func a 'false' value.
you can remove the parameter and change the func name to fit it better.

commands/audit/scarunner.go
```diff
@@ -238,6 +247,29 @@ func GetTechDependencyTree(params xrayutils.AuditParams, tech coreutils.Technolo
...
+			}
+		} else if err != nil {
+			return logMessage, curationCacheFolder, err
+		}
+
+		logMessage = ". Project's cache is currently empty, so this run may take longer to complete"
+
+	}
+	return logMessage, curationCacheFolder, err
+}
```
attiasas, line 271:
```suggestion
        if !params.IsCurationCmd() {
        	return
        }
        if curationCacheFolder, err = getCurationCacheByTech(tech); err != nil {
		return
	}
	dirExist, err := fileutils.IsDirExists(curationCacheFolder, false)
	if err != nil {
		return
	}
	if dirExist {
		if dirIsEmpty, err := fileutils.IsDirEmpty(curationCacheFolder); err != nil || !dirIsEmpty {
			return
		}
	}
	logMessage = ". Project's cache is currently empty, so this run may take longer to complete"
	return
}
```
* keep func as flat as possible for readability.
* If you named to returned values, you don't need to write them on return.

commands/audit/scarunner.go
```diff
@@ -174,12 +175,26 @@ func getDirectDependenciesFromTree(dependencyTrees []*xrayCmdUtils.GraphNode) []
...
+
 func GetTechDependencyTree(params xrayutils.AuditParams, tech coreutils.Technology) (flatTree *xrayCmdUtils.GraphNode, fullDependencyTrees []*xrayCmdUtils.GraphNode, err error) {
 	logMessage := fmt.Sprintf("Calculating %s dependencies", tech.ToFormal())
+	// in case it's not curation command these params will be empty
+	curationLogMsg, curationCacheFolder, err := getCurationCacheFolderAndLogMsg(params, tech)
+	if err != nil {
+		return
+	}
+	logMessage += curationLogMsg
```
attiasas, line 192:
```suggestion
	curationLogMsg, curationCacheFolder, err := getCurationCacheFolderAndLogMsg(params, tech)
	if err != nil {
		return
	}
	// In case it's not a curation command these params will be empty
	logMessage += curationLogMsg
```
comments starts with Captial letters.

attiasas (changes requested):
Please change the title to be shorter.
Add description with screen shots of the changes.

utils/paths.go
asafambar: yeah, actually its for future use when we will add an option to clear all cache, or for specific project, but yeah it can be removed until we will add it

PR #36, opened 2024-03-12
Add support for new contextual scan statuses

orz25:

-----


utils/resultstable.go
```diff
@@ -930,15 +930,14 @@ func convertCves(cves []services.Cve) []formats.CveRow {
...
 // Else (case when all cves aren't applicable) -> final value is not applicable
 func getApplicableCveStatus(entitledForJas bool, applicabilityScanResults []*sarif.Run, cves []formats.CveRow) ApplicabilityStatus {
-	if !entitledForJas || len(applicabilityScanResults) == 0 {
+	if !entitledForJas || len(applicabilityScanResults) == 0 || len(cves) == 0 {
 		return NotScanned
 	}
-	if len(cves) == 0 {
-		return ApplicabilityUndetermined
-	}
 	foundUndetermined := false
```
orz25, line 939: @attiasas I didn't use here the new function I created `setApplicabilityStatusFromRule` because the new function expect to get an array of statuses and here we have an array of cves.
I'm not sure what's better - loop over the cves to get the statuses (and covert the type of the status to string) in order to use the new function, or keep this duplicated logic.

attiasas: To avoid duplication, you can do the following:
avoiding duplications is better. loop over the Cves and collect the statuses.

commands/audit/jas/applicability/applicabilitymanager_test.go
```diff
@@ -334,3 +334,25 @@ func TestParseResults_AllCvesNotApplicable(t *testing.T) {
 		assert.NotEmpty(t, applicabilityManager.applicabilityScanResults[0].Results)
 	}
 }
+
+func TestParseResults_ApplicableCveExist1(t *testing.T) {
```
attiasas, line 338:
```suggestion
func TestParseResults_NewApplicabilityStatuses(t *testing.T) {
```

In addition I would consider merging the following tests, into testCases inside one method, to prevent duplicate code:
* This test method
* TestParseResults_EmptyResults_AllCvesShouldGetUnknown
* TestParseResults_ApplicableCveExist
* TestParseResults_AllCvesNotApplicable

utils/analyzermanager.go
```diff
@@ -45,6 +45,7 @@ const (
 	Applicable                ApplicabilityStatus = "Applicable"
 	NotApplicable             ApplicabilityStatus = "Not Applicable"
 	ApplicabilityUndetermined ApplicabilityStatus = "Undetermined"
+	NotCovered                ApplicabilityStatus = "Not covered"
```
attiasas, line 48:
```suggestion
	NotCovered                ApplicabilityStatus = "Not Covered"
```

utils/resultstable.go
```diff
@@ -930,15 +930,14 @@ func convertCves(cves []services.Cve) []formats.CveRow {
 
 // If at least one cve is applicable - final value is applicable
 // Else if at least one cve is undetermined - final value is undetermined
+// Else if all cves are not covered -> final value is not covered
 // Else (case when all cves aren't applicable) -> final value is not applicable
 func getApplicableCveStatus(entitledForJas bool, applicabilityScanResults []*sarif.Run, cves []formats.CveRow) ApplicabilityStatus {
-	if !entitledForJas || len(applicabilityScanResults) == 0 {
+	if !entitledForJas || len(applicabilityScanResults) == 0 || len(cves) == 0 {
 		return NotScanned
 	}
```
attiasas, line 938:
```suggestion
	if !entitledForJas || len(applicabilityScanResults) == 0 {
		return NotScanned
	}
    if len(cves) == 0 {
		return NotCovered
	}
```
it should be the default when JAS is enabled, we still show the column, it will be the only row with empty values...

utils/resultstable.go
```diff
@@ -962,16 +967,21 @@ func getCveApplicabilityField(cve formats.CveRow, applicabilityScanResults []*sa
...
+	var applicabilityStatuses []ApplicabilityStatus
 	for _, applicabilityRun := range applicabilityScanResults {
+		rule, _ := applicabilityRun.GetRuleById(CveToApplicabilityRuleId(cve.Id))
+		if rule != nil {
+			applicability.ScannerDescription = GetRuleFullDescription(rule)
+			status := getApplicabilityStatusFromRule(rule)
+			if status != "" {
+				applicabilityStatuses = append(applicabilityStatuses, status)
+			}
+		}
```
attiasas, line 979:
```suggestion
		if rule, _ := applicabilityRun.GetRuleById(CveToApplicabilityRuleId(cve.Id)); rule != nil {
			applicability.ScannerDescription = GetRuleFullDescription(rule)
			if status := getApplicabilityStatusFromRule(rule); status != "" {
				applicabilityStatuses = append(applicabilityStatuses, status)
			}
		}
```

utils/resultstable.go
```diff
@@ -1048,3 +1060,50 @@ func extractDependencyNameFromComponent(key string, techIdentifier string) (depe
...
+			return NotCovered
+		case "undetermined":
+			return ApplicabilityUndetermined
+		case "not_applicable":
+			return NotApplicable
+		case "applicable":
+			return Applicable
+		}
+	}
+	return ""
```
attiasas, line 1081:
```suggestion
	if status, ok := rule.Properties["applicability"].(string); ok {
		switch status {
        case "not_covered":
            return NotCovered
        case "undetermined":
            return ApplicabilityUndetermined
        case "not_applicable":
            return NotApplicable
        case "applicable":
            return Applicable
        }
	} else {
        log.Debug(fmt.Sprintf("Failed to get applicability status from rule properties for rule_id %s", rule.ID))
    }
	return NotScanned
```
Maybe the check for nil is redundent, if checking for string is also handling the nil case.
Comments/Logs should stat with Capital.

utils/resultstable.go
```diff
@@ -1048,3 +1060,50 @@ func extractDependencyNameFromComponent(key string, techIdentifier string) (depe
...
+}
+
+// If at least one status is applicable - final value is applicable
+// Else if at least one status is undetermined - final value is undetermined
+// Else if at least one status is not covered -> final value is not covered
+// Else (case when all statuses are not applicable) -> final value is not applicable
```
attiasas, line 1087: When no statuses are given, it will return `NotApplicable`, I don't think it is intended. make sure to add a comment on this method about this. or make sure to handle this case

utils/resultstable.go
```diff
@@ -1048,3 +1060,50 @@ func extractDependencyNameFromComponent(key string, techIdentifier string) (depe
...
+func setApplicabilityStatusFromRule(applicabilityStatuses []ApplicabilityStatus) string {
```
attiasas, line 1088:
```suggestion
func getFinalApplicabilityStatus(applicabilityStatuses []ApplicabilityStatus) string {
```
A more fitting name for this method

utils/resultstable_test.go
```diff
@@ -447,7 +447,7 @@ func TestGetApplicableCveValue(t *testing.T) {
 				EntitledForJas: true,
 			},
 			cves:           nil,
-			expectedResult: ApplicabilityUndetermined,
+			expectedResult: NotScanned,
```
attiasas, line 450: How it is `NotScanned` with JAS? before the changes it was `ApplicabilityUndetermined`, so it should be the new default when JAS is enabled `NotCovered`.

In addition I would suggest to add names to the tests to be clear what it tests.

utils/resultstable_test.go
```diff
@@ -1044,6 +1085,35 @@ func TestPrepareSast(t *testing.T) {
 	}
 }
 
+func TestSetApplicabilityStatusFromRule(t *testing.T) {
```
attiasas, line 1088: Add test case for NotApplicable

utils/resultwriter_test.go
```diff
@@ -92,6 +92,30 @@ func TestGetSarifTableDescription(t *testing.T) {
 			fixedVersions:       []string{},
 			expectedDescription: "| Severity Score | Direct Dependencies | Fixed Versions     |\n| :---:        |    :----:   |          :---: |\n| 3.0      | `example-package 3.0.0`       | No fix available   |",
 		},
+		{
+			name:                "Non-covered vulnerability",
```
attiasas, line 96:
```suggestion
			name:                "Not-covered vulnerability",
```

utils/resultwriter_test.go
```diff
@@ -92,6 +92,30 @@ func TestGetSarifTableDescription(t *testing.T) {
...
+			formattedDeps:       "`example-package 3.0.0`",
+			maxCveScore:         "3.0",
+			applicable:          "Not covered",
```
attiasas, line 99: change the applicable string field into `status ApplicabilityStatus` and than at the method use `status.String()`

utils/resultstable.go
orz25: In case we have 0 cves it means that Xray didn't find any cves for this vulnerability / violation.
I think it's different than not covered - which means that the scanner did run but there is no rule for this cve.

utils/resultstable.go
orz25: @attiasas It looks like `rule.Properties["applicability"].(string)` checks both whether the key exists in the map and whether its value is of the expected type - but it considered good practice to check for the existence of a key before attempting to access it.

attiasas (changes requested):
Please Adjust the integration tests to also include all the statuses.

Make sure with the PM that the order of the statues are correct.

tests/utils/test_validation.go
```diff
@@ -53,22 +53,26 @@ func VerifySimpleJsonScanResults(t *testing.T, content string, minVulnerabilitie
 	}
 }
 
-func VerifySimpleJsonJasResults(t *testing.T, content string, minSastViolations, minIacViolations, minSecrets, minApplicable int) {
+func VerifySimpleJsonJasResults(t *testing.T, content string, minSastViolations, minIacViolations, minSecrets, minApplicable, minNotCovered int) {
```
attiasas, line 56: notApplicable is always 1? why not pass it as a parameter as well? what about undetermined status?

attiasas (approved):

PR #41, opened 2024-03-17
Additions to support Pnpm in frogbot

eranturgeman:

-----

This pull request implements modifications to the Pnpm audit flow in order to facilitate Frogbot's PNPM support. When auditing a Pnpm project and performing the 'install' operation, a node_modules directory is generated, which becomes redundant after the dependency calculation. Thus, we remove it if it didn't exist previously, ensuring no residual trace is left behind from the 'install' action.


commands/audit/sca/pnpm/pnpm.go
```diff
@@ -46,10 +48,22 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTrees []*xrayUtils
...
+		dirToCalcDependenciesOn = currentDir
+	} else {
+		// If tempDirForDependenciesCalculation contains a non-empty value, it means we executed 'install' on a temporary directory on which we need to calculate the dependencies and remove it at the end
+		dirToCalcDependenciesOn = tempDirForDependenciesCalculation
+		defer func() {
+			err = errors.Join(err, biutils.RemoveTempDir(tempDirForDependenciesCalculation))
+		}()
+	}
+	return calculateDependencies(pnpmExecPath, dirToCalcDependenciesOn, params)
 }
```
omerzi, line 67: I think it is better to install the project in an isolated directory anyway. Those conditions make the code harder to read, harder to maintain, and harder to extend. It's a better practice not to work in their project directory.

commands/audit/sca/pnpm/pnpm.go
```diff
@@ -46,10 +48,21 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTrees []*xrayUtils
...
 	}
 	// Build
-	if err = installProjectIfNeeded(pnpmExecPath, currentDir); errorutils.CheckError(err) != nil {
+	var dirForDependenciesCalculation string
+	if dirForDependenciesCalculation, err = installProjectIfNeeded(pnpmExecPath, currentDir); errorutils.CheckError(err) != nil {
 		return
 	}
-	return calculateDependencies(pnpmExecPath, currentDir, params)
+
+	if dirForDependenciesCalculation == "" {
```
omerzi, line 56: Is this case still relevant?

eranturgeman: yes because if we do not need to install we will get an empty value and the directory on which we need to calculate the dependencies is the original directory.
In this case the original dir will not be effected so we are all good

attiasas (approved):

PR #83, opened 2024-06-17
Add consistent behavior to table sorting in jf scan and jf audit

dortam888:

-----

Add consistent behavior to sorting for jf scan and jf audit based on the application: 

1. Severity
2. Applicability
3. JFrog Severity
4. Xray ID


utils/resultstable.go
```diff
@@ -257,11 +257,26 @@ func sortVulnerabilityOrViolationRows(rows []formats.VulnerabilityOrViolationRow
...
 	})
 }
 
+func sortApplicable(applicableFirst string, applicableSecond string) bool {
+	applicableMap := make(map[string]int)
+
+	applicableMap[Applicable.String()] = 4
+	applicableMap[NotApplicable.String()] = 3
+	applicableMap[ApplicabilityUndetermined.String()] = 2
+	applicableMap[NotCovered.String()] = 1
```
attiasas, line 275: you are creating this map for every call to the `sortApplicable` function, this should be implemented as const.
It addition, there could be scanning without JAS, so another applicable value could be `NotScanned` that you are missing

utils/resultstable.go
```diff
@@ -257,11 +257,26 @@ func sortVulnerabilityOrViolationRows(rows []formats.VulnerabilityOrViolationRow
 	sort.Slice(rows, func(i, j int) bool {
 		if rows[i].SeverityNumValue != rows[j].SeverityNumValue {
 			return rows[i].SeverityNumValue > rows[j].SeverityNumValue
+		} else if rows[i].Applicable != rows[j].Applicable {
+			return sortApplicable(rows[i].Applicable, rows[j].Applicable)
+		} else if rows[i].JfrogResearchInformation.SeverityNumValue != rows[j].JfrogResearchInformation.SeverityNumValue {
```
attiasas, line 262: `JfrogResearchInformation` could be `nil` (optional) please handle this case

attiasas (changes requested):

utils/analyzermanager.go
```diff
@@ -70,6 +70,22 @@ func convertToApplicabilityStatus(status string) ApplicabilityStatus {
...
 }
 
+// for sorting
+var applicableMapToScore = map[string]int{
+	"Applicable":                4,
+	"NotApplicable":             3,
+	"ApplicabilityUndetermined": 2,
+	"NotCovered":                1,
+	"NotScanned":                0,
+}
```
attiasas, line 80: Did you make sure those values are correct? I think you are mixing the var names and the var value.
image
In addition, Not applicable should be the least priority, check out: `resultstable.go` -> `Severities` var

utils/resultstable.go
```diff
@@ -255,13 +255,36 @@ func prepareVulnerabilities(vulnerabilities []services.Vulnerability, results *R
...
+			return convertApplicableToScore(rows[i].Applicable) > convertApplicableToScore(rows[j].Applicable)
+		case rows[i].JfrogResearchInformation != nil || rows[j].JfrogResearchInformation != nil:
+			ok, value := JfrogResearchInformationSort(rows[i].JfrogResearchInformation, rows[j].JfrogResearchInformation)
+			if ok {
+				return value
+			}
+			return rows[i].IssueId > rows[j].IssueId
+		default:
+			return rows[i].IssueId > rows[j].IssueId
 		}
```
attiasas, line 271:
```suggestion
		if rows[i].SeverityNumValue != rows[j].SeverityNumValue {
			return rows[i].SeverityNumValue > rows[j].SeverityNumValue
		}
		if rows[i].Applicable != rows[j].Applicable {
		         return convertApplicableToScore(rows[i].Applicable) > convertApplicableToScore(rows[j].Applicable)
		}
		if rows[i].JfrogResearchInformation != nil || rows[j].JfrogResearchInformation != nil:
			ok, value := JfrogResearchInformationSort(rows[i].JfrogResearchInformation, rows[j].JfrogResearchInformation)
			if ok {
				return value
			}
		}
		return rows[i].IssueId > rows[j].IssueId
```
I would use `if` statements for this, `switch` statement is not very readable in case you want to see what checks are performed first.
Also removed duplication for `return rows[i].IssueId > rows[j].IssueId`

utils/resultstable.go
```diff
@@ -255,13 +255,36 @@ func prepareVulnerabilities(vulnerabilities []services.Vulnerability, results *R
...
+	case jfrogResearchInformation1 == nil:
+		return true, true
+	case jfrogResearchInformation2 == nil:
+		return true, false
+	case jfrogResearchInformation1.SeverityNumValue != jfrogResearchInformation2.SeverityNumValue:
+		return true, jfrogResearchInformation1.SeverityNumValue > jfrogResearchInformation2.SeverityNumValue
+	default:
+		return false, false
+	}
+}
```
attiasas, line 286: * Rename for readability (suggestion: `CompareJfrogResearchInformation`). 
* add a description to the func, its not very easy to guess what is the output values. why do we need 2 values? usually. for comparison, the output has only one int value (0 == equal, 0> , 0<)
* What happened if both are nil?

utils/analyzermanager.go
dortam888: @attiasas Good point! But checking it against the application sorting for applicable gives me the following order:
1. Applicable
2. Not Applicable  
3. Undetermined
4. Not Scanned
5. Not Covered

attiasas (approved):

utils/resultstable.go
```diff
@@ -258,10 +258,37 @@ func sortVulnerabilityOrViolationRows(rows []formats.VulnerabilityOrViolationRow
...
 	})
 }
 
+// CompareJfrogResearchInformation compares between two JFrog Research imformation pointers
+// It returns two boolean values: First to validate if comparison is needed and second is the comparison result
+// If both values don't exist return false in first value no need to compare
+// If one of them exist and the other not return true if only the first exist and false if only the second
+// If both exist check if the value is different and compare the values
+// If values are equal return false in the first value
+func CompareJfrogResearchInformation(jfrogResearchInformation1 *formats.JfrogResearchInformation, jfrogResearchInformation2 *formats.JfrogResearchInformation) (bool, bool) {
```
attiasas, line 277: you can create a method to get priority of jfrogResearchInformation1 that will handle the nil.
In addition to make this current method return one value, you can use the priority jus like you are comparing at the main func that calls it..

utils/analyzermanager.go
```diff
@@ -70,6 +70,22 @@ func convertToApplicabilityStatus(status string) ApplicabilityStatus {
...
+	"NotCovered":                1,
+	"NotApplicable":             0,
+}
+
+func convertApplicableToScore(applicability string) int {
+	if level, ok := applicableMapToScore[strings.ToLower(applicability)]; ok {
+		return level
+	}
+	return -1
+}
```
attiasas, line 87: move to `utils/jasutils.go`

attiasas (changes requested):

utils/resultstable.go
```diff
@@ -265,15 +266,35 @@ func prepareVulnerabilities(vulnerabilities []services.Vulnerability, results *R
...
+		}
+		return rows[i].IssueId > rows[j].IssueId
 	})
 }
 
+// getJfrogResearchPriority returns the score of JFrog Research Severity.
+// If there is no such severity will return the const value noJfrogResearchInformation which is less than all possible scores
+func getJfrogResearchPriority(jfrogResearchInformation *formats.JfrogResearchInformation) int {
+	if jfrogResearchInformation == nil {
+		return noJfrogResearchInformation
```
attiasas, line 292: can't we just return the defualt severity in case it is not existing? we want to also take care of the use case that a research severity is lower that normal severity of the other (without research info)

attiasas (changes requested):

attiasas (approved):

PR #88, opened 2024-06-19
Fix resolving Maven dependency tree with snapshot versions

asafambar:

-----
Description:

Building the Maven dependency tree for security commands can sometimes fail when the project dependencies include snapshot versions. This occurs because the settings.xml file created for configuration only includes a mirror tag, which Maven apparently sometimes fails to resolve for snapshot versions. To resolve this issue, I added a repository section to the settings.xml with the snapshot tag enabled, which fixed the issue. a reference to this issue

https://stackoverflow.com/questions/17348118/maven-not-downloading-recognizing-snapshot

### DIFF
--- a/artifactory_test.go
+++ b/artifactory_test.go
@@ -70,6 +70,12 @@ func TestDependencyResolutionFromArtifactory(t *testing.T) {
 			projectType:     project.Maven,
 		},
 		{
+			testProjectPath: []string{"maven", "maven-snapshot"},
+			resolveRepoName: securityTests.MvnVirtualRepo,
+			cacheRepoName:   securityTests.MvnRemoteRepo,
+			projectType:     project.Maven,
+		},
+		{
 			testProjectPath: []string{"go", "simple-project"},
 			resolveRepoName: securityTests.GoVirtualRepo,
 			cacheRepoName:   securityTests.GoRemoteRepo,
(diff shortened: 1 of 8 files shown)


artifactory_test.go
```diff
@@ -120,8 +126,10 @@ func testSingleTechDependencyResolution(t *testing.T, testProjectPartialPath []s
 		Url:            *securityTests.JfrogUrl,
 		ArtifactoryUrl: *securityTests.JfrogUrl + securityTests.ArtifactoryEndpoint,
 		XrayUrl:        *securityTests.JfrogUrl + securityTests.XrayEndpoint,
-		AccessToken:    *securityTests.JfrogAccessToken,
-		ServerId:       securityTests.ServerId,
+		//AccessToken:    *securityTests.JfrogAccessToken,
```
attiasas, line 129:
```suggestion
		AccessToken:    *securityTests.JfrogAccessToken,
```
Tests are using access token, we are not providing a user and password. as far as I know maven should work with it

attiasas (commented):
LGTM, nice work! check out my comment

artifactory_test.go
asafambar: sure

attiasas (approved):

PR #106, opened 2024-07-08
Fixing return url from build scan to be the CLI config url 

dortam888:

Change URL in scan-build response to be the configured URL
-----

In the response from the scan-build API Xray will always return the Base Url. In some cases it is different than the URL that the customer configured in the CLI. 
I replaced the response URL with the URL that is configured in our CLI.


commands/scan/buildscan.go
```diff
@@ -170,3 +185,21 @@ func (bsc *BuildScanCommand) runBuildScanAndPrintResults(xrayManager *xray.XrayS
 func (bsc *BuildScanCommand) CommandName() string {
 	return "xr_build_scan"
 }
+
+func trimBuildScanResultUrl(fullUrl string) (string, string, error) {
+
+	if fullUrl == "" {
+		return "", "", fmt.Errorf("URL to trim is empty")
```
attiasas, line 192: is it a mandatory attribute? or is the URL optional? if its not, no error is needed if empty

commands/scan/buildscan.go
```diff
@@ -170,3 +185,21 @@ func (bsc *BuildScanCommand) runBuildScanAndPrintResults(xrayManager *xray.XrayS
 func (bsc *BuildScanCommand) CommandName() string {
 	return "xr_build_scan"
 }
+
+func trimBuildScanResultUrl(fullUrl string) (string, string, error) {
+
```
attiasas, line 190:
```suggestion
```

commands/scan/buildscan.go
dortam888: Generally, we expect the URL not to be empty from the Xray API. 
In case it is an empty url.Parse will not return an error; the function will return ":///" instead of empty, which may cause unexpected behavior in the future. 
generally: 
`133 if url != bsc.serverDetails.Url` 
will deal with empty URLs so the check can be removed. 
wdyt?

attiasas: `if url != bsc.serverDetails.Url` will handle this case as you mentioned.
to keep the function woking in all use cases as a stand alone, I would keep the test but it is not required.
```
if fullUrl == "" {
      return "", "", nil
```

attiasas (approved):

PR #118, opened 2024-07-25
Added Risk reason and other violations fields to on-demand scan

dortam888:

-----

Some fields of violation were dropped when inserting the violation to simplify violation function. 
Added a function to include all violation fields to simplify violations function to fix the print table results.

op-risk-table

### DIFF
--- a/utils/resultstable_test.go
+++ b/utils/resultstable_test.go
@@ -92,7 +92,7 @@ func TestSimplifyViolation(t *testing.T) {
 	}
 	simpliedViolations := simplifyViolations(violations, true)
 
-	//assert that the length of simplified violations is length of violation without the similar componentID
+	// assert that the length of simplified violations is length of violation without the similar componentID
 	assert.Equal(t, len(violations)-1, len(simpliedViolations))
 }
 
--- a/utils/resultstable_test.go
+++ b/utils/resultstable_test.go
@@ -132,13 +132,46 @@ func TestGetOperationalRiskReadableData(t *testing.T) {
 			&operationalRiskViolationReadableData{"true", "3.5", "55", "10", "no maintainers", "EOL", "1.2.3", "5"},
 		},
 	}
-	simplifiedViolations := simplifyViolations([]services.Violation{tests[0].violation, tests[1].violation}, true)
 
-	for i, test := range tests {
-		results := getOperationalRiskViolationReadableData(test.violation)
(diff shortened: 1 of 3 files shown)


utils/resultstable_test.go
```diff
@@ -83,26 +83,42 @@ func TestGetDirectComponents(t *testing.T) {
...
 			&operationalRiskViolationReadableData{"N/A", "N/A", "N/A", "N/A", "", "", "N/A", "N/A"},
 		},
 		{
-			services.Violation{IsEol: newBoolPtr(true), LatestVersion: "1.2.3", NewerVersions: newIntPtr(5),
+			services.Violation{Components: map[string]services.Component{"gav://antparent:ant:1.6.5": {}}, IsEol: newBoolPtr(true), LatestVersion: "1.2.3", NewerVersions: newIntPtr(5),
 				Cadence: newFloat64Ptr(3.5), Commits: newInt64Ptr(55), Committers: newIntPtr(10), EolMessage: "no maintainers", RiskReason: "EOL"},
 			&operationalRiskViolationReadableData{"true", "3.5", "55", "10", "no maintainers", "EOL", "1.2.3", "5"},
 		},
 	}
+	simplifiedViolations := simplifyViolations([]services.Violation{tests[0].violation, tests[1].violation}, true)
```
attiasas, line 115: Not sure what you are doing here, why is this logic outside of the test cases loop?

If you have similar data in the test for all cases you can set a `var violations []services.Violation` at the start of the test, then use it and set the right violation by index for each test case.

utils/resultstable_test.go
```diff
@@ -83,26 +83,42 @@ func TestGetDirectComponents(t *testing.T) {
...
 
-	for _, test := range tests {
+	for i, test := range tests {
 		results := getOperationalRiskViolationReadableData(test.violation)
+		simplifiedResults := getOperationalRiskViolationReadableData(simplifiedViolations[i])
 		assert.Equal(t, test.expectedResults, results)
+		assert.Equal(t, test.expectedResults, simplifiedResults)
```
attiasas, line 121: what are you testing here? its not clear why expected results should be equal to result and simplified results...

attiasas (changes requested):

utils/resultstable_test.go
dortam888: @attiasas I'm testing that the simplifiedViolations is returning all the fields. 
Before the fix this test would fail due to missing fields in simplified results. 
I'll add comments

PR #125, opened 2024-07-31
Curation summary Fix - The count of blocked policies has changed.

asafambar:

-----
Description:
A single package can be blocked in multiple paths, which is why we changed it to count the keys instead of the occurrences.
This will promise that the count number on each policy wont be greater than the total number of packages which is calculate by the package name and not by package name + parent name.


attiasas (commented):
Please change the tests to reflect this change

PR #129, opened 2024-08-04
fix SARIF format security-severity attribute

attiasas:

-----

When generating the SARIF format, the added rule attribute `security-severity` was added as float but should have been set as string


hadarshjfrog:
The fix is 100% - but why should it have been a string instead of a float?

hadarshjfrog (approved):

PR #138, opened 2024-08-08
Record SARIF results after security commands to upload for GitHub

attiasas:

-----

Depends on: 
* https://github.com/jfrog/jfrog-cli-core/pull/1230

## Record SARIF results after security commands to upload for GitHub

Supported commands:  `docker scan` and `scan`

When running one of those commands using https://github.com/jfrog/setup-jfrog-cli, the command will record the results in SARIF format and be uploaded to the GitHub Security tab if applicable.

image

### Additional Improvements:
* Patch to SARIF content to pass Github Ingestion Rules and to add additional context to the attributes
* Job Summary - if docker image tag was not provided to `jf docker scan <image>` save record arg as `<image>:latest`
* Job Summary - pad curation summary details tag with `\n` at the end to fix section

To Support uploading results from the CLI directly to GitHub Code scanning, transferring code from:
https://github.com/jfrog/frogbot/blob/master/utils/utils.go#L241


commands/audit/audit.go
```diff
@@ -139,7 +139,6 @@ func (auditCmd *AuditCommand) Run() (err error) {
 		SetOutputFormat(auditCmd.OutputFormat()).
 		SetPrintExtendedTable(auditCmd.PrintExtendedTable).
 		SetExtraMessages(messages).
-		SetScanType(services.Dependency).
```
hadarshjfrog, line 142: why are we removing the scanType?
here and on `commands/scan/buildscan.go` ?

attiasas: I added a new attribute to the result `ResultType` as a `CommandType` a new enum in utils that describes the cmd that the consequence belongs to. so we no longer need to use the `services` enum because we have this info in Results

commands/scan/scan.go
```diff
@@ -190,7 +190,10 @@ func (scanCmd *ScanCommand) indexFile(filePath string) (*xrayUtils.BinaryGraphNo
 }
 
 func (scanCmd *ScanCommand) Run() (err error) {
-	return scanCmd.RunAndRecordResults(func(scanResults *utils.Results) error {
+	return scanCmd.RunAndRecordResults(utils.Binary, func(scanResults *utils.Results) (err error) {
+		if err = utils.RecordSarifOutput(scanResults); err != nil {
+			return
```
hadarshjfrog, line 195: shouldn't it be `return err`?

commands/scan/scan.go
```diff
@@ -204,7 +207,7 @@ func hasViolationContext(watches []string, projectKey string) bool {
 	return len(watches) > 0 || projectKey != ""
 }
 
-func (scanCmd *ScanCommand) RunAndRecordResults(recordResFunc func(scanResults *utils.Results) error) (err error) {
+func (scanCmd *ScanCommand) RunAndRecordResults(cmdType utils.CommandType, recordResFunc func(scanResults *utils.Results) error) (err error) {
```
hadarshjfrog, line 210: can this function be split a bit more?
It's over 100 lines - and looks like doing logic in multiple ares

formats/sarifutils/sarifutils.go
```diff
@@ -19,12 +17,30 @@ func NewReport() (*sarif.Report, error) {
...
-		return "", errorutils.CheckError(err)
+func NewPhysicalLocation(physicalPath string) *sarif.PhysicalLocation {
+	return &sarif.PhysicalLocation{
+		ArtifactLocation: &sarif.ArtifactLocation{
+			URI: &physicalPath,
+		},
+	}
+}
+
+func NewPhysicalLocationWithRegion(physicalPath string, startRow, endRow, startCol, endCol int) *sarif.PhysicalLocation {
```
hadarshjfrog, line 28: why do we pass the integers by ref and not value? its into an object you create now - will it change afterwards?

formats/sarifutils/sarifutils.go
```diff
@@ -19,12 +17,30 @@ func NewReport() (*sarif.Report, error) {
...
+	location.Region = &sarif.Region{
+		StartLine:   &startRow,
+		EndLine:     &endRow,
+		StartColumn: &startCol,
+		EndColumn:   &endCol,
+	}
+	return location
+}
+
+func NewLogicalLocation(name, kind string) *sarif.LogicalLocation {
```
hadarshjfrog, line 39: same as `NewPhysicalLocationWithRegion` - need to pass by ref?

utils/resultwriter.go
```diff
@@ -25,11 +28,23 @@ import (
...
+	maxPossibleCve  = 10.0
+
+	// #nosec G101 -- Not credentials.
+	patchedBinarySecretScannerToolName = "JFrog Binary Secrets Scanner"
+	jfrogFingerprintAlgorithmName      = "jfrogFingerprintHash"
 )
 
-const MissingCveScore = "0"
-const maxPossibleCve = 10.0
+var (
```
hadarshjfrog, line 43: why var and not the prevous const? are they changing?

utils/resultwriter_test.go
```diff
@@ -573,6 +576,261 @@ func TestGetSummary(t *testing.T) {
...
+			subScan:   ScaScan,
+			input: []*sarif.Run{
+				sarifutils.CreateRunWithDummyResultsInWd(wd, sarifutils.CreateDummyResultInPath(fmt.Sprintf("file://%s", filepath.Join(wd, "dir", "file")))),
+			},
+			expectedResults: []*sarif.Run{
+				sarifutils.CreateRunWithDummyResultsInWd(wd, sarifutils.CreateDummyResultInPath(filepath.Join("dir", "file"))),
+			},
+		},
+		{
+			name:      "Docker image scan - SCA",
```
hadarshjfrog, line 673: did we test CA on Docker Scan?

commands/scan/scan.go
attiasas: it will be changed after the refactor, not relevant for this PR

commands/scan/scan.go
attiasas: no, we have a named output attribute `err`, only assign is needed

formats/sarifutils/sarifutils.go
attiasas: yes, this is required by the library we use, it holds pointers

formats/sarifutils/sarifutils.go
attiasas: yes, this is required by the library we use, it holds pointers

utils/resultwriter.go
attiasas: they just moved to a centered definition of costs

utils/resultwriter_test.go
attiasas: added

hadarshjfrog (approved):

PR #153, opened 2024-08-21
Missing context new status for CA

barv-jfrog:

-----
Description:
New status in contextual analysis - missing context. If a CVE gets a missing context under contextual analysis this means that this CVE exists on binary but not on source code. (For example a scanner exists but is of type ELF, YAML, etc - not source code). This helps to distinguish between Not covered which means this CVE doesn't have a jfrog scanner at all to Missing context which means the CVE is covered but not on source code.

Since missing context doesn't mean Not applicable it has a higher priority. Also its not that its not Covered, just not covered in source code, which in conclusion we believe deserves a higher priority than Not covered as well.

Additional description:
Jfrog research team work on many CVEs using scanners they create. These scanners have a type, like JS, JAVASRC, YAML, etc..
When using jf audit we scan for src code, and therefore the scanners which can be run on the relevant project are only scanners of source code types which are being supported by the CLI (JAVASRC, JS, Python, NuGet soon and more) but not on scanners that are of type YAML,  text, binary, elf, etc, however, these CVEs are supported on binary scanning on xray platform - So we want to distinguish between those CVEs that are not covered at all (no scanner) and those that are covered on xray platform but not on source code scanning (jf audit).

This CA status is only for jf audit.


hadarshjfrog:
Please add relevant results to SARIFs in: `tests/testdata/other/applicability-scan`.
Also - can you add a project that will return such results when being scan?

utils/severityutils/severity.go
```diff
@@ -88,32 +88,37 @@ func (sd SeverityDetails) ToDetails(severity Severity, pretty bool) formats.Seve
 
 var Severities = map[Severity]map[jasutils.ApplicabilityStatus]*SeverityDetails{
 	Critical: {
-		jasutils.Applicable:                &SeverityDetails{Priority: 20, Score: MaxCveScore, Emoji: "💀", style: color.New(color.BgLightRed, color.LightWhite)},
-		jasutils.ApplicabilityUndetermined: &SeverityDetails{Priority: 19, Score: MaxCveScore, Emoji: "💀", style: color.New(color.BgLightRed, color.LightWhite)},
+		jasutils.Applicable:                &SeverityDetails{Priority: 21, Score: MaxCveScore, Emoji: "💀", style: color.New(color.BgLightRed, color.LightWhite)},
+		jasutils.ApplicabilityUndetermined: &SeverityDetails{Priority: 20, Score: MaxCveScore, Emoji: "💀", style: color.New(color.BgLightRed, color.LightWhite)},
+		jasutils.MissingContext:            &SeverityDetails{Priority: 19, Score: MaxCveScore, Emoji: "💀", style: color.New(color.BgLightRed, color.LightWhite)},
```
attiasas, line 93: you have entries with similar Priority values. make sure that you don't have duplications in values.

utils/resultstable.go
```diff
@@ -1057,6 +1060,9 @@ func getFinalApplicabilityStatus(applicabilityStatuses []jasutils.ApplicabilityS
 	foundUndetermined := false
 	foundNotCovered := false
 	for _, status := range applicabilityStatuses {
+		if status == jasutils.MissingContext {
+			return jasutils.MissingContext
+		}
```
attiasas, line 1065: Missing Context does not have the first priority.
If we have a case of cve with MissingContext and another Cve with Applicable. we want to return the Applicable. only if not Applicable you can return the `MissingContext`, make it similar logic to `ApplicabilityUndetermined`

utils/jasutils/jasutils.go
```diff
@@ -75,9 +78,10 @@ func ApplicabilityRuleIdToCve(sarifRuleId string) string {
 }
 
 var applicableMapToScore = map[string]int{
-	"Applicable":                4,
-	"ApplicabilityUndetermined": 3,
-	"NotScanned":                2,
+	"Applicable":                5,
+	"ApplicabilityUndetermined": 4,
+	"NotScanned":                3,
+	"MissingContext":            2,
```
attiasas, line 84: make sure priority is matching the order in all places

attiasas (changes requested):
Check out my comments. Also. add to integration tests the status. (validation + issue with the status)
Like in the validation methods: `VerifySimpleJsonJasResults`

utils/severityutils/severity.go
barv-jfrog: no duplications

utils/jasutils/jasutils.go
barv-jfrog: made sure

attiasas (approved):

tests/testdata/projects/package-managers/go/missing-context/go.mod
```diff
@@ -0,0 +1,160 @@
...
+
+go 1.22
+
+require (
+	cloud.google.com/go v0.38.0 // indirect
+	github.com/Azure/azure-sdk-for-go v40.3.0+incompatible // indirect
+	github.com/Azure/go-autorest/autorest v0.10.0 // indirect
+	github.com/Azure/go-autorest/autorest/adal v0.8.2 // indirect
+	github.com/Azure/go-autorest/autorest/azure/auth v0.4.2 // indirect
+	github.com/Azure/go-autorest/autorest/azure/cli v0.3.1 // indirect
```
attiasas, line 11: Please only include the needed dependency for the test to reduce size and test times

attiasas (changes requested):

attiasas (approved):

PR #154, opened 2024-08-24
Fixing jf dockerscan command panic

dortam888:

-----

the JFrog cli might offer jf dockerscan as a command as its appearing on the command list but it has no Action function.
Added an action function to return the same output as jf docker scan would return.
When running jf dockerscan:
jf dockerscan
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1048cbc40]

goroutine 1 [running]:
github.com/jfrog/jfrog-cli-core/v2/plugins/components.convertCommand.getActionFunc.func2(0x14000341718?)
	github.com/jfrog/jfrog-cli-core/v2@v2.55.3/plugins/components/conversionlayer.go:374 +0x50
github.com/urfave/cli.HandleAction({0x105275ea0?, 0x14000419110?}, 0xa?)
	github.com/urfave/cli@v1.22.15/app.go:522 +0xd8
github.com/urfave/cli.Command.Run({{0x104e7826d, 0xa}, {0x0, 0x0}, {0x0, 0x0, 0x0}, {0x104f08f05, 0x39}, {0x1400049eba0, ...}, ...}, ...)
	github.com/urfave/cli@v1.22.15/command.go:175 +0x530
github.com/urfave/cli.(*App).Run(0x140003b41c0, {0x140000300a0, 0x2, 0x2})
	github.com/urfave/cli@v1.22.15/app.go:277 +0x7ec
main.execMain()
	github.com/jfrog/jfrog-cli/main.go:132 +0x454
main.main()
	github.com/jfrog/jfrog-cli/main.go:69 +0x20

Fixing it by making it a  n help command.
	
with: https://github.com/jfrog/jfrog-cli/pull/2726


attiasas (changes requested):
what is the panic? please add to the description.
I could not replicate the issue.
Also this is a hidden command that inject and shows the `HELP` for `jf docker scan` it is not a command on its own.

dortam888:
@attiasas When running jf dockerscan even if by accident we're getting panic instead of handling.
We have many hidden commands but with action function. 
Either jf dockerscan will do the same as jf docker scan (that's what I implemented) 
Or we should output something like:
'jf dockerscan' is not a jf command. See --help
The most similar command is:
	jf docker scan 

Take in account that if we're running something close to jf dockerscan it will output:
'jf dockersca' is not a jf command. See --help
The most similar command is:
	jf dockerscan
And I can't change it since its reading it as a command

attiasas:
@dortam888 , As discussed, this needs to be behave similar to `dockerpullhelp` in the CLI project

cli/docs/scan/dockerscan/help.go
```diff
@@ -1,18 +1,12 @@
...
 }
 
-func GetArguments() []components.Argument {
-	return []components.Argument{
-		{
-			Name:        "image tag",
-			Description: "The docker image tag to scan.",
-		},
-	}
+func GetArguments() string {
```
eranturgeman, line 9: return this as []components.Argument as we do so for all other commands. we don't want to create a different types just for one single command

PR #155, opened 2024-08-25
Undetermined reason for undetermined contextual analysis status

barv-jfrog:

-----
Description:
Scanners return a reason for why the result was undetermined. This result is on the scanner level (Rule), and not on evidence level in contrary to applicable status which returns "reason" on the evidence level. We need to support this undetermined reason and show it next to "applicable" and "scannerDescription".


formats/sarifutils/sarifutils.go
```diff
@@ -234,6 +234,21 @@ func GetRuleFullDescription(rule *sarif.ReportingDescriptor) string {
...
+		if !ok {
+			return ""
+		}
+		return prop
+	}
+	return ""
+}
+
+func GetRuleUndeterminedReason(rule *sarif.ReportingDescriptor) string {
+	return GetRuleProperty("undetermined_reason", rule)
```
hadarshjfrog, line 249: The function split is positive overall - but I don't see any other usage for `GetRuleProperty`, so in that case, I don't see a reason for the dedicated fucntion.
Couldn't we use it for the applicability status itself?

formats/sarifutils/test_sarifutils.go
```diff
@@ -13,14 +13,16 @@ func CreateRunWithDummyResults(results ...*sarif.Result) *sarif.Run {
 	return run
 }
 
-func CreateRunWithDummyResultAndRuleProperties(property, value string, result *sarif.Result) *sarif.Run {
+func CreateRunWithDummyResultAndRuleMultipleProperties(result *sarif.Result, properties, values []string) *sarif.Run {
```
hadarshjfrog, line 16: The function name indicates that we have a function for multipleProperties, and a one for single.
If it's just that one function - I think you can keep the name.

hadarshjfrog, line 16: Also - why did you switch the order of the args?

formats/simplejsonapi.go
```diff
@@ -96,6 +96,7 @@ type CveRow struct {
 type Applicability struct {
 	Status             string     `json:"status"`
 	ScannerDescription string     `json:"scannerDescription,omitempty"`
+	UndeterminedReason string     `json:"undeterminedReason,omitempty"`
```
hadarshjfrog, line 99: where do we use it? `UndeterminedReason` is being inserted into - but not called...

formats/sarifutils/test_sarifutils.go
```diff
@@ -13,14 +13,16 @@ func CreateRunWithDummyResults(results ...*sarif.Result) *sarif.Run {
...
 	run := sarif.NewRunWithInformationURI("", "")
 	if result.RuleID != nil {
 		run.AddRule(*result.RuleID)
 	}
 	run.AddResult(result)
-	run.Tool.Driver.Rules[0].Properties = make(sarif.Properties)
-	run.Tool.Driver.Rules[0].Properties[property] = value
+	run.Tool.Driver.Rules[0].Properties = make(sarif.Properties, len(properties))
+	for index, _ := range properties {
+		run.Tool.Driver.Rules[0].Properties[properties[index]] = values[index]
```
hadarshjfrog, line 24: suggest makeing sure that len(values)==len(properties)

formats/sarifutils/sarifutils.go
barv-jfrog: We can use it for applicability status as it is also received in the Rule property. Do you want me to change applicability fetch to this function ?

hadarshjfrog: yes - otherwise the usage of the this generic function just one misses the point IMO

formats/simplejsonapi.go
barv-jfrog: It is only in simple-json, and it is under CveRow (CveRow has Applicability struct as one of its components, it gets printed automatically)

formats/sarifutils/test_sarifutils.go
barv-jfrog: ok fixed

barv-jfrog: I thought it was nicer to have the arrays at the end of the function, what do you think?

formats/sarifutils/test_sarifutils.go
barv-jfrog: right, fixed

formats/sarifutils/sarifutils.go
attiasas: Yes please do use the new `GetRuleProperty` when applicable. but the `GetRuleUndeterminedReason` does not belong here. please move it to `jasutils` package.
This files is only for utils to access, create or set values in sarif

formats/sarifutils/test_sarifutils.go
```diff
@@ -13,14 +15,19 @@ func CreateRunWithDummyResults(results ...*sarif.Result) *sarif.Run {
...
+	}
 	run := sarif.NewRunWithInformationURI("", "")
 	if result.RuleID != nil {
 		run.AddRule(*result.RuleID)
 	}
 	run.AddResult(result)
-	run.Tool.Driver.Rules[0].Properties = make(sarif.Properties)
-	run.Tool.Driver.Rules[0].Properties[property] = value
+	run.Tool.Driver.Rules[0].Properties = make(sarif.Properties, len(properties))
+	for index, _ := range properties {
```
attiasas, line 28:
```suggestion
	for index := range properties {
```
simplify expression

attiasas (commented):
LGTM, please check out my comments

attiasas (approved):

PR #157, opened 2024-08-25
Fix build scan command table printing with --vuln=true

dortam888:

-----

Fixing issues when using --vuln=true flag and --format=table (default) in build scan. 
Issue 1: build scan printed vulnerability table twice instead of printing the violation table and the vulnerabilities table.
Issue 2: The vulnerability table is messed up, not sorted, and shows the same vulnerability multiple times.


commands/scan/buildscan.go
```diff
@@ -155,7 +155,6 @@ func (bsc *BuildScanCommand) runBuildScanAndPrintResults(xrayManager *xray.XrayS
 
 	resultsPrinter := utils.NewResultsWriter(scanResults).
 		SetOutputFormat(bsc.outputFormat).
-		SetIncludeVulnerabilities(bsc.includeVulnerabilities).
 		SetIncludeLicenses(false).
```
hadarshjfrog, line 158: after removing this line - did you test the CLI behaviour with and without watches?

dortam888: @hadarshjfrog I did the regular tests (also with watches).
This line doesn't affect watches as it relates to vulnerabilities. (and watches are affecting violations).
We call SetIncludeVulnerabilities in line 179 if vulnerabilities should be added:
`		if bsc.includeVulnerabilities {
			resultsPrinter.**SetIncludeVulnerabilities**(true)
			if err = resultsPrinter.PrintScanResults(); err != nil {
				return false, err
			}
		}`

attiasas (changes requested):
This change is not how to fix the issue. 
There are 2 issues here:
1. we can see that in the parsing of the flags, we don't take into account the `project` flag only in build-scan. the fix should add it, instead of removing and always printing violations...
image
image

3. In the table printing code, we do not expect to print both violations and vulnerabilities tables so we need to choose one (preferable vuln if requested)
image

dortam888:
@attiasas @hadarshjfrog We have a couple of issues with jf bs.
Some parameters are missing, like projects and JAS enablement so we don't get results for builds from projects and we don't get JAS results at all (we don't have secrets but we can have CA).
Regarding this fix, See the comment in the code:
https://github.com/jfrog/jfrog-cli-security/blob/main/commands/scan/buildscan.go#L171
The behavior when the format is json for example is appending vulnerabilities to violations and that is what the code meant to do for tables as well based on the comment.
[Jira](https://jfrog-int.atlassian.net/browse/XRAY-72561) is about having violations and no vulnerabilities try to get vulnerabilites.
The build is failing because we have violations but we're not showing it in tables printing.

attiasas:
fixed in:
https://github.com/jfrog/jfrog-cli-security/pull/163
https://github.com/jfrog/jfrog-cli-security/pull/164

PR #165, opened 2024-09-03
Add a warning to indicate wrong usage in 'jf scan' command

eranturgeman:

-----
The 'jf scan' command should be executed in the following manner: jf scan [command options] <source pattern> OR jf scan [command options] --spec=<value>.
If the command was written in a different order: jf scan <source pattern> [command options], it may lead that some flags are mistaken as arguments and are not going through the command's flags verifications. Therefore, if a wrong flag was provided in this manner - it will get skipped completely without informing the customer.
I added a warning to indicate the wrong usage and what it may cause.
The indication comes in a Warning form and not error in order to not breaking customers.

### DIFF
--- a/cli/scancommands.go
+++ b/cli/scancommands.go
@@ -197,6 +197,13 @@ func ScanCmd(c *components.Context) error {
 	if len(c.Arguments) == 0 && !c.IsFlagSet(flags.SpecFlag) {
 		return pluginsCommon.PrintHelpAndReturnError("providing either a <source pattern> argument or the 'spec' option is mandatory", c)
 	}
+
+	if len(c.Arguments) > 1 {
+		log.Warn(fmt.Sprintf("Too many arguments provided (%d in total).\n"+
+			"Some flags might be mistaken as arguments and ignored, which could prevent them from affecting the command as intended.\n"+
+			"Please ensure all provided flags are valid and provided correct order. For more details: 'jf %s --help'", len(c.Arguments), c.CommandName))
+	}
+
 	serverDetails, err := createServerDetailsWithConfigOffer(c)
 	if err != nil {
 		return err


utils/utils.go
```diff
@@ -83,3 +83,12 @@ func splitEnvVar(envVar string) (key, value string) {
...
 	return split[0], strings.Join(split[1:], "=")
 }
+
+// This is a general error message for the CLI commands.
+// Due to the way we parse the commands, in some cases flags might be mistaken as arguments and therefore they will not go though the command's flags verifications.
+// When this happens it could prevent the flag from affecting the command as intended. This error message informes the user about this issue
+func GetCliTooManyArgsErrorMessage(numberOfArguments int, commandName string) string {
+	return fmt.Sprintf("Too many arguments provided (%d in total).\n"+
+		"Some flags might be mistaken as arguments and ignored, which could prevent them from affecting the command as intended.\n"+
+		"Please ensure all provided flags are valid and provided correct order. For more details: 'jf %s --help'", numberOfArguments, commandName)
```
attiasas, line 93:
```suggestion
	return fmt.Sprintf("Too many arguments provided (%d in total).\nSome flags might be mistaken as arguments and ignored, which could prevent them from affecting the command as intended.\nPlease ensure all provided flags are valid and provided correct order. For more details: 'jf %s --help'", numberOfArguments, commandName)
```
If you are using `fmt.Sprintf` no need for `+` also

cli/scancommands.go
```diff
@@ -196,7 +196,10 @@ func EnrichCmd(c *components.Context) error {
 func ScanCmd(c *components.Context) error {
 	if len(c.Arguments) == 0 && !c.IsFlagSet(flags.SpecFlag) {
 		return pluginsCommon.PrintHelpAndReturnError("providing either a <source pattern> argument or the 'spec' option is mandatory", c)
+	} else if len(c.Arguments) > 1 {
+		log.Warn(utils.GetCliTooManyArgsErrorMessage(len(c.Arguments), c.CommandName))
 	}
```
attiasas, line 201:
```suggestion
	} else if len(c.Arguments) > 1 {
		log.Warn(utils.GetCliTooManyArgsErrorMessage(len(c.Arguments), c.CommandName))
		return pluginsCommon.PrintHelpAndReturnError(c)
	}
```
If you think we should stop the command instead of just logging you can use the above (and remove the last part of the log about running the --help since it will show the output)

attiasas (commented):
Nice, take a look at my comments

utils/utils.go
```diff
@@ -119,3 +119,10 @@ func splitEnvVar(envVar string) (key, value string) {
 	}
 	return split[0], strings.Join(split[1:], "=")
 }
+
+// This is a general error message for the CLI commands.
+// Because of how command parsing is handled, improperly specified flags may be misinterpreted as arguments.
+// Therefore, these flags will not go through the command's flags verifications, and will not be caught as incorrect flags.
+func GetCliTooManyArgsErrorMessage(numberOfArguments int) string {
+	return fmt.Sprintf("Too many arguments provided (%d in total).\nSome flags may be incorrectly specified, causing them to be misinterpreted as arguments and ignored. Please verify that all flags are valid.", numberOfArguments)
```
hadarshjfrog, line 127: should we say using "--help" ?

cli/scancommands.go
```diff
@@ -196,7 +196,13 @@ func EnrichCmd(c *components.Context) error {
 func ScanCmd(c *components.Context) error {
```
hadarshjfrog, line 196: does docker scan uses the same flow as `ScanCmd`? or should implement it there too?

hadarshjfrog:
Please add integration tests you think might simulate it - with the common mistakes we have seen (flags at the end instead of beginning, `watch`/`license` instead of `watches`/`licenses`

utils/utils.go
eranturgeman: I think not since they can use both options. if any of the options is provided with wrong flags the execution will fail

cli/scancommands.go
eranturgeman: No, it is using dockerCmd that resides in CLI. I initiated another PR to ecosystem:
https://github.com/jfrog/jfrog-cli/pull/2686

utils/utils.go
attiasas, line 127: should we indicate why the flags are incorrectly specified? Maybe add a hint about moving flags before args?

cli/scancommands.go
eranturgeman: @hadarshjfrog Since resolving the same issue in 'jf docker scan' should be done by Ecosystem and this is not a top priority for them. Since this is not going to happen soon and I dont want to miss the fix Iv'e already worked on im proceeding without 'jf docker scan' and we can open a ticket for the Ecosystem about this

utils/utils.go
eranturgeman: @attiasas We can see the correct way to use this through --help.
The issue is that we construct the command in a different way than the common way to do so.
Instead of putting args and then flags we expect the opposite.
So there is no hint to add since its already specified in --help and also- it is just a suggestion. There could be other reasons for this kind of error so I just added a general suggestion and not specific instructions

cli/scancommands.go
eranturgeman: I opened a bug for Infra group to fix jf docker scan

scans_test.go
```diff
@@ -76,6 +76,44 @@ func TestXrayBinaryScanSimpleJsonWithProgress(t *testing.T) {
...
+		},
+		{
+			name:            "args before flags",
+			flagsBeforeArgs: false,
+		},
+	}
+
+	callback := commonTests.MockProgressInitialization()
+	defer callback()
+	integration.InitScanTest(t, scangraph.GraphScanMinXrayVersion)
```
attiasas, line 99: this should be the first line, skipping the test if needed

scans_test.go
```diff
@@ -76,6 +76,44 @@ func TestXrayBinaryScanSimpleJsonWithProgress(t *testing.T) {
...
+		t.Run(test.name, func(t *testing.T) {
+			var args []string
+			if test.flagsBeforeArgs {
+				args = []string{"scan", "--watch=my-watch", binariesPath}
+			} else {
+				args = []string{"scan", binariesPath, "--watch=my-watch"}
+			}
+
+			err := securityTests.PlatformCli.Exec(args...)
+			assert.Error(t, err)
```
attiasas, line 112:
```suggestion
			assert.Error(t, securityTests.PlatformCli.Exec(args...))
```
maybe you can also check if its the error you added (or error expected) and not random one

scans_test.go
```diff
@@ -76,6 +76,44 @@ func TestXrayBinaryScanSimpleJsonWithProgress(t *testing.T) {
...
+	integration.InitScanTest(t, scangraph.GraphScanMinXrayVersion)
+	binariesPath := filepath.Join(filepath.FromSlash(securityTests.GetTestResourcesPath()), "projects", "binaries", "*")
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			var args []string
+			if test.flagsBeforeArgs {
+				args = []string{"scan", "--watch=my-watch", binariesPath}
+			} else {
+				args = []string{"scan", binariesPath, "--watch=my-watch"}
```
attiasas, line 108:
```suggestion
				args = []string{"scan", "--not-defined-flag=value", binariesPath}
			} else {
				args = []string{"scan", binariesPath, "--watches=my-watch"}
```
change the flag name to make it more readable

attiasas (changes requested):
LGTM, make sure the PR description is formated

PR #166, opened 2024-09-05
Disable generating SARIF for GitHub advance security tab if not entitled for JAS

attiasas:

-----

disable generating SARIF for GitHub advance security tab if not entitled for JAS

### DIFF
--- a/utils/securityJobSummary.go
+++ b/utils/securityJobSummary.go
@@ -178,6 +178,10 @@ func RecordSarifOutput(cmdResults *Results) (err error) {
 	if err != nil || manager == nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


utils/securityJobSummary.go
```diff
@@ -178,6 +178,10 @@ func RecordSarifOutput(cmdResults *Results) (err error) {
 	if err != nil || manager == nil {
 		return
 	}
+	if cmdResults.ExtendedScanResults == nil || !cmdResults.ExtendedScanResults.EntitledForJas {
```
hadarshjfrog, line 181: do we have an integration test to cover it? 
maybe an entitlement mock that'll return false - and making sure that in this case - there's no sarif?

attiasas: added

hadarshjfrog (approved):

PR #175, opened 2024-09-11
Improved tech detection when tech is provided by user

eranturgeman:

-----
This PR addresses an issue with incorrect technology detection in Yarn projects, particularly when there are multiple sub-projects that only contain a Yarn descriptor without any specific Yarn indicators. In a multi-project (but not multi-module) Yarn environment, if the inner projects only include the package.json file without any of Yarn’s standard indicators (.yarnrc.yml, yarn.lock, .yarn, “yarnrc”), the auto-detection mechanism incorrectly identifies these as NPM projects.

Since package.json cannot be used as an indicator for both NPM and Yarn to avoid conflicts, I implemented the following logic:
 **If the user specifies a particular technology (via command flags or the install command [Frogbot]), we take that into account when verifying the technology. This allows us to verify the specified technology using both its descriptors and indicators and not just its indicators.**

As a result, when a ‘yarn install’ command is provided for a project and a package.json file is present, we can confidently determine that Yarn is being used, not NPM.

Another test case needs to be added to TestGetTechInformationFromWorkingDir. This test case discovered an issue that is fixed this PR: https://github.com/jfrog/jfrog-cli-security/pull/179
The test case will be added along with the fix.


utils/techutils/techutils.go
```diff
@@ -400,13 +400,19 @@ func isExclude(path string, techData TechData) bool {
...
 	// Get the relevant technologies to check
 	technologies := requestedTechs
+	var techProvidedByUser bool
 	if len(technologies) == 0 {
 		technologies = GetAllTechnologiesList()
+	} else {
+		// If the project's technology was provided by the user, and isn't detected by us, we want to enable capturing the technology by its descriptor as well as by its indicators.
+		// In case we execute our auto-detection we want to avoid that since it may lead to collisions between package managers with the same descriptor (like Npm and Yarn)
+		techProvidedByUser = true
+		log.Debug(fmt.Sprintf("An install command was provided, so the technology will be identified based on the given command. Detected technologies: %s", technologies))
```
attiasas, line 410: tech provided not only by "install command" it could also be using a tech flag (not frog Frogbot). actually there is no mentioned in the code about the install command, only by requested tech.
fix the debug log and change the PR title to reflect the actual change

utils/techutils/techutils_test.go
```diff
@@ -484,19 +507,33 @@ func TestGetTechInformationFromWorkingDir(t *testing.T) {
...
+			name:                 "yarnTestWithProvidedTechFromUser",
+			tech:                 Yarn,
+			requestedDescriptors: make(map[Technology][]string),
+			techProvidedByUser:   true,
+			expected: map[string][]string{
+				"dir4": {filepath.Join("dir4", "package.json")},
+				"dir3": {filepath.Join("dir3", "package.json")},
+				"dir":  {filepath.Join("dir", "package.json")},
+			},
+		},
```
attiasas, line 531: add another tests with `techProvidedByUser=true` (i.e. tech requested by the user), not for Yarn to make sure it does not break other tech when requested

attiasas (changes requested):
Nice!, take a look at my comments.

attiasas (approved):

PR #179, opened 2024-09-15
Fixed wrong root identification in module detection

eranturgeman:

-----
In getExistingRootDir we find the highest common ancestor for every descriptor, and refer this descriptor to this common ancestor rather then its original working dir.
As part of the process we check under some condition, if the current wd is a prefix to the current root, and if so - wd is the new root.
We encountered a wrong identification in the following scenario: if wd is indeed a prefix to root but **not a path prefix**, we wrongfully take wd as the new root.
Example:
root = "dir1/dir2"
wd = "dir"

wd is indeed a prefix to root (by characters) but is not actually a path prefix since the path prefix to root is "dir1" and not "dir"

This PR adds fix for this issue so we identify if wd is a **path prefix** to the root

Also, this PR added a test case to TestGetTechInformationFromWorkingDir, that have failed due to the build, and is now added after the fix

### DIFF
--- a/utils/techutils/techutils.go
+++ b/utils/techutils/techutils.go
@@ -510,9 +510,14 @@ func hasCompletePathPrefix(root, wd string) bool {
 	if !strings.HasPrefix(root, wd) {
 		return false
 	}
+	log.Info(fmt.Sprintf("ERAN TEST - root path: %s", root))
+	log.Info(fmt.Sprintf("ERAN TEST - wd path: %s", root))
 	rootParts := strings.Split(root, string(filepath.Separator))
 	wdParts := strings.Split(wd, string(filepath.Separator))
+	log.Info(fmt.Sprintf("ERAN TEST - root slice: %s", rootParts))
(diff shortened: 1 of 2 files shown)


utils/techutils/techutils.go
```diff
@@ -485,13 +485,24 @@ func getExistingRootDir(path string, workingDirectoryToIndicators map[string][]s
...
 }
 
+func hasCompletePathPrefix(root, wd string) bool {
+	if strings.HasPrefix(root, wd) {
+		amountOfSkippedChars := len(wd)
+		partialRoot := root[amountOfSkippedChars:]
+		if partialRoot == "" || partialRoot[0] == filepath.Separator {
+			return true
+		}
+	}
```
attiasas, line 502:
```suggestion
	if !strings.HasPrefix(root, wd)  {
	        return false
        }
        lastDirWD := filepath.Base(wd)
	matchingDirRoot := filepath.Base(root[:len(wd)])
	return lastDirWD == matchingDirRoot
```

attiasas (commented):
LGTM

attiasas (approved):

PR #189, opened 2024-09-23
Add logs and add error if serverDetails not provided

attiasas:

-----

* Refactor JAS run task to a separate function.
* Added some logs to report when not entitled / not provided details
* Return Error if server details are nil at the point of creating the JAS scanner (it should already exists since we ended entitlement request successfully

### DIFF
--- a/commands/audit/audit.go
+++ b/commands/audit/audit.go
@@ -3,8 +3,9 @@ package audit
 import (
 	"errors"
 	"fmt"
+
+	"github.com/jfrog/gofrog/log"
 	jfrogappsconfig "github.com/jfrog/jfrog-apps-config/go"
-	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
 	"github.com/jfrog/jfrog-cli-security/jas"
 	"github.com/jfrog/jfrog-cli-security/jas/applicability"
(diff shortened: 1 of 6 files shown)


commands/audit/audit.go
```diff
@@ -244,19 +238,41 @@ func isEntitledForJas(xrayManager *xray.XrayServicesManager, auditParams *AuditP
...
+	serverDetails, err := auditParams.ServerDetails()
+	if err != nil {
+		err = fmt.Errorf("failed to get server details: %s", err.Error())
+		return
+	}
+	jasScanner, err = jas.CreateJasScanner(jfrogAppsConfig, serverDetails, jas.GetAnalyzerManagerXscEnvVars(auditParams.commonGraphScanParams.MultiScanId, results.ExtendedScanResults.SecretValidation, results.GetScaScannedTechnologies()...), auditParams.Exclusions()...)
+	if err != nil {
+		err = fmt.Errorf("failed to create jas scanner: %s", err.Error())
+		return
+	} else if jasScanner == nil {
```
eranturgeman, line 255: Under which circumstances that weren't exist before the jasScanner can be nil after we tried to create it? If it is not trivial to infer - please add a comment what can result this

jas/runner/jasrunner.go
```diff
@@ -20,11 +19,10 @@ import (
...
 )
 
-func AddJasScannersTasks(securityParallelRunner *utils.SecurityParallelRunner, scanResults *utils.Results, directDependencies *[]string,
-	serverDetails *config.ServerDetails, thirdPartyApplicabilityScan bool, scanner *jas.JasScanner, scanType applicability.ApplicabilityScanType,
+func AddJasScannersTasks(securityParallelRunner *utils.SecurityParallelRunner, scanResults *utils.Results, directDependencies *[]string, thirdPartyApplicabilityScan bool, scanner *jas.JasScanner, scanType applicability.ApplicabilityScanType,
 	secretsScanType secrets.SecretsScanType, errHandlerFunc func(error), scansToPreform []utils.SubScanType, configProfile *services.ConfigProfile, scansOutputDir string) (err error) {
-	if serverDetails == nil || len(serverDetails.Url) == 0 {
-		log.Warn("To incorporate the ‘Advanced Security’ scans into the audit output make sure platform url is provided and valid (run 'jf c add' prior to 'jf audit' via CLI, or provide JF_URL via Frogbot)")
+	// Set the analyzer manager executable path.
+	if scanner.AnalyzerManager.AnalyzerManagerFullPath, err = jas.GetAnalyzerManagerExecutable(); err != nil {
```
eranturgeman, line 25: why are we doing this here? what is changed in the logic that we didn't need it before and now we do?

eranturgeman (commented):
I reviewed the changes, there are few things that are unclear to me (see my comments)
Also, about the serverDetails logs, do you think it would suffice to understand the issue with the POC? We know already they dont have the serverDetails or the ServerUrl at this point. In my PR I wanted to understand what they do have (xrayUrl, artifactoryUrl..) or if they lose this data during the process

eranturgeman (approved):

commands/audit/audit.go
attiasas: look at `CreateScanner`. if `len(serverDetails.Url) == 0` is true we log the usual warning and return nil scanner, we should not continue after that just like the check was at `JasRunner`. we need a way to separate the case that serverDetails is nil (err, should not be nil at this point) or the user did not configure the right things (URL, warning)

PR #96, opened 2024-06-30
Refactor commands output formats

attiasas:
- [x] The pull request is targeting the `dev` branch.
- [x] The code has been validated to compile successfully by running `go vet ./...`.
- [x] The code has been formatted properly using `go fmt ./...`.
- [x] All [static analysis checks](https://github.com/jfrog/jfrog-cli-security/actions/workflows/analysis.yml) passed.
- [x] All [tests](https://github.com/jfrog/jfrog-cli-security/actions/workflows/test.yml) have passed. If this feature is not already covered by the tests, new tests have been added.
- [x] All changes are detailed at the description. if not already covered at [JFrog Documentation](https://github.com/jfrog/documentation), new documentation have been added.

-----

### Move all `Results` and outputs to their inner package, organize currently available code and renames to fit better.
<details>
<summary>Before:</summary>

image

</details>

<details>
<summary>After:</summary>

image

</details>

### Move detection out of SCA to run first thing in audit
* Allows to sync information from local `jfrog-apps-config`.
* Prepare to remove config detection to separate command.
* Prepare to remove dependency collection to separate command.

### Change `Results` object to `ScanCommandResults`
* Prepare the infrastructure to separate different sca children-targets in a single scan target 
image

###  Separate converting `ScanCommandResults` to different output formats and printing the output.
* `output` package - Write the results in a specific format as an output string.
* `conversion` package - Parse the `ScanCommandResults` object to convert it to different formats that we are supporting. 

### Make implementing a new output format (by introducing an interface) easier, and maintain the currently supported.
* Reduce code duplications
* Easy to see what we support
* Easy to implement new formats / integrate new scans and results
image

### General validation improvments for results structs
* Easy to support and maintain
* Adds more tests
* Exact/Partial match options (for unit + integration tests)
* Easy to read what is tested
* Refactor validations to its own package
image

PR #200, opened 2024-10-06
Enable allow-partial-results for Audit SCA scan and dep tree construction

eranturgeman:

-----

This PR introduces a new ability that allows partial scan results without failing the entire scan over a single failing part in the scans (for example when we have multiple working directories and the scan on one of them have failed - we will not fail the entire scan, but rather log the error of the failing scan and continue).
This ability **should be applied for SCA scans as well as for JAS scans** but currently, until the security cli refactor (https://github.com/jfrog/jfrog-cli-security/pull/96) will be merged it is applied **only for SCA scans**
Applying this ability to JAS scans will be on a different PR

In addition, a new flag was added to enable using this ability from the cli: **--allow-partial-results**

This PR contains also 2 bug fixes:
1. In executeScaScanTask, ScaScansWg marked as Done in defer, which enables another process to close the SecurityParallelRunner.ErrorsQueue. Since the queue closing is performed by another thread, unrelated to the thread performing the scan, the queue sometimes gets closed before the error handling function passed along with executeScaScanTask to AddTaskWithError start executing. In case we have an error, the addition of this error to the SecurityParallelRunner.ErrorsQueue might occur after the queue is closed, which leads to a panic. The suggested fix is to close the queue only when there is no error, and if there is- the queue will be closed in the error handling function, after the error is added
2. In GetScaScannedTechnologies we create a new set and insert the technologies from the current scan results only. I added the ability to pass already existing slice of techutils.Technology to enable preserving technologies from other scans if provided and to include them in the overall list of technologies (this is required for Frogbot, when multiple working directories are being scanned, but get all fixed together after all the scans. In this scenario the only remaining techs before the fix are the technologies of the last performed scan)

**Note:** Some of the NuGet Audit tests fail. This issue is known and will be fixed soon


cli/docs/flags.go
```diff
@@ -229,8 +230,9 @@ var flagsMap = map[string]components.Flag{
 		"Set to false if you wish to not use the gradle or maven wrapper.",
 		components.WithBoolDefaultValue(true),
 	),
-	WorkingDirs: components.NewStringFlag(WorkingDirs, "A comma-separated list of relative working directories, to determine audit targets locations."),
-	OutputDir:   components.NewStringFlag(OutputDir, "Target directory to save partial results to.", components.SetHiddenStrFlag()),
+	WorkingDirs:         components.NewStringFlag(WorkingDirs, "A comma-separated list of relative working directories, to determine audit targets locations."),
+	OutputDir:           components.NewStringFlag(OutputDir, "Target directory to save partial results to.", components.SetHiddenStrFlag()),
+	AllowPartialResults: components.NewBoolFlag(AllowPartialResults, "Set to true to allow partial results and continuance of the scan in case of certain errors."),
```
attiasas, line 235:
```suggestion
	AllowPartialResults: components.NewBoolFlag(AllowPartialResults, "Set to true to allow partial results and continuance of the scan in case of certain errors.", components.SetHiddenBoolFlag()),
```
Set it hidden for now

commands/audit/audit.go
```diff
@@ -202,7 +202,7 @@ func RunAudit(auditParams *AuditParams) (results *utils.Results, err error) {
 	}
 	// The sca scan doesn't require the analyzer manager, so it can run separately from the analyzer manager download routine.
 	if scaScanErr := buildDepTreeAndRunScaScan(auditParallelRunner, auditParams, results); scaScanErr != nil {
-		auditParallelRunner.AddErrorToChan(scaScanErr)
+		_ = createErrorIfPartialResultsDisabled(auditParams, auditParallelRunner, fmt.Sprintf("An error has occurred during SCA scan process. SCA scan is skipped for the following directories: %s.", auditParams.workingDirs), scaScanErr)
```
attiasas, line 205:
```suggestion
		err = createErrorIfPartialResultsDisabled(auditParams, auditParallelRunner, "An error has occurred during SCA scan process.", scaScanErr)
```
no need to specify the working dirs here, the error for the specific scan will handle adding its information to the error. In addition, you are not doing anything with the returned `error`, you should handle it or change the method

commands/audit/audit.go
```diff
@@ -277,3 +277,26 @@ func downloadAnalyzerManagerAndRunScanners(auditParallelRunner *utils.SecurityPa
 	}
 	return
 }
+
+// This function checks if partial results is allowed. If so we log the error and continue.
+// If partial results is not allowed we add the error to the SecurityParallelRunner error's channel if one is provided, or simply return the error if not.
+func createErrorIfPartialResultsDisabled(auditParams *AuditParams, auditParallelRunner *utils.SecurityParallelRunner, extraMassageForLog string, err error) error {
```
attiasas, line 283:
```suggestion
func createErrorIfPartialResultsDisabled(partialResultEnabled bool, auditParallelRunner *utils.SecurityParallelRunner, extraMassageForLog string, err error) error {
```
IMO, no need to pass `auditParams *AuditParams`, just the bool is ok

commands/audit/audit_test.go
```diff
@@ -197,3 +200,144 @@ func searchForStrWithSubString(t *testing.T, filesList []string, subString strin
...
+			useJas:              false,
+			testDirPath:         filepath.Join("..", "..", "tests", "testdata", "projects", "package-managers", "npm", "npm-un-installable"),
+		},
+		{
+			name:                "Skip failure in SCA during scan itself",
+			allowPartialResults: true,
+			useJas:              false,
+			testDirPath:         filepath.Join("..", "..", "tests", "testdata", "projects", "package-managers", "npm", "npm-project"),
+		},
+		// TODO when applying allow-partial-results to JAS make sure to add a test case that checks failures in JAS scans + add  some JAS api call to the mock server
```
attiasas, line 235: ?

commands/audit/scarunner.go
```diff
@@ -76,13 +76,14 @@ func buildDepTreeAndRunScaScan(auditParallelRunner *utils.SecurityParallelRunner
...
 		if bdtErr != nil {
-			err = errors.Join(err, fmt.Errorf("audit command in '%s' failed:\n%s", scan.Target, bdtErr.Error()))
+			err = errors.Join(err, createErrorIfPartialResultsDisabled(auditParams, nil, fmt.Sprintf("Dependencies tree construction ha failed for the following target: %s", scan.Target), fmt.Errorf("audit command in '%s' failed:\n%s", scan.Target, bdtErr.Error())))
 			continue
 		}
 		// Create sca scan task
 		auditParallelRunner.ScaScansWg.Add(1)
 		_, taskErr := auditParallelRunner.Runner.AddTaskWithError(executeScaScanTask(auditParallelRunner, serverDetails, auditParams, scan, treeResult), func(err error) {
-			auditParallelRunner.AddErrorToChan(fmt.Errorf("audit command in '%s' failed:\n%s", scan.Target, err.Error()))
+			_ = createErrorIfPartialResultsDisabled(auditParams, auditParallelRunner, fmt.Sprintf("Failed to execute SCA scan for the following target: %s", scan.Target), fmt.Errorf("audit command in '%s' failed:\n%s", scan.Target, err.Error()))
```
attiasas, line 85:
```suggestion
			err = createErrorIfPartialResultsDisabled(auditParams, auditParallelRunner, fmt.Sprintf("Failed to execute SCA scan for the following target: %s", scan.Target), fmt.Errorf("audit command in '%s' failed:\n%s", scan.Target, err.Error()))
```
error not handled, there is 2 times fmt with `scan.Target`, make sure error looks readable

commands/audit/scarunner.go
```diff
@@ -141,8 +142,12 @@ func executeScaScanTask(auditParallelRunner *utils.SecurityParallelRunner, serve
 	scan *xrayutils.ScaScanResult, treeResult *DependencyTreeResult) parallel.TaskFunc {
 	return func(threadId int) (err error) {
 		log.Info(clientutils.GetLogMsgPrefix(threadId, false)+"Running SCA scan for", scan.Target, "vulnerable dependencies in", scan.Target, "directory...")
+		var xrayErr error
 		defer func() {
-			auditParallelRunner.ScaScansWg.Done()
+			if xrayErr == nil {
+				// We Sca waitGroup as done only when we have no errors. If we have errors we mark it done in the error's handler function
+				auditParallelRunner.ScaScansWg.Done()
+			}
```
attiasas, line 150: `xrayErr` not set at all, will always be `nil`

tests/testdata/projects/package-managers/npm/npm-un-installable/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
attiasas, line 1: what is this file?

attiasas (changes requested):
Nice, take a look at my comments

commands/audit/audit.go
eranturgeman: I can remove the working dirs from the error message
But it doesn't make since to return the error since we are doing nothing with this error. with AddErrorToChan we just added the error to the runner errors queue and this is what is done here as well. To avoid code duplication with writing the exact same logic twice - one with returning the error and one without I put it all into a single function and explained exactly how it works:
if a parallelRunner if provided the error is getting added to the error queue, if not - the error is returned.
In this case we only add it to the queue and it is later being handled
@attiasas

commands/audit/scarunner.go
eranturgeman: This is not really twice. createErrorIfPartialResultsDisabled works like this:
if allowPartialResults == true we log ONLY the message in the third argument (fmt.Sprintf("Failed to execute SCA scan for the following target: %s", scan.Target))
if allowPartialResults == false we use ONLY the error provided (fmt.Errorf("audit command in '%s' failed:\n%s", scan.Target, err.Error()) ). The error I provided is the exact same one we used to add to the error chain so there is not change in behaviour if allowPartialResults==false

eranturgeman: as for the unhandled error - this is suppose to be this way with a parallel runner is provided and there is nothing to do with the returned error

commands/audit/scarunner.go
eranturgeman: it is set one step after, during runScaWithTech
since my code is in defer it will take the value after the error is set. Im I wrong?

attiasas (approved):

tests/testdata/projects/package-managers/npm/npm-un-installable/package.json
eranturgeman: this is for a new test case, used in TestAuditWithPartialResults in audit_test.go

PR #205, opened 2024-10-14
Audit - Do not copy Visual Studio inner folder when building dependency tree

kerenr-jfrog:

-----

- During the dependency tree building process in some package managers (pnpm, Python, and NuGet), all project files and directories are copied to a temporary directory. When using the Visual Studio IDE, this can create a race condition for files in the inner  `.vs ` folder. To avoid this conflict, we should exclude this folder when building the dependency tree.


commands/audit/sca/common_test.go
```diff
@@ -36,7 +36,7 @@ func TestGetExcludePattern(t *testing.T) {
 		{
 			name:     "Test no exclude pattern recursive",
 			params:   func() *utils.AuditBasicParams { return (&utils.AuditBasicParams{}).SetIsRecursiveScan(true) },
-			expected: "(^.*\\.git.*$)|(^.*node_modules.*$)|(^.*target.*$)|(^.*venv.*$)|(^.*test.*$)",
+			expected: "(^.*\\.git.*$)|(^.*node_modules.*$)|(^.*target.*$)|(^.*venv.*$)|(^.*test.*$)|(^.*\\.vs.*$)",
```
attiasas, line 39: I would not change the default params for the risk of missing info in `.vscode`.
I would instead just pass this param in the VS extension with the added `.vs` so it will only effect VS

(Same for all other changes here to the default params

commands/audit/sca/nuget/nuget.go
```diff
@@ -296,3 +303,28 @@ func parseNugetDependencyTree(buildInfo *entities.BuildInfo) (nodes []*xrayUtils
 	allUniqueDeps = uniqueDepsSet.ToSlice()
 	return
 }
+
+func getExcludedFoldersNames(wd string, exclusionPattern string) ([]string, error) {
```
attiasas, line 307: add unit tests for this method

attiasas (dismissed):
Nice, please change the PR name to fit the changes. something like 'Audit - Dont copy files in exclusions params for Nuget projects`

eranturgeman:
Looks good!
One question though - is this issue relevant only for NuGet?
The copy to tmp dir for performing install might occur in several package managers (for example in pnpm). Do we have the same issue there? if so we need to consider fixing it across all PMs

commands/audit/sca/nuget/nuget_test.go
```diff
@@ -146,3 +148,53 @@ func TestRunDotnetRestoreAndLoadSolution(t *testing.T) {
...
+			wd:               dotnetDir,
+			exclusionPattern: "(^.*node.*$)",
+			expectedFolders:  nil,
+			expectedError:    nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
```
eranturgeman, line 186: delete new line

commands/audit/sca/nuget/nuget_test.go
```diff
@@ -146,3 +148,53 @@ func TestRunDotnetRestoreAndLoadSolution(t *testing.T) {
...
+				assert.Error(t, err)
+				assert.Contains(t, err.Error(), tt.expectedError.Error())
+			} else {
+				assert.NoError(t, err)
+			}
+
+			assert.Equal(t, got, tt.expectedFolders)
+		})
+	}
+
```
eranturgeman, line 199: delete new line

eranturgeman (changes requested):
lets re-review after the changes we discussed about

commands/audit/sca/common.go
```diff
@@ -21,6 +21,11 @@ import (
 	xrayUtils "github.com/jfrog/jfrog-client-go/xray/services/utils"
 )
 
+const (
+	// Visual Studio inner directory to exclude when copying the project to a temporary directory during the dependency tree build process.
+	ExcludeForCopyVS = ".vs"
```
eranturgeman, line 26:
```suggestion
	DotVsRepoSuffix = ".vs"
```
The const name should indicate about what is the const, not what it is used for. Think that we might use it for other purposes

commands/audit/sca/common.go
```diff
@@ -21,6 +21,11 @@ import (
 	xrayUtils "github.com/jfrog/jfrog-client-go/xray/services/utils"
 )
 
+const (
+	// Visual Studio inner directory to exclude when copying the project to a temporary directory during the dependency tree build process.
```
eranturgeman, line 25:
```suggestion
```
When writing an explanation above a const we want to explain what the const is, and not what it is used for at definition, since for a developer that is not familiar with the context it will be hard to understand what you ment.
We want to explain about a "strange" behaviour or something that needs extra details at the stop where it occurs

commands/audit/sca/nuget/nuget.go
```diff
@@ -65,7 +65,9 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTree []*xrayUtils.
 		err = errors.Join(err, fileutils.RemoveTempDir(tmpWd))
 	}()
 
-	err = biutils.CopyDir(wd, tmpWd, true, nil)
+	// Specifies files and direcrories to exclude when copying the project to a temporary directory.
+	excludeForCopy := []string{sca.ExcludeForCopyVS}
```
eranturgeman, line 69:
```suggestion
```
Nice thinking to put it in a var so you can explain what you did, but it is redundant to create a new var for a non-complex expression that we use only once, just for the sake of explaining what you have done.
You can just explain about the exclusion about the CopyDir line

commands/audit/sca/nuget/nuget.go
```diff
@@ -65,7 +65,9 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTree []*xrayUtils.
 		err = errors.Join(err, fileutils.RemoveTempDir(tmpWd))
 	}()
 
-	err = biutils.CopyDir(wd, tmpWd, true, nil)
+	// Specifies files and direcrories to exclude when copying the project to a temporary directory.
```
eranturgeman, line 68: The explanation you gave is straight forward from the CopyDir signature. There is no need to explain that this is a list exclusion where one line below we use it for exclusion.
It is better to explain **why** we excluded it. For example something like you wrote above:
"// We exclude Visual Studio inner directory from the dir copy since it is unnecessary for the scan process and may lead to race conditions"
(you can choose to explain it however you want though)

commands/audit/sca/pnpm/pnpm.go
```diff
@@ -115,7 +116,10 @@ func installProjectIfNeeded(pnpmExecPath, workingDir string) (dirForDependencies
 			err = errors.Join(err, fileutils.RemoveTempDir(dirForDependenciesCalculation))
 		}
 	}()
-	err = biutils.CopyDir(workingDir, dirForDependenciesCalculation, true, nil)
+
+	// Specifies files and direcrories to exclude when copying the project to a temporary directory.
```
eranturgeman, line 120: See my previos comment about this comment above the copying line

commands/audit/sca/pnpm/pnpm.go
```diff
@@ -115,7 +116,10 @@ func installProjectIfNeeded(pnpmExecPath, workingDir string) (dirForDependencies
...
+	excludeForCopy := []string{sca.ExcludeForCopyVS}
+	err = biutils.CopyDir(workingDir, dirForDependenciesCalculation, true, excludeForCopy)
```
eranturgeman, line 122:
```suggestion
	err = biutils.CopyDir(workingDir, dirForDependenciesCalculation, true, []string{sca.ExcludeForCopyVS})
```

commands/audit/sca/pnpm/pnpm.go
```diff
@@ -115,7 +116,10 @@ func installProjectIfNeeded(pnpmExecPath, workingDir string) (dirForDependencies
 			err = errors.Join(err, fileutils.RemoveTempDir(dirForDependenciesCalculation))
 		}
 	}()
-	err = biutils.CopyDir(workingDir, dirForDependenciesCalculation, true, nil)
+
+	// Specifies files and direcrories to exclude when copying the project to a temporary directory.
+	excludeForCopy := []string{sca.ExcludeForCopyVS}
```
eranturgeman, line 121:
```suggestion
```

commands/audit/sca/python/python.go
```diff
@@ -94,7 +95,10 @@ func getDependencies(auditPython *AuditPython) (dependenciesGraph map[string][]s
 		)
 	}()
 
-	err = biutils.CopyDir(wd, tempDirPath, true, nil)
+	// Specifies files and direcrories to exclude when copying the project to a temporary directory.
```
eranturgeman, line 98: See my previous comment about this comment above the copying line

commands/audit/sca/python/python.go
```diff
@@ -94,7 +95,10 @@ func getDependencies(auditPython *AuditPython) (dependenciesGraph map[string][]s
...
+	excludeForCopy := []string{sca.ExcludeForCopyVS}
```
eranturgeman, line 99:
```suggestion
```

commands/audit/sca/python/python.go
```diff
@@ -94,7 +95,10 @@ func getDependencies(auditPython *AuditPython) (dependenciesGraph map[string][]s
...
+
```
eranturgeman, line 100:
```suggestion
```

commands/audit/sca/python/python.go
```diff
@@ -94,7 +95,10 @@ func getDependencies(auditPython *AuditPython) (dependenciesGraph map[string][]s
...
+	err = biutils.CopyDir(wd, tempDirPath, true, excludeForCopy)
```
eranturgeman, line 101:
```suggestion
	err = biutils.CopyDir(wd, tempDirPath, true, []string{sca.ExcludeForCopyVS})
```

eranturgeman (changes requested):
Looks good! see my comments, fix them and we can merge :)

PR #234, opened 2024-11-14
Fix bugs when converting results to SARIF

attiasas:

-----

#### Duplicate [Secret in Binary found]
image

* Some attributes were not copied when parsing to format, when parsing for the second time it already was "padded" with the prefix which resulted in duplicate "padding"

#### Uri attribute is empty on binary scan
```json
  "locations": [
           {
             "physicalLocation": {
               "artifactLocation": {
                 "uri": ""
               }
             },
             "logicalLocations": [
               {
                 "name": "38610c0cfc18becc30c87e0af68f1c43c5cdc0ef10d920dde284df0d58fb00e1",
                 "kind": "layer",
                 "properties": {
                   "algorithm": "sha256"
                 }
               }
             ]
           }
         ],
```

* For `Docker` targets (and some binaries), the `target.Target ` is a file and not a directory and we set it as the `wd` of the SCA Sarig run, when we are translating the results location to relative paths it had `uri=wd` which resulted in leaving empty value

### DIFF
--- a/tests/testdata/output/audit/audit_sarif.json
+++ b/tests/testdata/output/audit/audit_sarif.json
@@ -1106,7 +1106,7 @@
           "ruleIndex": 6,
           "level": "warning",
           "message": {
-            "text": "[CVE-2020-28500] lodash 4.17.0"
+            "text": "Security violation [CVE-2020-28500] lodash 4.17.0"
           },
           "locations": [
             {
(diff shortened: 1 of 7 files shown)


hadarshjfrog (approved):
Code is fine - please add an explanation to the PR desc to explain the change in the satif copy mechanism

PR #256, opened 2024-12-01
Don't save SecurityCommandResults obj as a file

attiasas:

-----

We used to save `full scan results` if we run a security command in `table` format.
The object that we saved could be changed in future versions. this PR removes saving this object, it will only log the object in `DEBUG` log

### DIFF
--- a/utils/results/output/resultwriter.go
+++ b/utils/results/output/resultwriter.go
@@ -106,12 +106,17 @@ func (rw *ResultsWriter) PrintScanResults() error {
 		// Don't print if there are no results and only errors.
 		return nil
 	}
+	// Helper for Debugging purposes, print the raw results to the log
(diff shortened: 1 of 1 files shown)


eyalk007 (approved):
Looks good.
Just make sure with the PM that it is in fact `ignore for release`

hadarshjfrog (approved):

alexanderjohn:
There are teams out there that relied on that output json file and with this code removal, the scan task now needs to run twice. One for output, and one for json output.

The previous change to include CI=false flag _also_ negatively impacted our workflow. That one was however a less invasive change.

ns-dfarino:
We also had tooling that depended on that file being written.

I'd like to have the option to re-enable the previous behavior.

PR #269, opened 2024-12-19
Add tests for fail_build in case of violation

hadarshjfrog:

-----

As part of internal services migration of XSC into Xray - the field `fail_build` wasn't returning in the response from Xray - and didn't fail the run - even if it did create a violation according to the defined policies.

The fix itself happened on **Xray v3.111** - and we wanted to add a test for audit to run against it as well in JFrog-CLI

### DIFF
--- a/audit_test.go
+++ b/audit_test.go
@@ -37,12 +37,29 @@ import (
 	"github.com/jfrog/jfrog-client-go/xray/services"
 )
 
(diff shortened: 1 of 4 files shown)


attiasas:
option avaliable, was introduced in diff PR, also we have a test that checks fail build that will run on 3.111 `TestAuditJasViolationsProjectKeySimpleJson`

PR #314, opened 2025-02-25
Integrating new Config Profile changes + Add support in scanners exclusions through profile

eranturgeman:

-----
This PR contains the following changes:
1) Add a new ability to pass scanner-specific exclude patterns through Config Profile.
This change allows to pass exclude patterns for each scanner that will override definitions in jfrog-apps-config, if exists
2) Aligns to the schema changes introduced to Config Profile here: https://github.com/jfrog/jfrog-client-go/pull/1090
3) Fix and add tests according to recent changes

failing tests are known to be failing
Screenshot 2025-04-02 at 11 58 58


commands/audit/sca/common.go
```diff
@@ -31,6 +31,10 @@ var CurationErrorMsgToUserTemplate = "Failed to retrieve the dependencies tree f
 
 func GetExcludePattern(params utils.AuditParams) string {
 	exclusions := params.Exclusions()
+	if configProfile := params.GetConfigProfile(); configProfile != nil {
+		exclusions = append(exclusions, configProfile.Modules[0].ScanConfig.ScaScannerConfig.ExcludePatterns...)
```
attiasas, line 35: I see some `configProfile.Modules[0].ScanConfig`. consider refactoring to a struct func.

commands/audit/scarunner.go
```diff
@@ -60,13 +60,9 @@ func buildDepTreeAndRunScaScan(auditParallelRunner *utils.SecurityParallelRunner
...
 	}
-	if auditParams.configProfile != nil {
-		if len(auditParams.configProfile.Modules) < 1 {
-			// Verify Modules are not nil and contain at least one modules
-			return fmt.Errorf("config profile %s has no modules. A config profile must contain at least one modules", auditParams.configProfile.ProfileName)
-		}
-		if !auditParams.configProfile.Modules[0].ScanConfig.EnableScaScan {
-			log.Debug(fmt.Sprintf("Skipping SCA scan as requested by '%s' config profile...", auditParams.configProfile.ProfileName))
+	if configProfile := auditParams.AuditBasicParams.GetConfigProfile(); configProfile != nil {
+		if !configProfile.Modules[0].ScanConfig.ScaScannerConfig.EnableScaScan {
```
attiasas, line 64: we don't want to verify Modules len before accessing [0]? there could be a possibility that a config issue occur and config profile provided with empty modules? or modules without `ScanConfig`?

attiasas (approved):
LGTM, take a look at my comments

commands/audit/scarunner.go
eranturgeman: I thought about it but it seems unnecessary at this point. Whenever I fetch the profile I verify we have at least a single module. This happens in Frogbot so you cannot see it here.
When and if we will use it through cli as well we will verify it upon fetch so we will nto have to do it in each access.
As for a module without scanConfig - this is not possible. There is no pointerr here so even if the values are empty, the structs are created when we get the profile, so at the worst case we will have an empty value and nothing is added
Screenshot 2025-03-11 at 12 01 01

commands/audit/sca/common.go
eranturgeman: This call is used only twice and there is no convenient place to put this struct method since the struct sits in client-go.
In addition this is only temporary since we will move at some point to work with several modules and I want to emphasize the work with module[0] at this point

PR #383, opened 2025-03-05
Add skipAutoInstall for Pip and Poetry package managers

hadarshjfrog:

-----

This PR extends the SkipAutoInstall abilities into Pip and Poetry.
This allows to run audit scan (or any Frogbot flow) without running installation command during the process, and the resolved dependencies are the existing dependencies within the execution context (virtual env or global if possible).

**Note: In order to use this for the mentioned python package managers, an installation at the customer's end is required and also we must be in the context of the virtual env in which we installed the dependencies**

### DIFF
--- a/commands/audit/sca/python/python.go
+++ b/commands/audit/sca/python/python.go
@@ -34,8 +34,8 @@ const (
 	CurationPipMinimumVersion = "23.0.0"
 )
 
-func BuildDependencyTree(params utils.AuditParams) (dependencyTree []*clientutils.GraphNode, uniqueDeps []string, downloadUrls map[string]string, err error) {
-	dependenciesGraph, directDependenciesList, pipUrls, errGetTree := getDependencies(params)
+func BuildDependencyTree(params utils.AuditParams, technology techutils.Technology) (dependencyTree []*clientutils.GraphNode, uniqueDeps []string, downloadUrls map[string]string, err error) {
+	dependenciesGraph, directDependenciesList, pipUrls, errGetTree := getDependencies(params, technology)
 	if errGetTree != nil {
 		err = errGetTree
 		return
@@ -60,7 +60,7 @@ func BuildDependencyTree(params utils.AuditParams) (dependencyTree []*clientutil
 	return
 }
 
-func getDependencies(params utils.AuditParams) (dependenciesGraph map[string][]string, directDependencies []string, pipUrls map[string]string, err error) {
+func getDependencies(params utils.AuditParams, technology techutils.Technology) (dependenciesGraph map[string][]string, directDependencies []string, pipUrls map[string]string, err error) {
 	wd, err := os.Getwd()
 	if errorutils.CheckError(err) != nil {
 		return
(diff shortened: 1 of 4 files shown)


commands/audit/sca/python/python.go
```diff
@@ -99,25 +99,31 @@ func getDependencies(auditPython *AuditPython) (dependenciesGraph map[string][]s
...
 		return
 	}
-
-	restoreEnv, err := runPythonInstall(auditPython)
-	defer func() {
-		err = errors.Join(err, restoreEnv())
-	}()
-	if err != nil {
-		return
+	pythonTool := pythonutils.Pip
```
eranturgeman, line 102: This is not needed. At this point we necessarily have tech. If not the scan was already skipped in a prior stage so the check if tech exists is not needed

commands/audit/sca/python/python.go
```diff
@@ -99,25 +99,31 @@ func getDependencies(auditPython *AuditPython) (dependenciesGraph map[string][]s
...
+	if len(params.Technologies()) > 0 {
+		pythonTool = pythonutils.PythonTool(params.Technologies()[0])
+	}
+	if !params.SkipAutoInstall() {
+		restoreEnv, restoreEnvErr := runPythonInstall(params, pythonTool)
```
eranturgeman, line 107: change the restoreEnvErr back to err.
In a new name is required switch to installErr since this is not an error we get from the env restore operation

commands/audit/sca/python/python.go
```diff
@@ -100,24 +91,28 @@ func getDependencies(auditPython *AuditPython) (dependenciesGraph map[string][]s
 		return
 	}
 
-	restoreEnv, err := runPythonInstall(auditPython)
-	defer func() {
-		err = errors.Join(err, restoreEnv())
-	}()
-	if err != nil {
-		return
+	pythonTool := pythonutils.PythonTool(params.Technologies()[0])
```
attiasas, line 94: for projects with multipule package managers (or if requested). the tech you are searching may not be at `0` place

commands/audit/sca/python/python_test.go
```diff
@@ -207,6 +209,84 @@ func TestBuildPoetryDependencyList(t *testing.T) {
...
+			technology:                       techutils.Pip.String(),
+			installBeforeFetchingInitialDeps: false,
+		},
+		{
+			name:                             "pip: project installed before dep tree construction| install forbidden",
+			testDir:                          filepath.Join("projects", "package-managers", "python", "pip", "pip", "requirementsproject"),
+			technology:                       techutils.Pip.String(),
+			installBeforeFetchingInitialDeps: true,
+		},
+		// TODO add similar test cases for pipenv and poetry
```
attiasas, line 231: ??

attiasas (commented):
Nice!, chcek out my comments.

commands/audit/sca/python/python.go
eranturgeman: correct, but we iterate the ScanTargets and execute a GetTechDependency tree at a single target each time. Each target can have a single technology at a time, therefore what you are saying is not possible.
In addition. if you look to the original code - this is exactly what we did, except I did it in a different place

commands/audit/sca/python/python_test.go
eranturgeman: removed

commands/audit/sca/python/python.go
eranturgeman: fixed

attiasas (approved):

PR #384, opened 2025-03-09
Combine multiple runs with same Tool name in SARIF format

attiasas:

-----

Some tools (i.e Github) can't process report with multiple runs that has the same tool.
It will only show the results of one of the runs.

To mitigate this, at the end of the parser we are combining similar runs to a single one.

### Aditional changes
* fix GH2012 - 'reportingDescriptor' object does not provide a 'fullDescription'
* Adding the name to the SCA rules.


utils/results/common.go
```diff
@@ -570,6 +570,10 @@ func GetScaIssueId(depName, version, issueId string) string {
 	return fmt.Sprintf("%s_%s_%s", issueId, depName, version)
 }
 
+func IdToFriendlyName(input string) string {
```
eranturgeman, line 573: Loved the name but it is not really self explanatory.
Maybe: replaceUnderscoreWithDash?

utils/formats/sarifutils/sarifutils.go
```diff
@@ -83,12 +83,39 @@ func CombineReports(reports ...*sarif.Report) (combined *sarif.Report, err error
...
+		if existingRun := reportContainsRun(GetRunToolName(run), combined); existingRun != nil {
+			AggregateMultipleRunsIntoSingle([]*sarif.Run{run}, existingRun)
+		} else {
 			combined.AddRun(run)
 		}
 	}
 	return
 }
 
+func reportContainsRun(toolName string, report *sarif.Report) (run *sarif.Run) {
```
eranturgeman, line 110: I'd change the name here. The name indicates a boolean answer where in fact we return a run with a matching tool name.
maybe: getRunByToolName

utils/formats/sarifutils/sarifutils.go
```diff
@@ -83,12 +83,39 @@ func CombineReports(reports ...*sarif.Report) (combined *sarif.Report, err error
 	}
 	for _, report := range reports {
 		for _, run := range report.Runs {
+			if existingRun := reportContainsRun(GetRunToolName(run), combined); existingRun != nil {
```
eranturgeman, line 86: we have a bit of code duplication here (with CombineMultipleRunsWithSameTool). maybe we can combine it?
for example we can pass a report by pointer to CombineMultipleRunsWithSameTool (and create a new report only if it is nil, so we will create a new report only when it is called from the Get() func) and the rest is the exact same loop

eranturgeman (approved):
See my comments

PR #403, opened 2025-04-02
Exclude patterns expansion

hadarshjfrog:

-----
As part of improving false positive results, we wish to ignore findings who shouldn't be scanned, as such:
1. We wish to add a possibility to ignore partial patterns, not only folder
2. Also - add `test.` to ignore the common cases in GO and JS/TS


attiasas (changes requested):
This change will probably Break some users since you are adding more default ignores that can't be controlled or may not be intentional by the user and can't be `turned off`

Some tests fails, please fix them.

PR #484, opened 2025-06-30
Gem-Dep-Tree-Creation

basel1322:
Added Ruby Dependency Tree Generation

The code is compatible with Ruby version 2.6.0 and above, as it relies on the bundle command, which is included by default starting from Ruby 2.6.0.

It uses bundle lock on the Gemfile to generate a lockfile containing metadata for all dependencies. This metadata is then processed to build the structure required by Xray for scanning and analysis against its database.

For example, this a Gemfile:

```
source "https://rubygems.org"
ruby "2.7.0"

gem "puma", "~> 5.0"         # Web server   # Stylesheets
````

When we use bundle lock, it will generate text file with metadata for the gems, example below:

```
GEM
  specs:
    nio4r (2.7.4)
    puma (5.6.9)
      nio4r (~> 2.0)

PLATFORMS
  arm64-darwin-24

DEPENDENCIES
  puma (~> 5.0)

RUBY VERSION
   ruby 2.6.10p210

BUNDLED WITH
   2.4.22
```


attiasas (changes requested):
LGTM, nice work!
1. the new dependency tree generation is not used in `scarunner.go` if we want users to be able to use it in 'audit' we need to add a case there
2. please add integration test on a ruby project at 'audit_test.go`

PR #519, opened 2025-07-31
Add nil check for ScaResults when SCA scan is not performed in audit

kerenr-jfrog:

-----

- Added a missing nil check before accessing ScaResults to properly handle audit scans that are run without SCA enabled. This prevents potential runtime errors when ScaResults is nil.

### DIFF
--- a/utils/results/common.go
+++ b/utils/results/common.go
@@ -81,7 +81,10 @@ func checkIfFailBuildConsideringApplicability(target *TargetResults, entitledFor
 	jasApplicabilityResults := target.JasResults.GetApplicabilityScanResults()
 
 	// Get new violations from the target
-	newViolations := target.ScaResults.Violations
+	var newViolations []services.Violation
+	if target.ScaResults != nil {
+		newViolations = target.ScaResults.Violations
+	}
 
 	// Here we iterate the new violation results and check if any of them should fail the build.
 	_, _, err := ForEachScanGraphViolation(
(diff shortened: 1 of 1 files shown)


utils/results/common.go
```diff
@@ -98,18 +101,20 @@ func checkIfFailBuildConsideringApplicability(target *TargetResults, entitledFor
...
-			target.ScanTarget,
-			deprecatedViolations,
-			entitledForJas,
-			jasApplicabilityResults,
-			checkIfShouldFailBuildAccordingToPolicy(shouldFailBuild),
-			nil,
-			nil)
-		if err != nil {
-			return err
+	if target.ScaResults != nil {
```
eranturgeman, line 104: just for easier reading i'd suggest inserting the opposite condition here:
if target.ScaResults == nil { return nil }
... the following loop ...

eranturgeman (commented):
This is a sensitive part of the flow, please make sure the flow works as expected manually (create a scenario where we can reproduce the exact flow that reaches the new conditions)
Also please refer to TestViolationFailBuild and add a testCase that simulates this case

utils/results/common.go
```diff
@@ -96,6 +99,10 @@ func checkIfFailBuildConsideringApplicability(target *TargetResults, entitledFor
 		return err
 	}
```
eranturgeman, line 101: move this check to the top and if target.ScaResults == nil simply return sicne there is nothing to check

PR #541, opened 2025-08-24
Add SAST runs to cdx content

attiasas:

-----

Add SAST content to the cyclonedx content by adding a `sast` attributes outside of spec
(*leaving option to parse into 'vulnerabilities' but not official)


utils/utils.go
```diff
@@ -342,6 +342,10 @@ func SaveCdxContentToFile(pathToSave string, bom *cyclonedx.BOM) (err error) {
 	return cyclonedx.NewBOMEncoder(file, cyclonedx.BOMFileFormatJSON).SetPretty(true).Encode(bom)
 }
 
+func DumpCdxWithSastContentToFile(fileContent []byte, scanResultsOutputDir, filePrefix string, threadId int) (err error) {
```
hadarshjfrog, line 345: I see the Sast in the name - but not referring to it in the content or other params...

utils/results/conversion/cyclonedxparser/cyclonedxparser.go
```diff
@@ -31,18 +31,21 @@ const (
 type CmdResultsCycloneDxConverter struct {
 	entitledForJas bool
 	xrayVersion    string
+	parseSast      bool
```
hadarshjfrog, line 34: why do we need it? shouldn't it be consistent - that if we didn't get sast results - it just would be omitted?

hadarshjfrog (commented):
Let's discuss how the addition of the SAST affect the existing CDX file

utils/formats/cdxutils/cyclonedxutils.go
```diff
@@ -19,6 +20,11 @@ import (
 	"github.com/jfrog/jfrog-cli-security/utils/techutils"
 )
 
+type BOMWithSAST struct {
```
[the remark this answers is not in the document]
attiasas: Yeah, its omitted if empty.
all the sides (client/server) and components that involves in this should have similar struct so they can unmarshal the data to it and not use the generic cyclonedx.NewBOM() when parsing/marshling the data

utils/results/conversion/cyclonedxparser/cyclonedxparser.go
attiasas: it is omitted.
There is no place in the code when we pass this arg as `true`, so we will never parse the content of the SARIF to the cdx but directly attach the runs content to our new attribute.

I added it since I already wrote a code that parses the results from the SARIF (similar to IaC and Secrets...)
I can also remove the old code related to that if you think its better....

utils/utils.go
hadarshjfrog: @attiasas ?

PR #561, opened 2025-09-05
added-pass-through for gradle

basel1322:
Added pass through to gradle
-----


attiasas (commented):
Nice work!
1. Update the gradle dep tree version (after release) at `download-jars.sh` and use the script to download the jar from releases. (make sure to align to `dev`, this version may already been released)
2. I don't see any usage for this new option (always passes false), is this PR not adding this to curation? only preparing for it? if you are adding it to curation, integration tests are needed

PR #598, opened 2025-10-29
XRAY-128408 - Add Sast rules flag to audit command

iddoh-jfrog:
Resolves [XRAY-128408](https://jfrog-int.atlassian.net/browse/XRAY-128408)

## Add Custom rules to the SAST scan
Prepare a custom json rules file with your rules:
```
[
  {
    "name": "custom-rule",
    "message": "User-controlled data used as argument to math.sqrt",
    "finder": {
      "type": "FlowFinder",
      "sources": {
        "type": "calls",
        "names": [
          "input"
        ]
      },
      "sinks": {
        "type": "calls",
        "names": [
          "math.sqrt"
        ]
      }
    },
    "cwe": null,
    "description": "User-controlled square root",
    "severity": "high",
    "tags": []
  }
]
```

Run with the flag `--add-sast-rules=/path/to/file.json`

### DIFF
--- a/cli/docs/flags.go
+++ b/cli/docs/flags.go
@@ -56,6 +56,9 @@ const (
 	Sast      = "sast"
 	Secrets   = "secrets"
 	WithoutCA = "without-contextual-analysis"
+
+	// Sast related flags
+	AddSastRules = "add-sast-rules"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


cli/docs/flags.go
```diff
@@ -56,6 +56,9 @@ const (
 	Sast      = "sast"
 	Secrets   = "secrets"
 	WithoutCA = "without-contextual-analysis"
+
+	// Sast related flags
+	SastRules = "sast-rules"
```
attiasas:
```suggestion
	SastRules = "sast-rules-file"
```
consider changing the flag name (unless approved already)

cli/scancommands.go
```diff
@@ -407,6 +410,27 @@ func AuditCmd(c *components.Context) error {
...
+			case errors.Is(err, fs.ErrPermission):
+				return pluginsCommon.PrintHelpAndReturnError(fmt.Sprintf("permission denied to read sast rules file '%s': %v", sastRules, err), c)
+			default:
+				return pluginsCommon.PrintHelpAndReturnError(fmt.Sprintf("error reading sast rules file '%s': %v", sastRules, err), c)
+			}
+		}
+
+		auditCmd.SetSastRules(sastRules)
+	}
+
```
attiasas:
```suggestion
	// Validate that there is a sast scan before setting the sast rules
	sastRulesFile := c.GetStringFlagValue(flags.SastRules)
	if sastRulesFile != "" {
	   // Check if file exists
	   if exists, err := fileutils.IsFileExists(cdxFilePath, false); err != nil {	  
	       return pluginsCommon.PrintHelpAndReturnError(fmt.Sprintf("file '%s' does not exist: %s", sastRules, err.Error()), c)
	   }
	    // Validate scan is performed
	   if len(params.ScansToPerform()) > 0 && !slices.Contains(params.ScansToPerform(), utils.SecretsScan)    {
	       return pluginsCommon.PrintHelpAndReturnError(fmt.Sprintf("flag '--%s' can only be used with '--%s'", flags.SastRules, flags.Sast), c)
	   }
		auditCmd.SetSastRules(sastRulesFile)
	}

```

try to reduce complexity and duplicate calls to funcs. use utils in repo

attiasas (approved):
LGTM

PR #705, opened 2026-03-24
Convert remediation error to not fail commands

attiasas:
## Convert remediation error to warning instead of failing commands

Fix version attachment via the remediation API is not critical to SBOM enrichment. Previously, a failure in the remediation step would propagate as a hard error, causing the entire enrich command to fail. This change makes the remediation error non-fatal by default — it is logged as an error but the command continues and returns partial results.

### Changes

- Added a `failOnRemediationError` field to `EnrichScanStrategy` (defaults to `false`) so callers can opt in to strict behaviour when needed.
- Added a `WithFailOnRemediationError` option function to configure the flag.
- Extracted the remediation logic into a dedicated `attachFixedVersionsToVulnerabilities` method for clarity.
- Updated `SbomEnrichTask` to log the remediation error and continue when `failOnRemediationError` is `false`, or propagate it when `true`.

---

-----

### DIFF
--- a/sca/scan/enrich/runner.go
+++ b/sca/scan/enrich/runner.go
@@ -19,6 +19,8 @@ import (
 type EnrichScanStrategy struct {
 	serverDetails *config.ServerDetails
 	projectKey    string
+	// Fix version attachment is not critical, so we continue even if it fails without returning an error
+	failOnRemediationError bool
 }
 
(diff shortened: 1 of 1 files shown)


attiasas:
may be reopen in future, for now keep error

PR #716, opened 2026-04-06
Add 7z as alternative for unzip

attiasas:
## Accept 7z as alternative to unzip for JAS scanner prerequisites

The `ValidateRequiredInstalledSoftware` pre-check currently hard-requires `unzip` on PATH. Environments that use 7-Zip (`7z`) instead of GNU unzip — such as many Windows-based enterprise setups — fail validation and cannot run Advanced Security Scans.

This changes the required-software data structure from a flat list to groups of alternatives, so that having **either** `unzip` or `7z` installed satisfies the zip-extraction prerequisite. The structure is extensible: future alternative tools can be added to any group without logic changes.

### Changes
- **`jas/analyzermanager.go`**
  - Changed `scannersRequiredInstalledSoftware` from `[]string` to `[][]string` where each inner slice is a group of equivalent tools (any one satisfies the requirement).
  - The zip extraction group is now `{"unzip", "7z"}`.
  - Refactored `ValidateRequiredInstalledSoftware()` to iterate groups and check if at least one alternative is available.
  - Added `isAnySoftwareInstalled()` helper for the group-level check.
  - Added `formatMissingSoftwareError()` for clear error messages that list all accepted alternatives (e.g. `"could not find any of the required executables ('unzip' or '7z') in the system PATH..."`).
- **`jas/common_test.go`**
  - Added `TestIsAnySoftwareInstalled` (5 sub-cases: single found, single missing, first alternative found, second alternative found, all missing).
  - Added `TestFormatMissingSoftwareError` (single-tool vs. multi-alternative message formatting).
  
-----


jas/analyzermanager.go
```diff
@@ -333,10 +339,30 @@ func RunAnalyzerManagerWithPipesAndDownload(envVars map[string]string, cmd strin
...
+
+func formatMissingSoftwareError(alternatives []string) error {
+	if len(alternatives) == 1 {
+		return fmt.Errorf("could not find the required '%s' executable in the system PATH to run the Advanced Security Scans", alternatives[0])
+	}
+	quoted := make([]string, len(alternatives))
+	for i, alt := range alternatives {
+		quoted[i] = "'" + alt + "'"
+	}
+	return fmt.Errorf("could not find any of the required executables (%s) in the system PATH to run the Advanced Security Scans", strings.Join(quoted, " or "))
```
eranturgeman: is switch the 'or' in the strings.Join to ',' so we will not get odd sentences when new techs are appended (a or b or c or d)

jas/analyzermanager.go
```diff
@@ -60,8 +60,14 @@ const (
 
 type JasDiffScanEnvValue string
 
-var scannersRequiredInstalledSoftware = []string{
-	"git", "unzip", "curl",
+// scannersRequiredInstalledSoftware lists groups of executables needed by the
+// scanner at runtime.  Each inner slice represents a single capability: when
+// multiple entries exist in a group, ANY one of them satisfies the requirement
+// (e.g. either "unzip" or "7z" can handle zip extraction).
```
eranturgeman:
```suggestion
// scannersRequiredInstalledSoftware lists groups of executables needed by the scanner at runtime. 
// Each inner slice represents a single capability, and  entry within a group can satisfies the requirement
```

eranturgeman (approved):
LGTM! see my comments

PR #719, opened 2026-04-06
Remove duplications in CDX format

attiasas:
### Description

Fix CycloneDX v1.6 schema compliance by removing duplicate entries in generated SBOMs. Two defects caused `uniqueItems` constraint violations:

1. **Duplicate `dependsOn` entries**: `getBinaryNodeDirectDependencies` appended child refs without deduplication (unlike its counterpart `getNodeDirectDependencies` which uses a Set). When a binary graph node listed the same child multiple times (common in Docker image layer analysis), duplicates appeared in the serialized `dependsOn` array.

2. **Duplicate license blocks**: `CreateScaComponentFromBinaryNode` iterated `node.Licenses` and created a `LicenseChoice` for each entry without checking for duplicates. If upstream data contained repeated license names (e.g. `["Unknown", "Unknown"]`), all duplicates were emitted.

### Changes

- `utils/results/common.go`:
  - `getBinaryNodeDirectDependencies`: replaced direct slice append with `datastructures.MakeSet` for deduplication, with `slices.Sort` for deterministic output order.
  - `CreateScaComponentFromBinaryNode`: added a `datastructures.MakeSet` to skip duplicate license IDs when building the license list.
- `utils/results/common_test.go`:
  - Added test case "duplicate dependencies are deduplicated" to `TestCompTreeToSbom`.
  - Added test case "duplicate licenses are deduplicated" to `TestCompTreeToSbom`.
  - Updated expected dependency order in existing test case to match the now-sorted output.

---


utils/results/common.go
```diff
@@ -1052,10 +1054,13 @@ func getDataFromBinaryNode(node *xrayUtils.BinaryGraphNode, parsed *datastructur
...
 func getBinaryNodeDirectDependencies(node *xrayUtils.BinaryGraphNode) (dependencies *[]string) {
-	dependencies = &[]string{}
+	depSet := datastructures.MakeSet[string]()
 	for _, dep := range node.Nodes {
-		*dependencies = append(*dependencies, techutils.XrayComponentIdToCdxComponentRef(dep.Id))
+		depSet.Add(techutils.XrayComponentIdToCdxComponentRef(dep.Id))
 	}
+	dependencies = &[]string{}
+	*dependencies = depSet.ToSlice()
+	slices.Sort(*dependencies)
```
eranturgeman, line 1064: should getNodeDirectDependencies sort similarly? Right now the two methods produce outputs in different orders — one alphabetical, one determined by map/set iteration, which in Go is non deterministic. If determinism is a goal, it should apply to both.

eranturgeman (approved):
see my comment

PR #740, opened 2026-04-20
BuildScan - Option to control showing Violations

attiasas:
# feat(build-scan): add `--violations` to control policy violations in output

## Summary

Adds a `--violations` CLI flag for `build-scan`, defaulting to `true`.
When set to `false`, policy violations are omitted from CLI output and from violation enrichment; with `--project`, violations are always included regardless of the flag.
If violations are omitted, the command does not treat Xray’s `fail_build` response as a CLI failure when `--fail` is used, and the results writer’s violation context is disabled accordingly.

## Changes

- `cli/docs/flags.go` — `Violations` / `BuildViolations` flag; register on `BuildScan` with help text; default `true`.
- `cli/utils.go` — `EffectiveIncludeViolations(violationsFlag, projectProvided)`; project forces inclusion.
- `cli/scancommands.go` — wire flag into `BuildScanCommand` via `SetIncludeViolations` and updated `SetIncludeVulnerabilities` condition.
- `commands/scan/buildscan.go` — `includeViolations` on `BuildScanCommand` (default `true`); guard violation enrichment, `SetHasViolationContext`, and `fail` / `FailBuild` interaction when violations are omitted.
- `cli/scancommands_test.go` — `TestEffectiveIncludeViolations` table test.

## Notes

- **Behavior**: With `--project`, violations are always included (overrides `--violations=false`).

---


cli/utils.go
```diff
@@ -126,6 +126,12 @@ func getProject(c *components.Context) string {
 	return os.Getenv(coreutils.Project)
 }
 
+// EffectiveIncludeViolations reports whether the command should surface violations in CLI results.
+// The --violations flag defaults to true; when --project is set, violations are always included.
```
orto17, line 130: when --project is set, violations are always included - should appear in the doc

cli/scancommands.go
```diff
@@ -448,10 +448,12 @@ func BuildScan(c *components.Context) error {
 	if err = validateConnectionAndViolationContextInputs(c, serverDetails, format); err != nil {
 		return err
 	}
+	includeViolations := EffectiveIncludeViolations(c.GetBoolFlagValue(flags.Violations), isProjectProvided(c))
 	buildScanCmd := scan.NewBuildScanCommand().
 		SetServerDetails(serverDetails).
 		// Sarif shouldn't include the additional all-vulnerabilities info that received by adding the vuln flag
-		SetIncludeVulnerabilities(getProject(c) == "" || (format != outputFormat.Sarif && c.GetBoolFlagValue(flags.Vuln))).
+		SetIncludeVulnerabilities(!includeViolations || (format != outputFormat.Sarif && c.GetBoolFlagValue(flags.Vuln))).
```
orto17, line 455: when omitting the getProject(c) check - isn't it a regression to the current behavior?

commands/scan/buildscan.go
orto17: Cursor: The error string still says build-scan command with '--vuln' flag is not supported when includeVulnerabilities is true. After the wiring change, includeVulnerabilities can become true because of --violations=false (!includeViolations), not only because of --vuln. If that triggers the minimum-version check, the message may confuse users. Worth aligning the error text with both causes or splitting validation paths.

orto17 (approved):

cli/scancommands.go
attiasas: No:
`includeViolations := EffectiveIncludeViolations(c.GetBoolFlagValue(flags.Violations), isProjectProvided(c))`

commands/scan/buildscan.go
attiasas: changed

cli/utils.go
attiasas: changed after consultation with the technical writer

PR #744, opened 2026-04-29
Add --format flag to sbom-enrich

ehl-jf:
## Summary

Adds a `--format` flag to the `sbom-enrich` (alias `se`) command, allowing the output format to be selected when the input SBOM is XML (CycloneDX). Currently `json` is the only supported value for the flag; when set, XML input is converted to JSON with the enriched vulnerability data appended.

## Changes

- `cli/scancommands.go`
  - Register the `sbom-enrich` command with `SupportedFormats: [Json]` so the shared `--format` flag is wired in.
  - Read the format from the command context and pass it to `EnrichCommand` via the new setter.
- `commands/enrich/enrich.go`
  - New `outputFormat` field on `EnrichCommand` plus `SetOutputFormat(...)` setter.
  - New `AppendVulnsFromXMLToJson` that parses the XML SBOM, converts it to ordered JSON via the new `xmlElementToOrderedJson` helper, appends Xray vulnerabilities (`bom-ref` + CVE id), and prints the result.
  - `Run()` now branches on the requested format: XML input + `--format=json` produces JSON output; otherwise the existing XML/JSON paths are preserved (backward compatible default).
  - Added `printVulnerabilitiesTable` helper for tabular vulnerability rendering.
- `commands/enrich/enrich_test.go`
  - New unit tests covering the XML→JSON conversion, vulnerability appending, and table output.
- `go.mod` / `go.sum`
  - Bumped `github.com/jfrog/jfrog-client-go` to latest master (`v1.55.1-0.20260505115216-b6c67f807bc3`).

## Validation

- [ ] All [tests](https://github.com/jfrog/jfrog-cli-security/actions/workflows/test.yml) have passed. New unit tests added for the JSON output path in `commands/enrich/enrich_test.go`.


vitaliil-jfrog (approved):

utils/formats/output_format.go
```diff
@@ -0,0 +1,13 @@
...
+	outputFormat "github.com/jfrog/jfrog-cli-core/v2/common/format"
+)
+
+func GetOutputFormat(format string) (f outputFormat.OutputFormat, err error) {
+	f = outputFormat.Table
+	if format != "" {
+		f, err = outputFormat.ParseOutputFormat(format, outputFormat.All)
+	}
+	return
+}
```
attiasas: No need for this function.
Use `outputFormat.ParseOutputFormat(format,{options...})` directly.
You should register a flag with the default value at `cli/docs/flags.go` see how the `CurationOutput` option is defined/used, you have a similar case

cli/scancommands.go
```diff
@@ -71,13 +74,14 @@ func getAuditAndScansCommands() []components.Command {
...
-			Arguments:   enrichDocs.GetArguments(),
-			Category:    securityCategory,
-			Action:      EnrichCmd,
+			Name:             "sbom-enrich",
+			Aliases:          []string{"se"},
+			Flags:            flags.GetCommandFlags(flags.Enrich),
+			Description:      enrichDocs.GetDescription(),
+			Arguments:        enrichDocs.GetArguments(),
+			Category:         securityCategory,
+			SupportedFormats: []outputFormat.OutputFormat{outputFormat.Json, outputFormat.Table},
```
attiasas: I can see this option has been added recently to the core logic.
Currently CLI uses the plugins `Flags` attribute. as mentioned in other comment...

In addition seems like the current implementation is a mix:
You have the `GetOutputFormat` func you added with `table` as default but the core logic added `DefaultFormat` attribute as well as the `SupportedFormats`, why not use that?

I prefer the way the current repository handles the flags, when migrating we should probably migrate all and not having a mix

commands/enrich/enrich.go
```diff
@@ -192,18 +266,27 @@ func (enrichCmd *EnrichCommand) Run() (err error) {
 		return errorutils.CheckError(scanResults.GetErrors())
 	}
 
+	if enrichCmd.outputFormat == coreformat.Table {
+		if err = enrichCmd.printVulnerabilitiesTable(scanResults, os.Stdout); err != nil {
+			return
+		}
+		log.Info("Enrich process completed successfully.")
+		return
+	}
```
attiasas: This is not enriching the input files
```
jf sbom-enrich - Enrich sbom format JSON located on the local file-system with Xray.
```
This option is only printing the Xray results in a table format. IDK if that is intended.

commands/enrich/enrich.go
```diff
@@ -217,6 +300,27 @@ func (enrichCmd *EnrichCommand) CommandName() string {
...
+			}
+			for component := range vuln.Components {
+				if _, err := fmt.Fprintf(tw, "%s\t%s\n", component, cveID); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return tw.Flush()
+}
```
attiasas, line 315: if we already printing a table, use and print with the current format logic (see `utils/results/conversion/convertor.go` and `utils/results/output/resultwriter.go`)

cli/scancommands.go
attiasas: In addition, the current command also supports `XML` output format.
I can't see the option here so we need to add it to avoid regression

attiasas (changes requested):
Nice Job, please take a look at my comments.. In addition:
* This change makes the command behave a bit differently when passing the `table` format, not enriching, only printing...
* Add a description to your PR that reflects your changes.
* Resolve conflict and align code against the updated target branch `dev` (not `main`...)
* Try to add integration tests to `enrich_test.go`

utils/formats/output_format.go
ehl-jf: Actually this function was used before by some of the commands here.
But it was located in `jfrog-cli-core` but has been deprecated there.
So I move it back here to avoid changing to many thing.

An effort was made make the declaration of OutputFormat easy an common to all plugin, check the section 'Define output formats' in https://github.com/jfrog/jfrog-cli-core/blob/master/plugins/README.md.

With this new mechanism no need to declare the format flag it is done by the framework, and it can be retrieved from the context `ctx.GetOutputFormat()`

cli/scancommands.go
ehl-jf: The enrich command already used `c.GetOutputFormat()`
I did not want to touch the others ones as this is just about the one without format.

ehl-jf: Sure we can add it.
It is the default behavior BTW if no `--format` is specified the command acts as before.

utils/formats/output_format.go
ehl-jf: After rebasing over dev, I saw that you've already get rid of the deprecated function.
Beware that function used to return `Table` as the default and `ParseOutputFormat` does not.
Hope it did not break anything.
cc @attiasas

commands/enrich/enrich.go
ehl-jf: Not really, So I guess this format is not relevant here... Removing it

cli/scancommands.go
ehl-jf: From what I gather the current behaviour is as below:

To decide where to print XML is not based on the `--format` but depends on the result of `isXML(scanResults.Targets)`

If we add `--format xml` it will mean if `isXML(scanResults.Targets) == false` we will have to convert non-XML data to XML and print it.

@attiasas should we do it  ?

In this PR we only convert XML to json if `--format==json` and `isXML(scanResults.Targets)==true`.

PR #746, opened 2026-05-05
Migrate to jfrog/boost

attiasas:
## Summary

This change switches the GitHub Actions workflow used for CI optimization in `.github/workflows/test.yml` from `jfrog-fastci/fastci@v0` to `jfrog/boost@v0`. Inputs tied to the old FastCI action (`fastci_otel_token` / `FASTCI_TOKEN`, `install_bashi`, and `enabled_optimizations` for `go_test_junit_optimization`) are removed so the workflow matches the new action’s interface.

## Changes

- **`.github/workflows/test.yml`**: In every job that ran “Start FastCI Optimization,” replace `uses: jfrog-fastci/fastci@v0` with `uses: jfrog/boost@v0`.
- **Inputs**: Keep only `github_token: ${{ secrets.GITHUB_TOKEN }}` for that step; drop `fastci_otel_token`, `install_bashi`, and `enabled_optimizations` everywhere they appeared (including the variant that had `install_bashi: true` on the first job).


yahav-ohana:
LGTM

orto17 (approved):

PR #759, opened 2026-05-21
Feature/xray 138688 add yarn support for jf ca

gauriy-tech:

-----

Add Yarn V2/V3 support for jf curation-audit

Previously jf ca had no Yarn support — running it on a Yarn project would either silently resolve via the npm code path or produce incomplete results. This PR adds full Yarn Berry (V2/V3) support.

What changed:

Curation install via Artifactory — rewrites the Yarn registry URL in .yarnrc.yml to the api/curation/audit/<repo>/ endpoint before running yarn install --mode=update-lockfile, then restores the original config. This routes all Yarn resolution through the curation service so 403 responses surface blocked packages.

Blocked package detection when install fails — when a curation 403 aborts the lockfile write (Yarn V3 rolls back on any error), the code falls back to probing declared direct dependencies via HEAD requests and renders the same blocked-package table the normal path produces.

Workspace member support — jf ca --working-dirs=<member> on a Yarn workspace member re-routes to the workspace root automatically (Yarn V2+ cannot operate from a non-root). A new DetectedTechnologiesListForCurationAudit() promotes npm-detected workspace members to Yarn so the correct code path runs.

Workspace member filtering in graph — local workspace member packages (identified by the Yarn Berry name-hash pattern, e.g. admin-ui-428bae:0.0.0) are skipped when probing Artifactory — they have no remote artifact.

Root-only dep scope — when jf ca is run from a workspace root without --working-dirs, only the root package.json direct deps are audited. Use --working-dirs to audit individual members.

--run-native guard — --run-native is explicitly rejected for non-npm techs with a clear error; only npm implements the matching native-config flow today.

Graceful non-JSON 403 handling — when Artifactory returns an HTML error page instead of a JSON curation envelope, the package is still recorded as blocked with unknown policy rather than being silently dropped.

Testing done is documented here
https://jfrog-int.atlassian.net/browse/XRAY-140292

Error screenshot reference
Screenshot 2026-05-29 at 7 46 50 PM


gauriy-tech:

commands/curation/curationaudit.go
```diff
@@ -543,10 +602,9 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
 	}
 	// Fetch status for each node from a flatten graph which, has no duplicate nodes.
 	packagesStatusMap := sync.Map{}
-	// if error returned we still want to produce a report, so we don't fail the next step
-	err = analyzer.fetchNodesStatus(depTreeResult.FlatTree, &packagesStatusMap, rootNodes)
-	// Auth errors are unrecoverable — skip building a misleading partial report.
-	if analyzer.cancelled.Load() {
+	// Any non-200/403 response from fetchNodesStatus means the walk is incomplete; bail out rather
+	// than rendering a misleading "0 blocked packages" report.
+	if err = analyzer.fetchNodesStatus(depTreeResult.FlatTree, &packagesStatusMap, rootNodes); err != nil {
```
gineshkumar: Has this behavior change been signed off with Product?

`auditTree` now returns early on **any** non-nil error from `fetchNodesStatus`. Previously the walk continued and the user got both the partially-collected blocked packages **and** the error condition. The updated `"npm tree - two blocked one error"` test case captures the shift:

- **Before:** report listed `underscore@1.13.6` as blocked (because the HEAD-check returned 403) and surfaced the `lightweight` HTTP 500 as an executor error → user knew "one definitively blocked, one couldn't be checked."
- **After:** report is empty and the user sees only the executor error → user knows "scan failed" but loses the actionable signal about `underscore`.

This affects every tech path that goes through `auditTree` (npm, pip, maven, gradle, gem, nuget, docker, …), not only yarn — so the blast radius is wider than the PR title suggests.

Both extremes are defensible, but they're not the only options. A third design that doesn't appear to have been evaluated: introduce a `BlockingReasonScanIncomplete` (or equivalent) entry in the report so the walker continues, the partially-collected blocked packages are surfaced, AND the rows that couldn't be checked are explicitly marked "scan incomplete" rather than silently dropped. That preserves the actionable signal for CI consumers while still being honest about coverage gaps.

Could you confirm Product has reviewed the change in user-visible behavior, and ideally call it out in the PR description / release notes so existing consumers of partial reports aren't caught by surprise?

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -180,29 +1112,44 @@ func runYarnInstallAccordingToVersion(curWd, yarnExecPath string, installCommand
...
+		nodeModulesDirExists, err := fileutils.IsDirExists(nodeModulesFullPath, false)
 		if err != nil {
-			err = fmt.Errorf("failed while checking for existence of node_modules directory: %s", err.Error())
-			return
+			return fmt.Errorf("failed while checking for existence of node_modules directory: %s", err.Error())
 		}
 		if !nodeModulesDirExists {
 			defer func() {
-				err = errors.Join(err, fileutils.RemoveTempDir(nodeModulesFullPath))
+				_ = fileutils.RemoveTempDir(nodeModulesFullPath)
```
gineshkumar: The previous version of this `defer` joined the cleanup error into the returned `err` so a failure to remove `node_modules` was visible to the caller. The refactor replaced that with `_ = fileutils.RemoveTempDir(...)`, silently dropping any cleanup error. This looks incidental to the named-return → inline-return refactor — at minimum, log the failure so it's diagnosable.

```suggestion
		if !nodeModulesDirExists {
			defer func() {
				if cleanupErr := fileutils.RemoveTempDir(nodeModulesFullPath); cleanupErr != nil {
					log.Debug(fmt.Sprintf("yarn install: failed to remove transient node_modules directory '%s': %s", nodeModulesFullPath, cleanupErr.Error()))
				}
			}()
		}
```

commands/audit/auditbasicparams.go
```diff
@@ -28,6 +28,7 @@ type AuditParamsInterface interface {
 	Args() []string
 	InstallCommandName() string
 	InstallCommandArgs() []string
+	SetInstallCommandArgs(installCommandArgs []string) *AuditBasicParams
```
gineshkumar: `SetInstallCommandArgs` is added to `AuditParamsInterface` and `AuditBasicParams` but has no callers anywhere in the diff. Adding a method to a public interface forces every implementor and test double to add a stub. If this setter is for a follow-up PR, consider dropping it here and reintroducing it where it's actually used; if it's added for parity with the other setters, a short comment naming the intended caller would help future maintainers.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+
+// curationNoLockfileError builds an actionable error for when 'yarn install'
+// did not produce yarn.lock. Probes declared direct deps against the curation
+// repo and renders blocked ones in a table. Error text is version-specific:
+// V2 has no lockfile-only install mode; V3+ reaching here means curation is
+// blocking manifests (not just tarballs).
+func curationNoLockfileError(params technologies.BuildInfoBomGeneratorParams, curWd, yarnExecPath, workspaceMemberRel string, installErr error) error {
+	probed, totalProbed := probeBlockedDirectDeps(params, curWd, workspaceMemberRel)
+	outputRef := "table"
+	if params.OutputFormat == "json" {
```
gineshkumar: The literal `"json"` is compared against `params.OutputFormat` here and at `yarn.go:325` and `yarn.go:947`, with a matching `"table"` literal at `yarn.go:288`. Both values already exist as typed constants in `common/format` (`outFormat.Json`, `outFormat.Table`) and are used consistently elsewhere in this repo (e.g. `commands/curation/curationaudit.go:89,489,765`).

Two equally good options:

- Change `BuildInfoBomGeneratorParams.OutputFormat` from `string` to `format.OutputFormat` (it's already a string alias, so the change is purely a type tightening).
- Or import `format` here and compare against `format.Json` / `format.Table` directly.

Either way removes a silent-break risk: a future rename of `outFormat.Json` would not touch these comparisons.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+			root.Details.Dependencies = append(root.Details.Dependencies, bibuildutils.YarnDependencyPointer{Locator: locator})
+			synthesised = append(synthesised, fmt.Sprintf("%s@%s", name, resolvedVer))
+			continue
+		}
+		if isRangeOrTag {
+			unresolvedRanges = append(unresolvedRanges, fmt.Sprintf("%s@%s", name, spec))
+		}
+	}
+	if len(synthesised) > 0 {
+		sort.Strings(synthesised)
```
gineshkumar: This project is on Go 1.25 (`go.mod`). Since Go 1.21 the `slices` package supersedes `sort.Strings` for the common ordered-string case — both expressed and inferred sort behaviour is identical, but `slices.Sort` is generic and avoids the `sort` import here (only used for these three calls at `yarn.go:422,714,720`).

```suggestion
	slices.Sort(synthesised)
```

Same change at `yarn.go:422` and `yarn.go:720`. The `"sort"` import on line 14 can then be dropped in favour of `"slices"`.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+		}
+		name, nameErr := dep.Name()
+		if nameErr != nil || name == "" {
+			continue
+		}
+		resolvedNames[name] = struct{}{}
+	}
+	var synthesised, unresolvedRanges []string
+	for name, spec := range declared {
+		if _, present := resolvedNames[name]; present {
```
gineshkumar, line 738: Could you verify this case-sensitivity assumption before merging? The current code keys `resolvedNames` on whatever `dep.Name()` returns and looks up `declared[name]` using the raw key from `package.json`. Both lookups are case-sensitive, but npm itself treats package names case-insensitively (the registry rejects `MyPackage` if `mypackage` is taken).

I haven't been able to confirm whether yarn normalises the lockfile to lowercase in all cases (V1 vs V2 vs V3 may differ; private registries and historical mixed-case packages like `underscore.string` are the edge cases worth probing). If yarn always lowercases, this is a non-issue and the comment can be resolved.

A quick way to test:

1. Add a fixture `package.json` with a mixed-case dep name (e.g. `"underscore.string": "^3.3.6"`, or any historical package — `Base64` works too).
2. Run the curation flow and check whether:
   - The lockfile entry's `Name()` returns lowercase or original case.
   - `resolvedNames[name]` lookup hits or misses on the `declared` map key.
3. If it misses, a synthetic locator gets appended and the curation walker issues a duplicate HEAD-check against an URL Artifactory may handle inconsistently.

If the test confirms the issue, normalising both sides resolves it:

```suggestion
		resolvedNames[strings.ToLower(name)] = struct{}{}
```

with the corresponding lookup-side change at line 692 (`if _, present := resolvedNames[strings.ToLower(name)]; present {`) and the synthesised locator at line 697 built from the lowercased name to match the `dependenciesMap` dup-check at line 698.

If yarn normalises lockfile keys to lowercase universally, feel free to dismiss.

gineshkumar:
Two new code paths in this PR handle the same upstream condition (a 403 with an unparseable or empty-`Errors` body) but produce very different report rows:

- **yarn V2 probe** (`yarn.go:857,911`): row shows `"Blocked by curation (response could not be parsed)"` — honest.
- **General walker** (`curationaudit.go:1098,1109`): row shows `Policy violations` with empty Policy / Explanation / Recommendation — looks identical to a real policy block.

The general walker runs for every tech (npm, pip, maven, gradle, yarn-V3-success-path, …), so the silent-collapse behaviour is the dominant case across the product. A support engineer triaging a later ticket has no way to tell "real policy block, empty payload" from "403 we couldn't decode."

Worth aligning the two paths before merge.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -180,29 +1112,44 @@ func runYarnInstallAccordingToVersion(curWd, yarnExecPath string, installCommand
 		// When executing 'yarn install...', the node_modules directory is automatically generated.
 		// If it did not exist prior to the 'install' command, we aim to remove it.
 		nodeModulesFullPath := filepath.Join(curWd, nodeModulesRepoName)
-		var nodeModulesDirExists bool
-		nodeModulesDirExists, err = fileutils.IsDirExists(nodeModulesFullPath, false)
+		nodeModulesDirExists, err := fileutils.IsDirExists(nodeModulesFullPath, false)
 		if err != nil {
-			err = fmt.Errorf("failed while checking for existence of node_modules directory: %s", err.Error())
-			return
+			return fmt.Errorf("failed while checking for existence of node_modules directory: %s", err.Error())
```
gineshkumar: `%s, err.Error()` breaks the `errors.Is` / `errors.As` chain — callers can't match a typed sentinel from this install path. Sibling files in the same package (`pnpm/pnpm.go:107`, `npm/npm.go:96`, `nuget/nuget.go:79`, `gem/gem.go:98`, `docker/docker.go:270`, `swift/swiftcommand.go:26`, `cocoapods/podcommand.go:28`) all wrap with `%w` — this is the established convention.

```suggestion
			return fmt.Errorf("failed while checking for existence of node_modules directory: %w", err)
```

The same change is needed at `yarn.go:1049` (`fmt.Errorf("failed to check the existence of '%s' file: %s", yarnLockPath, err.Error())`) and at the `errorutils.CheckErrorf("...%s", installErr.Error())` sites at `yarn.go:306,308,310,332` — each should pass `installErr` to `%w` or `errors.Join` it onto the returned error.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+// verifyYarnVersionSupportedForCuration returns an error for Yarn V1 and V4,
+// which cannot be routed through Artifactory for curation.
+func verifyYarnVersionSupportedForCuration(yarnExecPath, curWd string) error {
+	versionStr, err := bibuildutils.GetVersion(yarnExecPath, curWd)
+	if err != nil {
+		return err
+	}
+	yarnVersion := version.NewVersion(versionStr)
+	if yarnVersion.Compare(yarnV2Version) > 0 || yarnVersion.Compare(yarnV4Version) <= 0 {
+		return errorutils.CheckErrorf("'jf curation-audit' is not supported for Yarn V1 or Yarn V4 (detected: %s). Curation requires Artifactory-resolved installs, which the JFrog CLI Yarn integration only supports for Yarn V2 and V3.", versionStr)
```
gineshkumar: Minor terminology fix on the user-facing error message. The current text generalizes the V2/V3-only restriction to the whole JFrog CLI Yarn integration:

> *"Curation requires Artifactory-resolved installs, which the JFrog CLI Yarn integration only supports for Yarn V2 and V3."*

But the restriction is **curation-specific** — `jf audit` and `jf scan` happily support Yarn V4 (they read `yarn.lock` directly without needing Artifactory resolution). A user reading this error on V4 might reasonably conclude *"the JFrog CLI doesn't work with my yarn version at all,"* when really only `jf ca` is affected.

The internal comment at lines 89-93 actually states the scope correctly — *"The jfrog-cli yarn integration only resolves through Artifactory for Yarn V2/V3"* — the user-facing message just needs to mirror that precision:

```suggestion
		return errorutils.CheckErrorf("'jf curation-audit' is not supported for Yarn V1 or Yarn V4 (detected: %s). Curation requires Artifactory-resolved installs, which the curation flow only routes through Artifactory for Yarn V2 and V3 — 'jf audit' and 'jf scan' continue to support Yarn V4.", versionStr)
```

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+	if httpDetails.Headers == nil {
+		httpDetails.Headers = map[string]string{}
+	}
+	// Mirror the curation walker: this header asks Artifactory to include the
+	// curation policy details in the 403 response body so we can show them.
+	httpDetails.Headers["X-Artifactory-Curation-Request-Waiver"] = "syn"
+
+	var blocked []blockedDirectDep
+	var totalProbed int
+	for _, name := range names {
```
gineshkumar: This loop probes every declared direct dependency **serially**, diverging from the project's established convention for the same operation.

`fetchNodesStatus` in `commands/curation/curationaudit.go:984` already does concurrent HEAD-checks via `parallel.NewBounedRunner(nc.parallelRequests, false)`, bounded by the user-facing `--parallel` flag (`SetParallelRequests`), with `sync.Map` for results and `atomic.Bool`/`atomic.Value` for cancellation and first-error capture. The yarn V2 probe uses none of this — and as a side effect, the user's `--parallel` setting is silently ignored on this fallback path.

Impact: a workspace with 100 declared direct deps at 200ms RTT produces ~40s of probe latency; a monorepo root with several hundred deps exceeds a minute. And this code only runs after `yarn install` has already failed — the user is waiting on a known-degraded run.

Mirror the existing pattern so the probe inherits the convention and the `--parallel` flag works consistently across all tech paths.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+	explanation    string
+	recommendation string
+}
+
+// blockedDepJSONRow mirrors commands/curation.PackageStatus JSON tags so that
+// --format=json output from the V2 no-lockfile probe path uses the same schema
+// as normal curation audit output. Duplicated here (not imported) to avoid the
+// commands/curation ↔ yarn import cycle. Keep these tags in sync with
+// PackageStatus when that struct changes.
+type blockedDepJSONRow struct {
```
gineshkumar, line 384: `blockedDepJSONRow` (lines 368-379) and `blockedDepPolicyJSON` (lines 382-387) duplicate the JSON tags of `commands/curation.PackageStatus` and `commands/curation.Policy`. The code comment acknowledges this and asks future maintainers to keep them in sync manually:

> *"Duplicated here (not imported) to avoid the commands/curation ↔ yarn import cycle. Keep these tags in sync with PackageStatus when that struct changes."*

This is a real risk to the customer-facing JSON contract:

1. The compiler does not enforce field/tag alignment between the two structs.
2. Drift is silent — a renamed JSON field in `PackageStatus` won't produce any error here.
3. The result is **asymmetric customer JSON output**: the same `jf ca` run can emit rows with different field shapes depending on whether they came through the yarn V2 probe path or the general curation walker. Any consumer script that processes `--format=json` will break asymmetrically and be hard to diagnose.

The import cycle is real (`yarn` is consumed by `commands/curation` via the buildinfo tree builders), but it's solvable without duplication. Two options that keep the same shape without the sync risk:

- **Extract the JSON DTOs into a leaf package** (e.g. `commands/curation/types/` or a new `sca/curationtypes`) that neither `commands/curation` nor `yarn` transitively imports. Both then import the leaf package and use the shared struct.
- **Define the JSON shape via an interface** that both `PackageStatus` and the yarn probe rows satisfy, marshalled through a helper in the leaf package.

If neither is feasible right now, at minimum add a `_ = (*PackageStatus)(nil)` compile-time assertion or a comment-driven test that fails when `PackageStatus`'s tag set changes.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+// etc.) does not pollute the machine-readable JSON written to stdout.
+// Mirrors build.RunYarnCommand exactly except for the output destination.
+func runYarnCommandQuiet(executablePath, srcPath string, args ...string) error {
+	command := exec.Command(executablePath, args...)
+	command.Dir = srcPath
+	command.Stdout = io.Discard
+	command.Stderr = io.Discard
+	if err := command.Run(); err != nil {
+		if _, ok := err.(*exec.ExitError); ok {
+			return errors.New(err.Error())
```
gineshkumar: `errors.New(err.Error())` strips the typed `*exec.ExitError` and rebuilds it as a string-only error. The `if _, ok := err.(*exec.ExitError); ok` branch detects the type and then immediately discards it, so callers can no longer `errors.As(err, &exec.ExitError{})` to recover the exit code.

The branch is also redundant with the `return err` two lines below (which preserves the type). Either drop the branch entirely, or wrap with `%w` if a prefix is desired:

```suggestion
		if _, ok := err.(*exec.ExitError); ok {
			return fmt.Errorf("yarn: %w", err)
		}
```

sca/bom/buildinfo/technologies/yarn/yarn.go
gineshkumar: Stronger form: the only consumer of `golang.org/x/exp/maps.Keys(declared)` at line 421 is this `sort.Strings(names)` two lines below. With Go 1.25 in `go.mod`, both collapse to one stdlib call:

```suggestion
	names := slices.Sorted(maps.Keys(declared))
```

Stdlib `maps.Keys` (Go 1.21+) returns an `iter.Seq[string]`; `slices.Sorted` (Go 1.23+) materialises and sorts it in one pass. Same effect, drops the `golang.org/x/exp/maps` import and the separate sort call.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +59,1050 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+		}
+	}
+	yarnVersionStr, versionErr := bibuildutils.GetVersion(yarnExecPath, curWd)
+	if versionErr == nil {
+		yarnVersion := version.NewVersion(yarnVersionStr)
+		isV2 := yarnVersion.Compare(yarnV2Version) <= 0 && yarnVersion.Compare(yarnV3Version) > 0
+		if isV2 {
+			return errorutils.CheckErrorf("'jf curation-audit' against curation repo '%s' could not produce '%s' with Yarn %s — V2 has no lockfile-only install mode, so any blocked package aborts the install before the lockfile is written.%s Remove or replace the blocked direct dependencies in the %s above and re-run 'jf ca'; once they pass curation, install completes and the audit enumerates the full graph. Secondary option: upgrade the project to Yarn V3 ('yarn set version 3.6.4'). V3's '--mode=update-lockfile' writes the lockfile during resolve, so 'jf ca' can audit even while curation blocks tarballs. Underlying yarn error: %s", params.DependenciesRepository, yarn.YarnLockFileName, yarnVersionStr, tableNote, outputRef, installErr.Error())
```
gineshkumar: Use case: a Yarn V2 user with no prior `yarn.lock` runs `jf ca`. `yarn install` aborts on a 403 from a blocked direct dep — the only path V2 users reach this code. The user-facing recommendation here is "remove or replace the blocked direct dependencies."

For waiverable policies (those whose 403 message includes `[waivers allowed]`), the walker path elsewhere in the product offers an in-CLI `Y/n: request a waiver?` prompt and submits via `sendWaiverRequests`. This path doesn't — even though the probe already holds everything `sendWaiverRequests` needs at this scope: same `rtManager`, `rtAuth`, `httpDetails`, target URL (`buildNpmTarballURL` constructs the same shape Artifactory expects waiver GETs against), and `X-Artifactory-Curation-Request-Waiver` header (currently set to `"syn"` here). The `[waivers allowed]` marker is in the same `resp.Errors[0].Message` that `parseProbe403Body` already reads.

Was wiring the waiver prompt into this path considered? If it was deliberately excluded, what's the reason V2-no-lockfile users should have a different recovery UX from walker-path users?

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: Reverted to earlier implementation

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: The case-sensitive lookup is intentional — "express": "^5.2.1" and "Express": "^3.0.1" are treated as distinct packages by design. Adding strings.ToLower normalization would incorrectly collapse them into a single entry and skip curation probing for one of them. Since yarn resolves them as separate lockfile entries, the case-sensitive map keys faithfully represent the dep graph.

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: _ = (*PackageStatus)(nil) suggestion doesn't work here since they don't implement the same interface, but a reflect-based test does the same job
 The reflect test is an honest safety net that CI will catch.

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: The V2 no-lockfile probe path is deliberately waiver-free. It only triggers when yarn install fails entirely because a direct dependency is itself blocked — typically a malicious or policy-violating package. The recommended action is to remove or replace the package, not request an exception. The waiver prompt is reserved for the full walker path, which runs after a successful install and surfaces curation blocks on direct and transitive dependencies where a policy exception may be legitimate.

commands/curation/curationaudit.go
gauriy-tech: Resolved by showing error above table displaying partial results
Screenshot 2026-05-29 at 4 42 02 PM

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -1,23 +1,37 @@
...
+	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
 	"path/filepath"
+	"maps"
```
gineshkumar: `gofmt` reorders this import block (`maps` is out of alphabetical order). _(Source: go.dev/wiki/CodeReviewComments — Gofmt)_

```suggestion
	"io"
	"maps"
	"net/http"
	"os"
	"os/exec"
	"path/filepath"
```

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +62,1062 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+	if len(probed) > 0 {
+		if tableErr := printBlockedDirectDepsTable(probed, totalProbed, params.OutputFormat); tableErr != nil {
+			log.Debug(fmt.Sprintf("yarn curation probe: failed to render blocked deps table: %s", tableErr.Error()))
+		} else {
+		if params.OutputFormat == outFormat.Json {
+			tablePointer = " (listed in the JSON output above)"
+			} else {
+				tablePointer = " (listed in the table above)"
+			}
+		}
```
gineshkumar, line 346: `gofmt` fails here — this `if/else` is mis-indented. _(Source: go.dev/wiki/CodeReviewComments — Gofmt)_

```suggestion
		if tableErr := printBlockedDirectDepsTable(probed, totalProbed, params.OutputFormat); tableErr != nil {
			log.Debug(fmt.Sprintf("yarn curation probe: failed to render blocked deps table: %s", tableErr.Error()))
		} else {
			if params.OutputFormat == outFormat.Json {
				tablePointer = " (listed in the JSON output above)"
			} else {
				tablePointer = " (listed in the table above)"
			}
		}
```

commands/curation/curationaudit_test.go
```diff
@@ -212,6 +212,29 @@ func TestGetNameScopeAndVersion(t *testing.T) {
...
+		version string
+		want    bool
+	}{
+		{"workspace member — typical", "admin-ui-428bae", "0.0.0", true},
+		{"workspace member — root style", "root-workspace-0b6124", "0.0.0", true},
+		{"real package version 0.0.0", "my-pkg", "0.0.0", false},     // no hex suffix
+		{"real package with hex-looking name", "a-1b2c3d", "1.2.3", false}, // wrong version
+		{"Yarn V1 use.local", "my-pkg", "0.0.0-use.local", false},     // caught by earlier check
+		{"suffix too short", "pkg-4abc", "0.0.0", false},              // 4 chars, not 6
+		{"suffix uppercase", "pkg-4ABC12", "0.0.0", false},            // uppercase hex not matched
```
gineshkumar: `gofmt` re-aligns these trailing comments. _(Source: go.dev/wiki/CodeReviewComments — Gofmt)_

```suggestion
		{"real package version 0.0.0", "my-pkg", "0.0.0", false},           // no hex suffix
		{"real package with hex-looking name", "a-1b2c3d", "1.2.3", false}, // wrong version
		{"Yarn V1 use.local", "my-pkg", "0.0.0-use.local", false},          // caught by earlier check
		{"suffix too short", "pkg-4abc", "0.0.0", false},                   // 4 chars, not 6
		{"suffix uppercase", "pkg-4ABC12", "0.0.0", false},                 // uppercase hex not matched
```

`gofmt` also drops the trailing blank line at the end of this file (line 1959) — `gofmt -w commands/curation/curationaudit_test.go sca/bom/buildinfo/technologies/yarn/yarn.go` fixes every site at once.

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +62,1062 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+func runYarnCommandQuiet(executablePath, srcPath string, args ...string) error {
+	command := exec.Command(executablePath, args...)
+	command.Dir = srcPath
+	command.Stdout = io.Discard
+	command.Stderr = io.Discard
+	if err := command.Run(); err != nil {
+		return err
+	}
+	return nil
+}
```
gineshkumar, line 1035: For curation runs, `runYarnCommandQuiet` discards yarn's stderr unconditionally — not only when `--format=json` is active. When `yarn install` fails for a reason other than a curation 403 (registry auth, network, a genuine yarn config error) and no `yarn.lock` is produced, the only thing left in `installErr` is `*exec.ExitError`'s `"exit status 1"`; yarn's actual diagnostic is gone, and that bare string is what `curationNoLockfileError` surfaces. A support engineer reading the failure has nothing to act on.

Consider capturing stderr and folding it into the returned error (or discarding only when the format is JSON):

```suggestion
func runYarnCommandQuiet(executablePath, srcPath string, args ...string) error {
	command := exec.Command(executablePath, args...)
	command.Dir = srcPath
	var stderr bytes.Buffer
	command.Stdout = io.Discard
	command.Stderr = &stderr
	if err := command.Run(); err != nil {
		if msg := strings.TrimSpace(stderr.String()); msg != "" {
			return fmt.Errorf("%w: %s", err, msg)
		}
		return err
	}
	return nil
}
```

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +62,1062 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+		// Re-wrap with actionable context via enumerateAfterCurationInstallError.
+		if params.IsCurationCmd && deferredInstallErr != nil {
+			err = enumerateAfterCurationInstallError(params, currentDir, workspaceMemberRel, deferredInstallErr, err)
+		}
+		return
+	}
+	// Yarn V2+ always emits the project root as "<name>@workspace:.". Prefer
+	// that over build-info-go's heuristic, which can misidentify the root
+	// when package.json has no name field.
+	if workspaceRoot := findYarnWorkspaceRoot(dependenciesMap); workspaceRoot != nil {
```
gineshkumar, line 159: **Open Question:** the `findYarnWorkspaceRoot` override (and the new `if root == nil` guard below it) is *not* gated on `params.IsCurationCmd`, so it now also affects `jf audit` / `jf scan` on Yarn V2+ projects — whereas the workspace re-route earlier in this function is curation-only. For projects with a `name` in `package.json` the result matches build-info-go's heuristic, and for nameless projects it's a genuine fix (and the nil-guard prevents a potential nil-deref in `getXrayDependencyId`), so this looks beneficial. But it's outside the stated "Yarn support for `jf ca`" scope. Was changing the audit/scan root-resolution behaviour intentional, and is it covered by audit/scan regression tests?

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -48,129 +62,1062 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
+// URL against the curation-enabled Artifactory repository. Returns the deps
+// that responded with HTTP 403, parsed for policy details when the body is a
+// recognizable JFrog Curation error. All errors are logged at debug level and
+// swallowed — this is a best-effort diagnostic invoked from an existing fatal
+// error path; partial information is better than no information.
+//
+// probeBlockedDirectDeps HEAD-checks each declared direct dependency against
+// the curation registry. workspaceMemberRel, when non-empty, scopes the probe
+// to a single workspace member's package.json (used with --working-dirs).
+func probeBlockedDirectDeps(params technologies.BuildInfoBomGeneratorParams, curWd, workspaceMemberRel string) ([]blockedDirectDep, int) {
```
gineshkumar, line 416: `probeBlockedDirectDeps` is the heart of the V2 / workspace fallback path, but its orchestration is untested — the parallel HEAD→GET 403 escalation, the mutex accumulation, and the `totalProbed` counter all run only against a live Artifactory. The pure helpers (`parseProbe403Body`, `buildNpmTarballURL`, table/JSON renderers) are well covered, but a regression in the probe loop itself would ship silently. Here's a hermetic test using the project's mock-server helper (verified compiling and passing against this branch):

```go
func TestProbeBlockedDirectDeps(t *testing.T) {
	curationBody := `{"errors":[{"status":403,"message":"Package lodash:4.17.21 download was blocked by JFrog Packages Curation service due to the following policies violated {mal-policy, Malicious package, Package version is malicious, Remove it}."}]}`
	_, serverDetails, _ := coreCommonTests.CreateRtRestsMockServer(t, func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusForbidden)
		if r.Method == http.MethodGet {
			_, _ = w.Write([]byte(curationBody))
		}
	})

	curWd := t.TempDir()
	require.NoError(t, os.WriteFile(filepath.Join(curWd, "package.json"),
		[]byte(`{"name":"root","dependencies":{"lodash":"4.17.21"}}`), 0o644))

	params := technologies.BuildInfoBomGeneratorParams{
		ServerDetails:          serverDetails,
		DependenciesRepository: "tst-yarn-repo",
		ParallelRequests:       1,
	}

	blocked, totalProbed := probeBlockedDirectDeps(params, curWd, "")

	assert.Equal(t, 1, totalProbed)
	if assert.Len(t, blocked, 1) {
		assert.Equal(t, "lodash", blocked[0].name)
		assert.Equal(t, "4.17.21", blocked[0].probedVersion)
		assert.Equal(t, "blocked_policy", blocked[0].reason)
		if assert.Len(t, blocked[0].policies, 1) {
			assert.Equal(t, "mal-policy", blocked[0].policies[0].policy)
			assert.Equal(t, "Malicious package", blocked[0].policies[0].condition)
		}
	}
}
```

Add imports `net/http`, `coreCommonTests "github.com/jfrog/jfrog-cli-core/v2/common/tests"`, and `require`. A `requestToError`-style 500 case and a non-curation-403 (`unknown_403`) case make natural follow-up subtests. Separately, the walker-side `BlockingReasonUnknown` branches at `commands/curation/curationaudit.go:1098-1109` are also untested.

gauriy-tech:
> Two new code paths in this PR handle the same upstream condition (a 403 with an unparseable or empty-`Errors` body) but produce very different report rows:
> 
> * **yarn V2 probe** (`yarn.go:857,911`): row shows `"Blocked by curation (response could not be parsed)"` — honest.
> * **General walker** (`curationaudit.go:1098,1109`): row shows `Policy violations` with empty Policy / Explanation / Recommendation — looks identical to a real policy block.
> 
> The general walker runs for every tech (npm, pip, maven, gradle, yarn-V3-success-path, …), so the silent-collapse behaviour is the dominant case across the product. A support engineer triaging a later ticket has no way to tell "real policy block, empty payload" from "403 we couldn't decode."
> 
> Worth aligning the two paths before merge.

gauriy-tech:
https://github.com/jfrog/jfrog-cli-security/pull/759#issuecomment-4589629447

Fixed. Added a guard after extractPoliciesFromMsg: when blockingReason == BlockingReasonPolicy but len(policies) == 0 (regex found no {policy,...} groups), we now fall back to BlockingReasonUnknown. All three unparseable-403 paths in getBlockedPackageDetails now consistently produce "Blocked by curation (response could not be parsed)", matching the yarn V2 probe path.

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: Fixed, moved "maps" to its correct alphabetical position in the stdlib import block.

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: Fixed, corrected the indentation of the inner if/else inside the else branch.

commands/curation/curationaudit_test.go
gauriy-tech: Fixed — ran gofmt -w on both files; comment alignment, trailing blank line, and the maps import order are all corrected.

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: Applied as suggested — runYarnCommandQuiet now captures stderr into a bytes.Buffer and wraps it into the returned error. A bare exit status 1 will no longer reach curationNoLockfileError; yarn's actual diagnostic travels with the error.

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: Added TestProbeBlockedDirectDeps with three subtests (curation 403 with policy extraction, 500 non-blocking, non-curation 403 with unknown_403 reason) — all passing. The wantTotal for the 500 case is 1 because totalProbed is incremented for any non-nil HEAD response; only non-403s skip the blocked accumulation.

sca/bom/buildinfo/technologies/yarn/yarn.go
gauriy-tech: Yes, the behavior change is intentional and applies to all callers. Added TestRootResolutionAppliesToAuditAndScan (4 subtests) that: (1) pins via source-text inspection that the override is not gated on IsCurationCmd; (2) verifies a named Yarn V2 project resolves correctly; (3) verifies a nameless Yarn V2 project finds its root via the @workspace:. suffix alone, preventing the nil-deref the root == nil guard protects against; (4) verifies Yarn V1 dep maps return nil so build-info-go's heuristic continues to own that path. All pass.

gineshkumar (approved):

utils/techutils/techutils.go
```diff
@@ -429,6 +430,36 @@ func DetectedTechnologiesList() (technologies []string) {
...
+	if err != nil {
+		return
+	}
+	if len(detected) == 0 {
+		return
+	}
+	promoteYarnWorkspaceMembers(detected)
+	techStringsList := DetectedTechnologiesToSlice(detected)
+	log.Info(fmt.Sprintf("Detected: %s.", strings.Join(techStringsList, ", ")))
+	return techStringsList
```
attiasas, line 460: If this is a fix to detecting the actual technology, why do we have a split logic from audit? is it unique to ca flow? or a fix we should adjust DetectTechnologiesDescriptors and all can gain this fix?

commands/curation/curationaudit.go
```diff
@@ -440,7 +489,15 @@ func (ca *CurationAuditCommand) getBuildInfoParamsByTech() (technologies.BuildIn
 	return technologies.BuildInfoBomGeneratorParams{
 		XrayVersion:      ca.GetXrayVersion(),
 		ExclusionPattern: technologies.GetScaExcludePattern(ca.GetConfigProfile(), ca.IsRecursiveScan(), ca.Exclusions()...),
-		Progress:         ca.Progress(),
+		// Suppress the spinner when emitting machine-readable output: the
```
attiasas: what is this chnage for? progressMgr is nil if `CI=true` env var exists, why do we need this to be disabled for JSON?

commands/curation/curationaudit.go
```diff
@@ -312,12 +316,17 @@ func (ca *CurationAuditCommand) Run() (err error) {
...
 			err = errors.Join(err, e)
 		}
 	}
 	if ca.Progress() != nil {
 		err = errors.Join(err, ca.Progress().Quit())
 	}
+	// Print after the spinner has stopped so the message appears on the terminal.
+	if scanErr != nil {
+		log.Error("Curation audit encountered errors while checking some packages; the report below may be incomplete: " + scanErr.Error())
+	}
```
attiasas, line 328: why do we need this log.Error? it will be displayed as `err`...

commands/curation/curationaudit.go
```diff
@@ -312,12 +316,17 @@ func (ca *CurationAuditCommand) Run() (err error) {
 		}
 		// If error returned, continue to print results(if any), and return error at the end.
 		if e := ca.doCurateAudit(results); e != nil {
+			scanErr = errors.Join(scanErr, e)
 			err = errors.Join(err, e)
```
attiasas, line 319:
```suggestion
err = errors.Join(err, fmt.Errorf("Curation audit encountered errors while checking some packages; the report below may be incomplete: %s", err.Error()))
```

you can wrap the error...

commands/curation/curationaudit.go
```diff
@@ -379,11 +388,16 @@ func getPolicyAndConditionId(policy, condition string) string {
...
+	techs := techutils.DetectedTechnologiesListForCurationAudit()
 	if ca.DockerImageName() != "" {
 		log.Debug(fmt.Sprintf("Docker image name '%s' was provided, running Docker curation audit.", ca.DockerImageName()))
 		techs = []string{techutils.Docker.String()}
 	}
+	// Resolve npm→yarn when the project was configured with 'jf yarn-config' (yarn.yaml exists)
+	// but has no yarn.lock/.yarnrc.yml so the file-based detector picked npm instead.
+	for i, tech := range techs {
+		techs[i] = resolveNpmYarnTech(tech)
+	}
```
attiasas, line 399: maybe this can be added as a new tech detection option for techutils (like detecting also .npmrc like files...), not a must

attiasas (changes requested):
Nice job, see my comments

utils/techutils/techutils.go
gauriy-tech: yes it is curation specific as jf ca must resolve to a single tech per dir; audit/scan don't.

jf ca rewrites the registry URL in .yarnrc.yml/.npmrc to the api/curation/audit/<repo>/ endpoint and then runs the actual install through it. So it has to commit to exactly one package manager — npm vs yarn changes which install command runs, which lockfile is produced, and which curation answers come back. On a yarn workspace member that only has a bare package.json, the file-based detector routes it to npm, which would run npm ls and synthesize a dependency set that doesn't match what yarn actually resolved.

Audit/scan never face this "pick one" problem. They call DetectTechnologiesDescriptors directly and aggregate every detected tech into a set, then scan each independently (detectTechnologiesInTarget, audit.go:664). There's no single-resolver install step to get wrong.

commands/curation/curationaudit.go
gauriy-tech: reverted , would have gotten during interactive terminal state with --format=json testing

commands/curation/curationaudit.go
gauriy-tech: It's intentional. scanErr is joined into the returned err, so the framework does log it — but only at the very bottom, after the results table is printed. This log.Error surfaces the "report may be incomplete" notice above the table prefixed with red icon, so a user looking at the results sees the partial-report warning right away instead of having to scroll past the whole table to find out the report isn't complete. This is a product call

commands/curation/curationaudit.go
gauriy-tech: I'd prefer to keep the explicit log.Error rather than wrap it into the returned error. If we only wrap it, the framework surfaces the message at the very bottom — after the entire results table — so the user has to scroll past the whole report to find out it's incomplete. The log.Error prints the "report may be incomplete" notice on top, before the table, which is the heads-up we want.

The returned err still carries the underlying failure (for the non-zero exit code); this line is purely the up-front user-facing warning. Happy to revisit if you feel strongly, but the placement is intentional for UX.

commands/curation/curationaudit.go
gauriy-tech: The reason I kept it out of techutils for now: this resolution keys off the JFrog CLI project config (.jfrog/projects/yarn.yaml vs npm.yaml, written by jf yarn-config/jf npm-config) via project.GetProjectConfFilePath — not off source-tree files like .npmrc/.yarnrc.yml. techutils' detector today only inspects package-manager indicator/descriptor files in the project tree and has no notion of the CLI's project config, so adding this would expand its detection contract.

It's also curation-specific: the npm↔yarn equivalence only matters here because curation routes the actual install through the shared Artifactory npm API and needs to commit to one resolver. Generic detection (audit/scan) doesn't need it.

attiasas (approved):
LGTM

PR #768, opened 2026-05-27
XRAY-138689 - Add Poetry support for jf ca

Phavya-jfrog:

-----

Previously jf ca had no Poetry support — running it on a Poetry project either fell back to the pip code path or produced incomplete results. This PR adds full Poetry (1.x and 2.x) support.

What changed:

Curation install via Artifactory — points Poetry's source at the api/curation/audit/<repo>/ endpoint and runs poetry lock against a temporary copy of the project, so all resolution routes through curation and 403 responses surface blocked packages. The user's pyproject.toml and poetry.lock are never modified. The original source name is preserved in the temp copy so existing poetry.lock entries stay valid.

Smart lock handling — checks whether poetry.lock is missing, stale, or up-to-date before changing the source URL (Poetry 1.x stores the URL in the lock and would otherwise always look stale). Generates, refreshes, or skips the lock accordingly, with automatic fallbacks for both Poetry 1.x and 2.x flag differences.

Blocked package table from poetry.lock — parses both v1 and v2 lock layouts and probes Artifactory for each package so blocked ones show up in the same table users already see for npm/pip.

CVS-blocked detection for Poetry — Poetry reports CVS-hidden versions as "doesn't match any versions" instead of a 403; that pattern is now picked up and rendered as a blocked package.

Graceful 403 handling — Poetry-emitted 403s now render the standard curation-blocked message instead of raw Poetry output.

Minimum Poetry version — 1.2.0 required for curation, with a clear error otherwise.

Testing done is documented here https://jfrog-int.atlassian.net/browse/XRAY-141531


sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,21 +509,221 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+				return false, restoreEnv, wrapPoetryCurationErr(params.IsCurationCmd, lockErr)
+			}
+			log.Debug("Poetry: lock updated")
+		default:
+			log.Debug("Poetry: poetry.lock is up to date — skipping lock")
+		}
+	} else {
+		_, err = executeCommand("poetry", "install")
+	}
+	return false, restoreEnv, nil
```
gineshkumar: **[Critical] Non-curation `poetry install` error is silently swallowed**

`python.go` line 605 reads:

```go
} else {
    _, err = executeCommand("poetry", "install")
}
return false, restoreEnv, nil   // ← `nil` should be `err`
```

The variable `err` is assigned on line 604, then discarded on line 606. Any failure from `poetry install` — network error, missing dependency, auth failure — is silently swallowed when `IsCurationCmd == false`. The caller sees a clean `(false, restoreEnv, nil)` even though the install failed.

**Suggested test** (add to `python_test.go`, same `package python`):

```go
func TestInstallPoetryDepsNonCurationErrorPropagated(t *testing.T) {
	// Place a fake "poetry" on PATH that always exits non-zero.
	// LogExecutableVersion runs "poetry --version" first and ignores failure,
	// so a single failing stub covers the whole non-curation path.
	fakeDir := t.TempDir()
	fakePoetry := filepath.Join(fakeDir, "poetry")
	require.NoError(t, os.WriteFile(fakePoetry,
		[]byte("#!/bin/sh\necho 'install failed' >&2\nexit 1\n"), 0755))
	t.Setenv("PATH", fakeDir+string(os.PathListSeparator)+os.Getenv("PATH"))

	_, _, err := installPoetryDeps(technologies.BuildInfoBomGeneratorParams{
		IsCurationCmd:          false,
		DependenciesRepository: "",
	})

	require.Error(t, err, "non-curation poetry install failure must propagate to the caller")
}
```

go.mod
```diff
@@ -155,8 +155,6 @@ require (
 
 // replace github.com/jfrog/jfrog-cli-core/v2 => github.com/jfrog/jfrog-cli-core/v2 master
 
-//replace github.com/jfrog/jfrog-cli-artifactory => github.com/jfrog/jfrog-cli-artifactory main
-
-// replace github.com/jfrog/build-info-go => github.com/jfrog/build-info-go dev
+// replace github.com/jfrog/jfrog-cli-core/v2 => github.com/gauriy-tech/jfrog-cli-core/v2 v2.0.0-20260526032107-e8995d698251
```
gineshkumar: **[Low] Remove commented personal-fork `replace` directive**

`go.mod` line 158 contains:

```go
// replace github.com/jfrog/jfrog-cli-core/v2 => github.com/gauriy-tech/jfrog-cli-core/v2 v2.0.0-20260526032107-e8995d698251
```

This is a commented-out `replace` directive pointing to a personal fork (`gauriy-tech`). The `//` prefix makes it inert at build time, but it reveals that the PR was developed against a local fork override and the cleanup was incomplete.

Development-time `replace` directives don't belong in merged PRs. Please remove this line before merging.

(The adjacent `// replace github.com/jfrog/jfrog-cli-core/v2 => github.com/jfrog/jfrog-cli-core/v2 master` is a standard JFrog scaffolding comment and can stay.)

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,21 +509,221 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+// source NAME (not URL). If we renamed the source here, an existing lock
+// would suddenly reference a source that no longer exists, Poetry would
+// abort the relock with "Repository '<old-name>' does not exist".
+// Preserving the name keeps the lock valid and lets the normal post-lock
+// pipeline (with HEAD probes against the wheel URLs) run as designed.
+//
+// If pyproject.toml has no [[tool.poetry.source]] at all, we fall back to
+// adding a single entry named after the Artifactory repository so Poetry
+// has somewhere to resolve from.
+func setCurationSourceInPyproject(repoName, repoUrl string, majorVersion int) error {
```
gineshkumar, line 663: **[Medium / Open Question] Viper round-trip fidelity for complex `pyproject.toml` is untested**

`setCurationSourceInPyproject` rewrites the user's `pyproject.toml` via viper's round-trip (`ReadInConfig` → `Set` → `WriteConfig`). The existing tests use simple fixtures. Real-world projects use constructs that viper is known to handle imperfectly.

Here's a fixture adapted from Black's `pyproject.toml` (25.x) that exercises the four scenarios most likely to fail:

```toml
[tool.poetry]
name = "black"
version = "25.1.0"
authors = ["Łukasz Langa <lukasz@langa.pl>"]

[tool.poetry.dependencies]
python = ">=3.9"
tomli = {version = ">=1.1.0", python = "<3.11"}

[tool.poetry.group.dev.dependencies]
pytest = ">=7.2.1"

[[tool.poetry.source]]
name = "pypi-mirror"
url = "https://mycompany.jfrog.io/artifactory/api/pypi/python-virtual/simple"
priority = "primary"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

**What could break:**
1. **Unicode values** — `Łukasz Langa` is a string value (not a key), so viper shouldn't lowercase it, but TOML re-serialization may escape or corrupt non-ASCII characters.
2. **Inline-table dependency markers** — `{version = ">=1.1.0", python = "<3.11"}` — viper may lose the `python` marker or reorder fields inside the inline table.
3. **`[tool.poetry.group.*]` subsections** — viper may flatten `tool.poetry.group.dev.dependencies` into the `tool.poetry` map and lose the group structure, causing `poetry install` to fail on the temp copy.
4. **`priority = "primary"`** — Poetry 1.8+ field; if viper drops it, Poetry treats the source as secondary and may skip it during `poetry lock`.

**Suggested test** (add to `TestSetCurationSourceInPyproject`):

```go
t.Run("complex real-world pyproject round-trip preserves all sections", func(t *testing.T) {
	dir := t.TempDir()
	initial := []byte(`[tool.poetry]
name = "black"
version = "25.1.0"
authors = ["Łukasz Langa <lukasz@langa.pl>"]

[tool.poetry.dependencies]
python = ">=3.9"
tomli = {version = ">=1.1.0", python = "<3.11"}

[tool.poetry.group.dev.dependencies]
pytest = ">=7.2.1"

[[tool.poetry.source]]
name = "pypi-mirror"
url = "https://mycompany.jfrog.io/artifactory/api/pypi/python-virtual/simple"
priority = "primary"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
`)
	pyprojectPath := filepath.Join(dir, pyprojectToml)
	require.NoError(t, os.WriteFile(pyprojectPath, initial, 0600))
	t.Chdir(dir)

	const repoURL = "https://mycompany.jfrog.io/artifactory/api/curation/audit/python-virtual"
	require.NoError(t, setCurationSourceInPyproject("python-virtual", repoURL, 0))

	written, err := os.ReadFile(pyprojectPath)
	require.NoError(t, err)
	out := string(written)

	assert.Contains(t, out, repoURL, "source URL must be rewritten to the curation pass-through")
	assert.Contains(t, out, "Łukasz Langa", "unicode author value must survive viper round-trip")
	assert.Contains(t, out, `<3.11`, "conditional dependency marker must survive")
	assert.Contains(t, out, "dev.dependencies", "group dependency subsection must survive")
	assert.Contains(t, out, "priority", "poetry 1.8+ priority field must survive")
	assert.Contains(t, out, "poetry-core", "build-system block must survive")
})
```

If any of these assertions fail against the current implementation, `poetry lock` will silently operate on a degraded `pyproject.toml`, which may cause spurious lock failures or miss dependency groups in the dependency tree.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,21 +509,221 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+			_, checkErr := executeCommand("poetry", "check", "--lock")
+			if checkErr != nil && strings.Contains(checkErr.Error(), "does not exist") {
+				log.Debug("Poetry: 'poetry check --lock' not supported, falling back to 'poetry lock --check'")
+				_, checkErr = executeCommand("poetry", "lock", "--check")
+			}
+			lockIsStale = checkErr != nil
+			log.Debug(fmt.Sprintf("Poetry: stale check result: stale=%v", lockIsStale))
+		}
+	}
+
```
gineshkumar, line 574: **[Medium] `poetry check --lock` error context is lost when re-lock also fails**

Lines 542–546:

```go
_, checkErr := executeCommand("poetry", "check", "--lock")
if checkErr != nil && strings.Contains(checkErr.Error(), "does not exist") {
    _, checkErr = executeCommand("poetry", "lock", "--check")
}
lockIsStale = checkErr != nil
```

When `poetry check --lock` exits non-zero for a reason unrelated to staleness (malformed `pyproject.toml`, permission error, broken environment), `lockIsStale = true` triggers a re-lock attempt. If that re-lock also fails, only the re-lock error is returned — `checkErr` is discarded. The operator sees a secondary error with no trace of the root cause.

**Suggested test** (add to `python_test.go`, same `package python`):

```go
func TestInstallPoetryDepsLockCheckErrorSurfacedOnRelockFailure(t *testing.T) {
	// Fake poetry: --version succeeds, check --lock fails, lock also fails.
	// Both errors should appear in the returned error chain.
	fakeDir := t.TempDir()
	fakePoetry := filepath.Join(fakeDir, "poetry")
	script := `#!/bin/sh
case "$*" in
  *"--version"*) echo "Poetry (version 1.8.0)"; exit 0 ;;
  *"check"*"--lock"*) echo "Error: SyntaxError in pyproject.toml at line 12" >&2; exit 1 ;;
  *"lock"*) echo "Error: cannot resolve dependencies" >&2; exit 1 ;;
  *) echo "unexpected call: $*" >&2; exit 2 ;;
esac
`
	require.NoError(t, os.WriteFile(fakePoetry, []byte(script), 0755))
	require.NoError(t, os.Chmod(fakePoetry, 0755))
	t.Setenv("PATH", fakeDir+string(os.PathListSeparator)+os.Getenv("PATH"))

	dir := t.TempDir()
	require.NoError(t, os.WriteFile(filepath.Join(dir, poetryLockFile), []byte("# lock\n"), 0600))
	require.NoError(t, os.WriteFile(filepath.Join(dir, pyprojectToml), []byte("[tool.poetry]\nname=\"x\"\n"), 0600))
	t.Chdir(dir)

	_, _, err := installPoetryDeps(technologies.BuildInfoBomGeneratorParams{
		IsCurationCmd: true,
	})
	require.Error(t, err)

	// Today: only the re-lock failure surfaces; SyntaxError is gone.
	// After fix: checkErr is wrapped into the returned error so the root
	// cause appears alongside the re-lock failure.
	assert.Contains(t, err.Error(), "SyntaxError",
		"original check error must appear in the returned error chain")
}
```

**Fix**: wrap `checkErr` into the re-lock error rather than discarding it:

```go
case lockIsStale:
    _, lockErr := executeCommand("poetry", "lock", "--no-update", PoetryNoInteractionFlag)
    ...
    if lockErr != nil {
        return false, restoreEnv, wrapPoetryCurationErr(params.IsCurationCmd, errors.Join(checkErr, lockErr))
    }
```

No string matching against poetry output needed — `checkErr` is simply preserved in the error chain regardless of why the check failed.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,21 +509,221 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+			// Overwrite [[tool.poetry.source]] in the temp pyproject.toml with the curation
+			// pass-through URL.
+			if err = setCurationSourceInPyproject(params.DependenciesRepository, baseUrl, poetryMajor); err != nil {
 				return false, restoreEnv, err
 			}
 		}
+		if password != "" {
+			if params.IsCurationCmd {
+				// Set credentials in Poetry's global config (used by Poetry 2.x).
+				if err = artifactoryutils.RunPoetryConfig(baseUrl, username, password, params.DependenciesRepository); err != nil {
```
gineshkumar: **[Medium] Global Poetry config mutation is not safe for concurrent `jf ca` invocations**

`RunPoetryConfig` (called at line 563) writes credentials to the machine-wide `~/.config/pypoetry/config.toml`:

```
poetry config repositories.<repo-name> <url>
poetry config http-basic.<repo-name> <user> <pass>
```

The `restoreEnv` closure (lines 566–570) unconditionally unsets these keys when the install finishes:

```go
restoreEnv = func() error {
    _, e1 := executeCommand("poetry", "config", "--unset", "http-basic."+params.DependenciesRepository)
    _, e2 := executeCommand("poetry", "config", "--unset", "repositories."+params.DependenciesRepository)
    return errors.Join(e1, e2)
}
```

When two `jf ca` invocations run concurrently — a common CI matrix scenario — the following sequence is possible:

```
Goroutine A (repo-a)                Goroutine B (repo-b, same binary)
────────────────────────────────────────────────────────────────────
RunPoetryConfig("repo-a", ...)
  → sets http-basic.repo-a
                                    RunPoetryConfig("repo-b", ...)
                                      → sets http-basic.repo-b
A's restoreEnv fires (job A done)
  → unsets http-basic.repo-a
  → unsets repositories.repo-a
                                    poetry lock  ← runs now
                                    ← if B resolves through repo-a (e.g.
                                       same Artifactory instance, different
                                       repo name mapped to same host),
                                       credentials are now gone → 401
```

**Question for the developer**: Was a project-local config considered?

Poetry 1.2+ supports `--local` for project-scoped config:

```sh
poetry config --local http-basic.<name> <user> <pass>
```

This writes to `.poetry.toml` alongside `pyproject.toml`. Since `jf ca` already operates in a temp copy of the project, a project-local config would be fully isolated per invocation and eliminate the global mutation risk.

Alternatively, environment variables (`POETRY_HTTP_BASIC_<REPO>_USERNAME` / `POETRY_HTTP_BASIC_<REPO>_PASSWORD`) are process-scoped and require no cleanup.

Was there a specific reason the global config was chosen over `--local` or env vars?

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +258,241 @@ type pypiMetaData struct {
...
+			}
+		}
+		if strings.HasPrefix(line, "[") {
+			flush()
+			inMetadataFiles = line == "[metadata.files]"
+			currentMetaPkg = ""
+			continue
+		}
+		// lock v1.x: files live in [metadata.files] as  pkgname = [{file = "..."},]
+		if inMetadataFiles {
```
gineshkumar, line 470: **[Low] v1 `[metadata.files]` drops files for quoted/dotted package names**

In Poetry lock v1, packages with dotted names (e.g. `zope.interface`, `oslo.utils`) appear with quoted keys in `[metadata.files]`:

```toml
[metadata.files]
"zope.interface" = [
    {file = "zope.interface-5.0.0.tar.gz", hash = "sha256:aaa"},
    {file = "zope.interface-5.0.0-cp39-cp39-linux_x86_64.whl", hash = "sha256:bbb"},
]
```

`parsePoetryLockPackages` extracts the key with:

```go
currentMetaPkg = strings.ToLower(strings.TrimSpace(strings.SplitN(line, "=", 2)[0]))
```

When the key is quoted, `SplitN` returns `"zope.interface"` (including the double-quote characters). After `strings.ToLower` the lookup is `nameToIdx['"zope.interface"']` — the quotes are literal characters in the map key — which finds nothing. The file list for that package is silently dropped, and the package is excluded from the blocked-package map.

**Suggested test** (add as a new `t.Run` inside `TestParsePoetryLockPackages`):

```go
t.Run("v1 quoted dotted key in metadata.files", func(t *testing.T) {
	fixture := []byte(`[[package]]
name = "zope.interface"
version = "5.0.0"

[metadata]
lock-version = "1.1"

[metadata.files]
"zope.interface" = [
    {file = "zope.interface-5.0.0.tar.gz", hash = "sha256:aaa"},
    {file = "zope.interface-5.0.0-cp39-cp39-linux_x86_64.whl", hash = "sha256:bbb"},
]
`)
	got := parsePoetryLockPackages(fixture)
	require.Len(t, got, 1)
	assert.Equal(t, "zope.interface", got[0].Name)
	// Today: got[0].Files is empty — the quoted key is not matched.
	// After fix: both files must be present.
	assert.ElementsMatch(t, []string{
		"zope.interface-5.0.0.tar.gz",
		"zope.interface-5.0.0-cp39-cp39-linux_x86_64.whl",
	}, got[0].Files,
		"files for a dotted package with a quoted key in [metadata.files] must be collected")
})
```

**Fix**: strip surrounding double-quotes from the extracted key before the lookup:

```go
// Before:
currentMetaPkg = strings.ToLower(strings.TrimSpace(strings.SplitN(line, "=", 2)[0]))

// After:
raw := strings.TrimSpace(strings.SplitN(line, "=", 2)[0])
currentMetaPkg = strings.ToLower(strings.Trim(raw, `"`))
```

Real packages affected: `zope.interface`, `oslo.utils`, `oslo.config`, `backports.ssl-match-hostname`, `importlib-metadata` (in some lock versions), and any package whose canonical PyPI name contains a `.`.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,21 +509,221 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+	}
+	return names
+}
+
+func validateMinimumPoetryVersion(minVersion string) (int, error) {
+	out, err := executeCommand("poetry", "--version")
+	if err != nil {
+		log.Debug(fmt.Sprintf("Poetry is not installed or not on PATH: %s", err.Error()))
+		return 0, errorutils.CheckErrorf("JFrog CLI poetry curation requires Poetry to be installed (version %s or higher).", minVersion)
+	}
```
gineshkumar, line 757: **[Nit] Inconsistent casing in `validateMinimumPoetryVersion` error messages**

```go
"JFrog CLI poetry curation requires Poetry to be installed (version %s or higher)."
"JFrog CLI poetry curation requires poetry version %s or higher."
"JFrog CLI poetry curation requires poetry version %s or higher. The Current version is: %s"
```

Two issues:
- "Poetry" (capitalised, message 1) vs "poetry" (lowercase, messages 2/3) — pick one and apply consistently
- "The Current version" — spurious capital on "Current"

Also: `!version.NewVersion(v).AtLeast(minVersion)` reads more naturally than `version.NewVersion(v).Compare(minVersion) > 0` and matches the phrasing of the error message.

go.mod
```diff
@@ -11,16 +11,17 @@ require (
...
-	github.com/jfrog/build-info-go v1.13.1-0.20260521104402-1e35b9b5b0c6
+	github.com/jfrog/build-info-go v1.13.1-0.20260528065004-80409c046540
 	github.com/jfrog/froggit-go v1.22.0
 	github.com/jfrog/gofrog v1.7.6
 	github.com/jfrog/jfrog-apps-config v1.0.1
-	github.com/jfrog/jfrog-cli-artifactory v0.8.1-0.20260526100850-f85282fe6d9b
+	github.com/jfrog/jfrog-cli-artifactory v0.8.1-0.20260528073225-e2d59f90c8c6
 	github.com/jfrog/jfrog-cli-core/v2 v2.60.1-0.20260522091649-43f236276873
-	github.com/jfrog/jfrog-client-go v1.55.1-0.20260522071027-8b60a715d6e4
+	github.com/jfrog/jfrog-client-go v1.55.1-0.20260508101905-a17af78a38d7
```
gineshkumar: **[High] `go.mod` downgrades `jfrog-client-go` vs `origin/dev`**

This PR's `go.mod` (line 20) pins an older pseudo-version of `jfrog-client-go` than the current `dev` branch:

```diff
# go.mod line 20
- github.com/jfrog/jfrog-client-go v1.55.1-0.20260528115006-6ca9682a3255  ← origin/dev today
+ github.com/jfrog/jfrog-client-go v1.55.1-0.20260508101905-a17af78a38d7  ← this PR
```

The pseudo-version timestamp suffix encodes the commit date:
- PR head: `20260508` — May 8, 2026
- `origin/dev`: `20260528` — May 28, 2026
- Delta: **~20 days of `jfrog-client-go` commits missing from this PR**

The PR branch likely diverged from `dev` before the dependency bump landed. A rebase onto current `dev` will resolve this.

sca/bom/buildinfo/technologies/python/python.go
Phavya-jfrog: jf ca intentionally mirrors jf audit's supported formats, which are determined by build-info-go. build-info-go only supports [tool.poetry.dependencies] and the old [tool.poetry.dev-dependencies]. Constructs beyond that scope — [tool.poetry.group.dev.dependencies], inline table markers, unicode values, and priority — are not supported by build-info-go and therefore outside the scope of this PR.

sca/bom/buildinfo/technologies/python/python_test.go
```diff
@@ -306,3 +308,540 @@ func TestGetPipInstallArgs(t *testing.T) {
...
+	script := `#!/bin/sh
+case "$*" in
+  *"--version"*) echo "Poetry (version 1.8.0)"; exit 0 ;;
+  *"check"*"--lock"*) echo "Error: SyntaxError in pyproject.toml at line 12" >&2; exit 1 ;;
+  *"lock"*) echo "Error: cannot resolve dependencies" >&2; exit 1 ;;
+  *) echo "unexpected call: $*" >&2; exit 2 ;;
+esac
+`
+	require.NoError(t, os.WriteFile(fakePoetry, []byte(script), 0755))
+	require.NoError(t, os.Chmod(fakePoetry, 0755))
```
gineshkumar: **[Nit] Redundant `os.Chmod` immediately after `os.WriteFile` with the same mode**

`os.WriteFile` already creates the file with mode `0755`. The subsequent `os.Chmod(fakePoetry, 0755)` is a no-op. The sibling test `TestInstallPoetryDepsNonCurationErrorPropagated` uses only `os.WriteFile`.

```suggestion
	require.NoError(t, os.WriteFile(fakePoetry, []byte(script), 0755))
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +258,242 @@ type pypiMetaData struct {
...
+	for _, m := range hrefMatches {
+		candidate, _, _ := strings.Cut(m[1], "#")
+		if _, ok := wanted[path.Base(candidate)]; ok {
+			return candidate
+		}
+	}
+	return ""
+}
+
+func NormalizePypiName(name string) string {
```
gineshkumar, line 391: **[Low] Two PEP 503 normalization functions in the same package**

This PR adds `NormalizePypiName` to `python.go`. The same package already has `normalizePyPIName` in `python_cvs_fallback.go`. Both implement PEP 503 name normalization (collapse consecutive `-`, `_`, `.` separators into a single `-`, lowercase):

```go
// python.go:364 — new in this PR, loop-based
func NormalizePypiName(name string) string {
    name = strings.ToLower(name)
    var b strings.Builder
    prevSep := false
    for _, r := range name {
        if r == '-' || r == '_' || r == '.' {
            if !prevSep { b.WriteByte('-'); prevSep = true }
            continue
        }
        b.WriteRune(r); prevSep = false
    }
    return b.String()
}

// python_cvs_fallback.go:49 — pre-existing, regex-based
var pypiNameNormalizeRegex = regexp.MustCompile(`[-_.]+`)
func normalizePyPIName(name string) string {
    return strings.ToLower(pypiNameNormalizeRegex.ReplaceAllString(name, "-"))
}
```

Both produce identical results today. The DRY violation means that if PEP 503 normalization ever changes, only one gets updated.

**Fix:** Replace calls to `normalizePyPIName` in `python_cvs_fallback.go` with `NormalizePypiName` and delete `normalizePyPIName`.

commands/curation/curationaudit.go
```diff
@@ -1114,7 +1118,7 @@ func getPythonNameVersion(id string, downloadUrlsMap map[string]string) (downloa
 	if dl, ok := downloadUrlsMap[normalizedId]; ok {
```
gineshkumar, line 1646: **[Low] Normalization fallback path in `getPythonNameVersion` is untested**

`Test_getPythonNameVersion` covers exact-match lookups but has no test case for the normalization fallback path (lines 1116–1118):

```go
// curationaudit.go:1116-1118 — not exercised by any test case
normalizedName := strings.ReplaceAll(strings.ToLower(strings.TrimSpace(parts[0])), "-", "_")
normalizedId := python.PythonPackageTypeIdentifier + normalizedName + ":" + strings.TrimSpace(parts[1])
if dl, ok := downloadUrlsMap[normalizedId]; ok {
    downloadUrls = []string{dl}
```

This path is reachable in practice: the dependency graph node ID can carry the raw lock name (e.g. `pypi://Flask-Babel:1.0`), while the download-URLs map key uses the normalized form (`pypi://flask_babel:1.0`). The exact lookup fails; the fallback resolves it.

**Suggested test case** to add to `Test_getPythonNameVersion`:

```go
{
    name:             "hyphenated name resolved via normalization fallback",
    id:               "pypi://Flask-Babel:1.0",
    downloadUrlsMap:  map[string]string{"pypi://flask_babel:1.0": exampleUrl},
    wantDownloadUrls: []string{exampleUrl},
    wantName:         "Flask-Babel",
    wantVersion:      "1.0",
},
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,23 +510,245 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+		}
+		log.Info(fmt.Sprintf("Configured tool.poetry.source name:%q url:%q for curation", n, repoUrl))
+	}
+	if err = os.WriteFile(absPath, []byte(buf.String()), 0600); err != nil {
+		return errorutils.CheckErrorf("failed to write %s: %s", pyprojectToml, err)
+	}
+	return nil
+}
+
+func stripPoetrySourceBlocks(content string) string {
```
gineshkumar, line 702: **[Low] `stripPoetrySourceBlocks` has no unit test**

`stripPoetrySourceBlocks` is a pure string function that strips all `[[tool.poetry.source]]` TOML blocks before `setCurationSourceInPyproject` rewrites them. It has several edge cases:

- No source blocks in the file
- Single source block
- Two consecutive source blocks (e.g. both `[[tool.poetry.source]]` entries, then `[tool.poetry.dependencies]`)
- Source block at end of file (no trailing section)

`extractPoetrySourceNames` (the viper-based companion) has `TestExtractPoetrySourceNames` covering its cases. `stripPoetrySourceBlocks` has nothing.

Without a unit test, a regression in the stripping logic (e.g. accidentally consuming the first line of the next section when two source blocks are adjacent) would only surface in an end-to-end integration test with a real Poetry installation.

**Suggested test** (table-driven):

```go
func TestStripPoetrySourceBlocks(t *testing.T) {
    cases := []struct{ name, in, want string }{
        {
            name: "no source blocks — content unchanged",
            in:   "[tool.poetry]\nname = \"x\"\n",
            want: "[tool.poetry]\nname = \"x\"\n",
        },
        {
            name: "single source block stripped",
            in:   "[tool.poetry]\nname = \"x\"\n[[tool.poetry.source]]\nname = \"pypi\"\nurl = \"https://pypi.org\"\n[tool.poetry.dependencies]\npython = \"^3.9\"\n",
            want: "[tool.poetry]\nname = \"x\"\n[tool.poetry.dependencies]\npython = \"^3.9\"\n",
        },
        {
            name: "source block at EOF",
            in:   "[tool.poetry]\nname = \"x\"\n[[tool.poetry.source]]\nname = \"pypi\"\n",
            want: "[tool.poetry]\nname = \"x\"\n",
        },
    }
    for _, tc := range cases {
        t.Run(tc.name, func(t *testing.T) {
            assert.Equal(t, tc.want, stripPoetrySourceBlocks(tc.in))
        })
    }
}
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +258,242 @@ type pypiMetaData struct {
...
+	return urls, nil
+}
+
+// buildPoetryDownloadUrl is the Poetry equivalent of npm's buildNpmDownloadUrl: given a
+// package, it returns the absolute Artifactory download URL that curation will HEAD against.
+// It does so by fetching the package's simple-index HTML and matching one of the filenames
+// recorded in poetry.lock against the listed <a href>s.
+func buildPoetryDownloadUrl(rtManager artifactory.ArtifactoryServicesManager, clientDetails *httputils.HttpClientDetails, artiUrl, repository string, pkg poetryLockPackage) (string, error) {
+	normalized := NormalizePypiName(pkg.Name)
+	simpleIndexUrl := fmt.Sprintf("%s/api/pypi/%s/simple/%s/", artiUrl, repository, normalized)
```
gineshkumar, line 342: **[Medium] `buildPoetryDownloadUrl` bypasses the CVS pass-through endpoint**

`buildPoetryDownloadUrl` hardcodes the raw PyPI simple-index URL:

```go
// python.go:315
simpleIndexUrl := fmt.Sprintf("%s/api/pypi/%s/simple/%s/", artiUrl, repository, normalized)
```

Every other package manager routes its index/resolution requests through the curation pass-through layer when `IsCurationCmd=true`. For pip, `GetPypiRepoUrlWithCredentials` does this automatically:

```go
// jfrog-cli-artifactory — GetPypiRepoUrlWithCredentials
if isCurationCmd {
    rtUrl = rtUrl.JoinPath(coreutils.CurationPassThroughApi)  // "api/curation/audit/"
}
rtUrl = rtUrl.JoinPath("api/pypi", repository, "simple")
// result: <artiUrl>/api/curation/audit/api/pypi/<repo>/simple/
```

`buildPoetryDownloadUrlsMap` is **only called in curation mode** (guarded at line 195 by `if !params.IsCurationCmd { return }`). Its simple-index GETs should always use the curation URL — they never run outside `jf ca`.

Without the `api/curation/audit/` prefix, the simple-index requests bypass the CVS intercept layer. When Artifactory's CVS enforces policy at the index level (hiding or blocking package entries in the simple-index), Poetry will miss those signals entirely.

**Fix:** pass `isCurationCmd: true` into `buildPoetryDownloadUrl` and replace the `fmt.Sprintf` with `GetPypiRepoUrlWithCredentials(serverDetails, repository, true)` to derive the base URL — the same helper pip already uses.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +258,242 @@ type pypiMetaData struct {
 	Version string `json:"version"`
 }
 
+type poetryLockPackage struct {
+	Name    string
+	Version string
+	Files   []string
+}
+
+func buildPoetryDownloadUrlsMap(serverDetails *config.ServerDetails, repository string) (map[string]string, error) {
```
gineshkumar, line 270: **[Low] Poetry's simple-index GETs are sequential and outside the parallel curation-check pool**

The curation audit's parallel phase (`fetchNodeStatus`, bounded runner of 10 goroutines) HEAD-checks artifact download URLs. All package managers feed into that phase.

Poetry adds a **serial pre-phase** before it: `buildPoetryDownloadUrlsMap` makes one GET per lock entry sequentially to resolve filenames to download URLs:

```go
// python.go:296 — one blocking HTTP GET per package, no concurrency
for _, pkg := range packages {
    downloadUrl, lookupErr := buildPoetryDownloadUrl(rtManager, &httpClientDetails, artiUrl, repository, pkg)
```

Other package managers have no equivalent pre-phase:
- **pip** reads download URLs from a local `--report` JSON written during `pip install` — zero extra HTTP calls.
- **npm, gems, go, nuget, maven, gradle** derive download URLs from string operations on the component ID.

For a project with 100 lock entries at 50 ms RTT, this is ~5 s of sequential HTTP before the parallel HEAD phase even starts. The `rtManager.Client().SendGet` is thread-safe.

**Suggestion:** a small worker pool (e.g. `errgroup` + 8 goroutines) would reduce this to `ceil(N/8) × RTT`. If that's deferred, add a comment above the loop documenting the O(N × RTT) cost so the next reader understands it's structural, not an oversight.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +258,242 @@ type pypiMetaData struct {
...
+			flush()
+			inMetadataFiles = line == "[metadata.files]"
+			currentMetaPkg = ""
+			continue
+		}
+		// lock v1.x: files live in [metadata.files] as  pkgname = [{file = "..."},]
+		if inMetadataFiles {
+			if strings.Contains(line, "= [") {
+				raw := strings.TrimSpace(strings.SplitN(line, "=", 2)[0])
+				currentMetaPkg = strings.ToLower(strings.Trim(raw, `"`))
```
gineshkumar: **[Nit] Inner `raw` shadows the outer loop variable — rename for clarity**

The loop iterates over `raw` (each unprocessed line). Inside the `inMetadataFiles` block, a second `raw :=` is declared for the TOML key's left-hand side — a completely different thing with the same name. `go vet -shadow` flags this.

```suggestion
			lhs := strings.TrimSpace(strings.SplitN(line, "=", 2)[0])
			currentMetaPkg = strings.ToLower(strings.Trim(lhs, `"`))
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,23 +510,245 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+		}
+	} else {
+		_, err = executeCommand("poetry", "install")
 	}
-	// Run 'poetry install'
-	_, err = executeCommand("poetry", "install")
 	return false, restoreEnv, err
 }
 
+func wrapPoetryCurationErr(isCurationCmd bool, lockErr error) error {
```
gineshkumar: **[Low] `wrapPoetryCurationErr` takes `isCurationCmd bool` that is always `true` at every call site**

Both call sites are structurally inside `if params.IsCurationCmd { switch { ... } }`:

```go
// python.go:579-596
if params.IsCurationCmd {
    switch {
    case lockNeedsGenerate:
        ...
        return false, restoreEnv, wrapPoetryCurationErr(params.IsCurationCmd, lockErr)   // always true
    case lockIsStale:
        ...
        return false, restoreEnv, wrapPoetryCurationErr(params.IsCurationCmd, errors.Join(...))  // always true
    }
}
```

The `isCurationCmd` branch inside `wrapPoetryCurationErr` (`if isCurationCmd && isCvsVersionFilteredOutput(...)`) will never be false. The parameter was likely carried over from `installPipDeps` where `IsCurationCmd` can genuinely vary. Here it adds noise without value.

**Fix:** remove the `isCurationCmd bool` parameter from `wrapPoetryCurationErr` and inline the body unconditionally, or add a comment at the call sites explaining why the bool is always true.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,23 +510,245 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+	v.SetConfigFile(absPath)
+	if err = v.ReadInConfig(); err != nil {
+		return errorutils.CheckErrorf("failed to read %s: %s", pyprojectToml, err)
+	}
+
+	names := extractPoetrySourceNames(v.Get("tool.poetry.source"))
+	if len(names) == 0 {
+		names = []string{repoName}
+	}
+	raw, err := os.ReadFile(absPath)
```
gineshkumar, line 680: **[Low] `setCurationSourceInPyproject` reads `pyproject.toml` twice — document why**

The file is read once by viper (line 646) for structured parsing, then again by `os.ReadFile` (line 654) for raw byte manipulation:

```go
v := viper.New()
v.SetConfigFile(absPath)
v.ReadInConfig()                    // read #1 — viper parses TOML structure

names := extractPoetrySourceNames(v.Get("tool.poetry.source"))
...
raw, err := os.ReadFile(absPath)    // read #2 — raw bytes for string-based rewrite
```

The dual-read is intentional: viper is used for structured key extraction, while the raw bytes path preserves comments and formatting that viper would destroy on marshal. But without a comment, the next reader will try to consolidate to a single viper round-trip and silently corrupt the file (losing all TOML comments and formatting).

**Fix:** add a comment at the `os.ReadFile` call explaining the two-reader design:

```go
// viper was used above only to extract source names — it cannot faithfully
// round-trip TOML with comments preserved. Read the raw bytes separately
// for the string-based rewrite below.
raw, err := os.ReadFile(absPath)
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,23 +510,245 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+func validateMinimumPoetryVersion(minVersion string) (int, error) {
+	out, err := executeCommand("poetry", "--version")
+	if err != nil {
+		log.Debug(fmt.Sprintf("Poetry is not installed or not on PATH: %s", err.Error()))
+		return 0, errorutils.CheckErrorf("JFrog CLI poetry curation requires Poetry %s or higher to be installed.", minVersion)
+	}
+	v := parsePoetryVersion(out)
+	if v == "" {
+		log.Debug(fmt.Sprintf("Could not parse Poetry version from output: %q", out))
+		return 0, errorutils.CheckErrorf("JFrog CLI poetry curation requires Poetry %s or higher to be installed.", minVersion)
```
gineshkumar: **[Low] `validateMinimumPoetryVersion` returns the same error string for two distinct failures**

Two different failure modes produce identical user-facing messages:

```go
// Case 1: poetry binary not found on PATH
if err != nil {
    return 0, errorutils.CheckErrorf("JFrog CLI poetry curation requires Poetry %s or higher to be installed.", minVersion)
}

// Case 2: poetry found but `poetry --version` output is unrecognisable
v := parsePoetryVersion(out)
if v == "" {
    return 0, errorutils.CheckErrorf("JFrog CLI poetry curation requires Poetry %s or higher to be installed.", minVersion)
}
```

A user with a broken Poetry installation (binary on PATH but producing unexpected output) gets told to "install Poetry" instead of "your Poetry binary produced unrecognisable output". These are operationally different problems.

**Fix:** give the second case a distinct message:

```go
if v == "" {
    return 0, errorutils.CheckErrorf(
        "could not parse Poetry version from output %q — ensure Poetry %s or higher is installed correctly",
        out, minVersion,
    )
}
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +258,242 @@ type pypiMetaData struct {
...
+		}
+		if strings.HasPrefix(line, "[") {
+			flush()
+			inMetadataFiles = line == "[metadata.files]"
+			currentMetaPkg = ""
+			continue
+		}
+		// lock v1.x: files live in [metadata.files] as  pkgname = [{file = "..."},]
+		if inMetadataFiles {
+			if strings.Contains(line, "= [") {
```
gineshkumar: **[Low] `strings.Contains(line, "= [")` is an imprecise sentinel for v1 metadata.files entries**

The check looks for `= [` anywhere in the line to detect a v1 `[metadata.files]` key-value assignment:

```go
if strings.Contains(line, "= [") {
    lhs := strings.TrimSpace(strings.SplitN(line, "=", 2)[0])
    currentMetaPkg = strings.ToLower(strings.Trim(lhs, `"`))
```

A TOML value that happens to contain `= [` (e.g. a package whose description leaks into the files block, or a malformed lock) would produce a spurious `currentMetaPkg` update and silently drop files for a real package. The correct guard is to verify the `=` appears before `[`:

```go
if eqIdx := strings.Index(line, "="); eqIdx > 0 && strings.Contains(line[eqIdx:], "[") {
```

**Test that demonstrates the false positive:**

```go
func TestParsePoetryLockPackagesSpuriousContains(t *testing.T) {
    // A [metadata.files] block where a value line happens to contain "= ["
    // before the real package assignment. Under the current strings.Contains
    // guard the spurious line overwrites currentMetaPkg.
    fixture := []byte(`[[package]]
name = "flask"
version = "2.0.0"

[metadata]
lock-version = "1.1"

[metadata.files]
flask = [
    {file = "Flask-2.0.0.tar.gz", hash = "sha256:abc"},
]
`)
    // Inject a comment-like line that contains "= [" but is not a package key.
    // The parser skips blank lines and "#" lines but not arbitrary text inside
    // a file-list value, so a sufficiently creative lock value can trigger this.
    got := parsePoetryLockPackages(fixture)
    require.Len(t, got, 1)
    assert.ElementsMatch(t, []string{"Flask-2.0.0.tar.gz"}, got[0].Files,
        "files must be attributed to flask, not to a spurious metadata line")
}
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +258,242 @@ type pypiMetaData struct {
...
+	packages, err := readPoetryLockIfExists()
+	if err != nil {
+		return nil, err
+	}
+	log.Debug(fmt.Sprintf("Poetry: parsed %d package entries from poetry.lock", len(packages)))
+	rtAuth, err := serverDetails.CreateArtAuthConfig()
+	if err != nil {
+		return nil, err
+	}
+	rtManager, err := rtUtils.CreateServiceManager(serverDetails, 2, 0, false)
```
gineshkumar, line 286: **[Nit] Magic numbers in `CreateServiceManager(serverDetails, 2, 0, false)` — document or name them**

```go
rtManager, err := rtUtils.CreateServiceManager(serverDetails, 2, 0, false)
```

`2`, `0`, and `false` are opaque. Without reading `CreateServiceManager`'s signature, it's impossible to know what these control (retries? timeout? dry-run?).

```suggestion
	// 2 retries, 0 retry-wait-millis, not dry-run.
	rtManager, err := rtUtils.CreateServiceManager(serverDetails, 2, 0, false)
```

sca/bom/buildinfo/technologies/python/python.go
Phavya-jfrog: Pass-through is used to get the metadata
/simple url is used to get the download URL, and the raw download URL is then used for HEAD request

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +262,253 @@ type pypiMetaData struct {
...
+			compId := PythonPackageTypeIdentifier + normalizedName + ":" + pkg.Version
+			mu.Lock()
+			urls[compId] = downloadUrl
+			mu.Unlock()
+			return nil
+		})
+	}
+	_ = g.Wait()
+
+	log.Debug(fmt.Sprintf("Poetry: resolved %d download URLs (skipped %d entries with no files)", len(urls), skipped))
```
gineshkumar: **[Medium] Silent curation bypass on partial or total URL-resolution failure**

`buildPoetryDownloadUrlsMap` parallelises simple-index lookups, but every goroutine swallows its own error:

```go
if lookupErr != nil {
    log.Debug(fmt.Sprintf("Poetry: could not resolve download URL for %s:%s: %v", ...))
    return nil  // error eaten; g.Wait() always returns nil
}
```

Two failure modes are silently hidden from the user:

1. **Total failure** — Artifactory transiently unreachable → `urls` is empty → all packages bypass the HEAD-check.
2. **Partial failure** — some simple-index GETs fail → those packages are absent from `urls` → they bypass the HEAD-check too.

The per-goroutine `log.Debug` is invisible at the default log level. A user running `jf ca` gets no signal that part of their dependency graph was not audited.

**Fix:** after `g.Wait()`, compare resolved count to expected and warn on any shortfall:

```go
_ = g.Wait()
expected := len(packages) - skipped
resolved := len(urls)
if resolved < expected {
    log.Warn(fmt.Sprintf(
        "Poetry: resolved download URLs for %d/%d packages — %d package(s) will not be HEAD-checked by curation. "+
            "Re-run with JFROG_CLI_LOG_LEVEL=DEBUG to see per-package resolution errors.",
        resolved, expected, expected-resolved))
}
log.Debug(fmt.Sprintf("Poetry: resolved %d download URLs (skipped %d entries with no files)", resolved, skipped))
```

This catches both `0/N` (total blackout) and `K/N` (partial miss) without changing the error-tolerant behaviour of the parallel loop itself.

gineshkumar:
**[Question] V3+ non-curation path — which `--mode` is intended here?**

Lines 1195–1197 append two `--mode` flags for the non-curation path:

```go
if isCurationCmd {
    installCommandArgs = append(installCommandArgs, v3UpdateLockfileFlag)   // --mode=update-lockfile
} else {
    installCommandArgs = append(installCommandArgs, v3UpdateLockfileFlag, v3SkipBuildFlag) // both
}
```

Yarn's clipanion parser takes the **last** `--mode` value, so the effective command for non-curation V3+ today is:

```
yarn install --mode=update-lockfile --mode=skip-build
             ^^^^^^^^^^^^^^^^^^^^^^ ignored (overridden)
```

Two questions:

1. Is `--mode=skip-build` the intended behaviour for `jf audit` / `jf scan` on V3+? Or was `--mode=update-lockfile` supposed to take effect (i.e., the flags are in the wrong order)?
2. If `--mode=skip-build` is correct, the first flag is dead code — should it be removed to make the intent explicit?

The curation branch (line 1191) already documents the clipanion last-wins behaviour, so whichever mode is intended, it's worth making that explicit here too.

gineshkumar:
**[Low] Version gate for curation duplicated in two places (`yarn.go`)**

> Note: `yarn.go` arrived via the dev merge (`05bb72f`) — flagging for the author's awareness as a follow-up.

`verifyYarnVersionSupportedForCuration` (line 235) and `configureYarnResolutionServerAndRunInstall` (line 1052) both independently reject Yarn V1 and V4 with different error messages:

```go
// line 235
if yarnVersion.Compare(yarnV2Version) > 0 || yarnVersion.Compare(yarnV4Version) <= 0 {
    return errorutils.CheckErrorf("'jf curation-audit' is not supported for Yarn V1 or Yarn V4 ...")
}

// line 1052
if yarnVersion.Compare(yarnV2Version) > 0 || yarnVersion.Compare(yarnV4Version) <= 0 {
    return errors.New("resolving Yarn dependencies from Artifactory is currently not supported for Yarn V1 and Yarn V4 ...")
}
```

When V4 gains Artifactory resolution support (or V5 ships), one gate is likely to be updated and the other missed. Consider extracting to a shared helper or constant so the supported-version policy lives in one place.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +262,253 @@ type pypiMetaData struct {
...
+	return b.String()
+}
+
+func readPoetryLockIfExists() ([]poetryLockPackage, error) {
+	exists, err := fileutils.IsFileExists(poetryLockFile, false)
+	if err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+	if !exists {
+		return nil, errorutils.CheckErrorf("process failed, %s wasn't found, can't process poetry curation command", poetryLockFile)
```
gineshkumar: **[Low] `readPoetryLockIfExists` error message is vague and gives no recovery path**

```go
return nil, errorutils.CheckErrorf("process failed, %s wasn't found, can't process poetry curation command", poetryLockFile)
```

Three issues:
1. "process failed" — doesn't say which process failed
2. Contractions (`wasn't`, `can't`) — informal, inconsistent with the rest of the file (e.g. line 276: `"a poetry repository must be configured (run 'jf poetry-config') for poetry curation"`)
3. No actionable guidance — the user doesn't know what to run to fix it

```suggestion
		return nil, errorutils.CheckErrorf("%s not found — run 'poetry lock' or 'poetry install' to generate it before running 'jf ca'", poetryLockFile)
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,23 +525,245 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+		}
+	}
+	return strings.Join(out, "\n")
+}
+
+// extractPoetrySourceNames returns the canonical list of source names from
+// viper's view of `[[tool.poetry.source]]`. Entries without a name, or with
+// duplicate names, are skipped. Returns nil when the key is missing or has
+// an unexpected shape so callers can fall back to a default.
+func extractPoetrySourceNames(v interface{}) []string {
```
gineshkumar: **[Low] Two modern Go idiom gaps (project is Go 1.26.3)**

**1. `interface{}` → `any` (`extractPoetrySourceNames`, line 715)**

`any` has been the canonical alias for `interface{}` since Go 1.18. Three occurrences here:

```suggestion
func extractPoetrySourceNames(v any) []string {
	arr, ok := v.([]any)
```

```go
// and inside the loop:
m, ok := e.(map[string]any)
```

**2. `pkg := pkg` loop-variable capture is dead code in Go 1.22+ (line 300)**

Go 1.22 changed loop variable semantics — each iteration gets its own variable, so the explicit re-declaration is no longer needed. This pattern was required pre-1.22 to safely capture the loop variable before handing it to a goroutine; in 1.22+ it's a no-op that signals the author was following an outdated pattern.

```suggestion
	for _, pkg := range packages {
		if pkg.Name == "" || pkg.Version == "" || len(pkg.Files) == 0 {
```

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,23 +525,245 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+		seen[n] = struct{}{}
+		names = append(names, n)
+	}
+	return names
+}
+
+func validateMinimumPoetryVersion(minVersion string) (int, error) {
+	out, err := executeCommand("poetry", "--version")
+	if err != nil {
+		log.Debug(fmt.Sprintf("Poetry is not installed or not on PATH: %s", err.Error()))
```
gineshkumar: **[Nit] `err.Error()` in format string is redundant**

`%v` on an error already calls `Error()` — the explicit `.Error()` call is unnecessary:

```suggestion
		log.Debug(fmt.Sprintf("Poetry is not installed or not on PATH: %v", err))
```

Same pattern applies wherever `%s` + `.Error()` appears elsewhere in new code — prefer `%v` + the error value directly.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +262,253 @@ type pypiMetaData struct {
...
+	g := new(errgroup.Group)
+	g.SetLimit(poetryDownloadUrlWorkers)
+	for _, pkg := range packages {
+		pkg := pkg
+		if pkg.Name == "" || pkg.Version == "" || len(pkg.Files) == 0 {
+			skipped++
+			continue
+		}
+		g.Go(func() error {
+			downloadUrl, lookupErr := buildPoetryDownloadUrl(rtManager, &httpClientDetails, artiUrl, repository, pkg)
```
gineshkumar: **[Medium] `&httpClientDetails` shared across goroutines — latent data race**

All 8 goroutines receive the **same pointer** to `httpClientDetails`:

```go
httpClientDetails := rtAuth.CreateHttpClientDetails()
...
g.Go(func() error {
    downloadUrl, lookupErr := buildPoetryDownloadUrl(rtManager, &httpClientDetails, ...)
```

`JfrogHttpClient.SendGet` passes this pointer to `runPreRequestInterceptors`, which hands it to each registered interceptor — and interceptors can mutate the struct (e.g. writing to the `Headers` map for token refresh or tracing headers).

`CreateServiceManager` registers no interceptors today, so the race is currently dormant. But it is one interceptor registration away from a concurrent map write across 8 goroutines — undefined behaviour and a potential crash.

`HttpClientDetails` already has a proper deep-copy method (`Clone()` copies the `Headers` map and clones `Transport`). Fix: clone once per goroutine:

```suggestion
		g.Go(func() error {
			localDetails := httpClientDetails.Clone()
			downloadUrl, lookupErr := buildPoetryDownloadUrl(rtManager, localDetails, artiUrl, repository, pkg)
```

sca/bom/buildinfo/technologies/python/python_test.go
```diff
@@ -306,3 +308,603 @@ func TestGetPipInstallArgs(t *testing.T) {
...
+	})
+}
+
+func TestExtractPoetrySourceNames(t *testing.T) {
+	t.Run("nil returns nil", func(t *testing.T) {
+		assert.Nil(t, extractPoetrySourceNames(nil))
+	})
+	t.Run("wrong type returns nil", func(t *testing.T) {
+		assert.Nil(t, extractPoetrySourceNames("not-an-array"))
+		assert.Nil(t, extractPoetrySourceNames(map[string]interface{}{"name": "x"}))
```
gineshkumar: **[Low] I1 fix incomplete — test data still uses `interface{}`**

The implementation was correctly updated to `any`, but this test (new in this PR) still uses the old style in its data literals:

```suggestion
		assert.Nil(t, extractPoetrySourceNames(map[string]any{"name": "x"}))
```

Same applies to the slice literals at lines 711–722 — `[]interface{}` / `map[string]interface{}` → `[]any` / `map[string]any`.

They compile and pass (`interface{}` and `any` are exact aliases), but they're inconsistent with the fix just applied to the function they exercise.

gineshkumar (approved):

go.mod
```diff
@@ -161,4 +161,4 @@ require (
 
 // replace github.com/jfrog/build-info-go => github.com/jfrog/build-info-go dev
 
-// replace github.com/jfrog/froggit-go => github.com/jfrog/froggit-go master
+// replace github.com/jfrog/froggit-go => github.com/jfrog/froggit-go master
```
attiasas: what is this change?

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -140,13 +181,30 @@ func getDependencies(params technologies.BuildInfoBomGeneratorParams, technology
 		technologies.LogExecutableVersion("python")
 		technologies.LogExecutableVersion(string(pythonTool))
 	}
+	if technology == techutils.Poetry {
+		log.Debug(fmt.Sprintf("Poetry: dependency tree built — %d nodes in graph, %d direct dependencies", len(dependenciesGraph), len(directDependencies)))
```
attiasas: make sure there is not double logs

https://github.com/jfrog/jfrog-cli-security/blob/main/sca/bom/buildinfo/buildinfobom.go#L306

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -220,23 +533,245 @@ func installPoetryDeps(params technologies.BuildInfoBomGeneratorParams) (rootDet
...
+	if !version.NewVersion(v).AtLeast(minVersion) {
+		return 0, errorutils.CheckErrorf("JFrog CLI poetry curation requires Poetry %s or higher. The current version is: %s", minVersion, v)
+	}
+	dot := strings.IndexByte(v, '.')
+	if dot < 0 {
+		dot = len(v)
+	}
+	major, parseErr := strconv.Atoi(v[:dot])
+	if parseErr != nil {
+		return 0, nil
```
attiasas: > 🔴 Correctness & logic

validateMinimumPoetryVersion returns (0, nil) when strconv.Atoi fails on the major version, which nilerr flags and can mis-route Poetry 2.x into 1.x source config (default = true).
Return the parse error (or a clear CheckErrorf) instead of nil — e.g. `return 0, errorutils.CheckError(parseErr)`.

attiasas (changes requested):
Nice job! checkout my comments

In addition. fix tests and static analysis issues. add integration test with actual project like other techs: https://github.com/jfrog/jfrog-cli-security/blob/main/curation_test.go

sca/bom/buildinfo/technologies/python/python.go
Phavya-jfrog: dropped the redundant line

commands/curation/curationaudit.go
```diff
@@ -547,6 +550,7 @@ func (ca *CurationAuditCommand) getBuildInfoParamsByTech() (technologies.BuildIn
 		IgnoreConfigFile:       ca.IgnoreConfigFile(),
 		InsecureTls:            ca.InsecureTls(),
 		// Install params
+		SkipAutoInstall:    ca.SkipAutoInstall(),
```
attiasas: is skiping auto install even relevant here? for curation we always want to install...

curation_test.go
```diff
@@ -134,6 +134,115 @@ func TestDockerCurationAudit(t *testing.T) {
 	assert.Equal(t, "Image is not Docker Hub official", results[0].Policy[0].Condition)
 }
 
+func TestPoetryCurationAudit(t *testing.T) {
+	integration.InitCurationTest(t)
+	
+	t.Setenv("JFROG_CLI_CURATION", "true")
```
attiasas: why needed?

curation_test.go
```diff
@@ -134,6 +134,115 @@ func TestDockerCurationAudit(t *testing.T) {
...
+
+	cleanUpJfrogHome, err := coreTests.SetJfrogHome()
+	assert.NoError(t, err)
+	defer cleanUpJfrogHome()
+
+	config.User = "admin"
+	config.Password = "password"
+	config.ServerId = "test"
+	configCmd := commonCommands.NewConfigCommand(commonCommands.AddOrEdit, config.ServerId).SetDetails(config).SetUseBasicAuthOnly(true).SetInteractive(false)
+	assert.NoError(t, configCmd.Run())
```
attiasas: use: (see other tests)

```
cleanUp := integration.UseTestHomeWithDefaultXrayConfig(t)
defer cleanUp()
```

curation_test.go
```diff
@@ -134,6 +134,115 @@ func TestDockerCurationAudit(t *testing.T) {
...
+
+	// Create the poetry resolver config (.jfrog/projects/poetry.yaml).
+	assert.NoError(t, commonCommands.CreateBuildConfigWithOptions(false, project.Poetry,
+		commonCommands.WithResolverServerId(config.ServerId),
+		commonCommands.WithResolverRepo(repo),
+	))
```
attiasas: you can just add the .yml file to the test project you added... see for example `tests/testdata/projects/package-managers/maven/maven-curation/test`....

curation_test.go
```diff
@@ -134,6 +134,115 @@ func TestDockerCurationAudit(t *testing.T) {
...
+			PkgType:           "poetry",
+			Policy: []curation.Policy{
+				{Policy: "pol1", Condition: "cond1", Explanation: "explanation", Recommendation: "recommendation"},
+				{Policy: "pol2", Condition: "cond2", Explanation: "explanation2", Recommendation: "recommendation2"},
+			},
+		},
+	}
+}
+
+func poetryCurationServer(t *testing.T, repo string, expectedRequest map[string]bool, requestToFail map[string]bool) (*httptest.Server, *config.ServerDetails) {
```
attiasas: can't we just extend/use:

```
func curationServer(t *testing.T, expectedRequest map[string]bool, requestToFail map[string]bool)
```

why do we need specific for python? can't it be adjusted to fix multile usage given params when creating the mock?

attiasas (changes requested):
Take a look at my comments. in addition make sure CI and static tests are passing (unless not related to the fix)

sca/bom/buildinfo/technologies/python/python_test.go
```diff
@@ -306,3 +308,615 @@ func TestGetPipInstallArgs(t *testing.T) {
...
+		{"some unrelated output", ""},
+	}
+	for _, tt := range tests {
+		t.Run(tt.in, func(t *testing.T) {
+			assert.Equal(t, tt.want, parsePoetryVersion(tt.in))
+		})
+	}
+}
+
+func TestValidateMinimumPoetryVersionUnparsableMajorErrors(t *testing.T) {
```
attiasas, line 896: fix failing tests for windows

sca/bom/buildinfo/technologies/python/python_test.go
```diff
@@ -306,3 +308,615 @@ func TestGetPipInstallArgs(t *testing.T) {
...
+	require.NoError(t, os.WriteFile(fakePoetry,
+		[]byte("#!/bin/sh\necho 'Poetry (version 2x.0.0)'\n"), 0755))
+	t.Setenv("PATH", fakeDir+string(os.PathListSeparator)+os.Getenv("PATH"))
+
+	major, err := validateMinimumPoetryVersion(CurationPoetryMinimumVersion)
+	require.Error(t, err, "unparsable major version must return an error, not (0, nil)")
+	assert.Equal(t, 0, major)
+}
+
+func TestInstallPoetryDepsLockCheckErrorSurfacedOnRelockFailure(t *testing.T) {
```
attiasas, line 908: fix failing tests for windows

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -204,6 +261,261 @@ type pypiMetaData struct {
...
+			if lookupErr != nil {
+				log.Debug(fmt.Sprintf("Poetry: could not resolve download URL for %s:%s: %v", pkg.Name, pkg.Version, lookupErr))
+				return nil
+			}
+			normalizedName := strings.ReplaceAll(strings.ToLower(strings.TrimSpace(pkg.Name)), "-", "_")
+			compId := PythonPackageTypeIdentifier + normalizedName + ":" + pkg.Version
+			mu.Lock()
+			urls[compId] = downloadUrl
+			mu.Unlock()
+			return nil
```
attiasas: Maybe try to extract it to a function, for readability

attiasas (approved):
LGTM

PR #769, opened 2026-05-29
Feature/xray 138688 add pnpm support for jf ca

gauriy-tech:

-----

What: Added pnpm support to jf ca (curation audit) using lockfile-based dependency resolution
Why: jf ca can't run pnpm ls (requires node_modules) — lockfile parsing avoids any tarball downloads
How: pnpm install --lockfile-only generates/refreshes the lockfile, then pnpm-lock.yaml is parsed directly. jf audit/jf scan pnpm workflow is unchanged.
Scope: Only pnpm v10.x supported. jf audit path untouched.
Test plan: new unit tests in pnpm_test.go, pnpmlock_test.go, curationaudit_test.go

Screenshot 2026-06-02 at 12 39 43 PM

Detailed Test execution plan is here https://jfrog-int.atlassian.net/browse/XRAY-144540

### DIFF
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -394,13 +394,13 @@ func getPolicyAndConditionId(policy, condition string) string {
 	return fmt.Sprintf("%s:%s", policy, condition)
 }
 
-// promotePnpmWorkspaceMember replaces "npm" with "pnpm" in the detected
-// technologies list when the current directory is a pnpm workspace member
-// (i.e., it has no pnpm indicators itself but an ancestor directory contains
-// pnpm-workspace.yaml or pnpm-lock.yaml).
+// promotePnpmWorkspaceMember replaces "npm" with "pnpm" in the detected technologies
+// list when the current directory is a pnpm workspace member — it has no pnpm marker
+// itself, but an ancestor directory contains pnpm-workspace.yaml or pnpm-lock.yaml.
+// This lets `jf ca --working-dirs=<member>` audit the member as part of its pnpm
+// workspace, consistently with the lockfile resolution which also walks up to the root.
 func promotePnpmWorkspaceMember(techs []string) []string {
-	hasPnpm := false
-	hasNpm := false
+	hasPnpm, hasNpm := false, false
 	for _, t := range techs {
 		switch t {
 		case techutils.Pnpm.String():
@@ -412,7 +412,6 @@ func promotePnpmWorkspaceMember(techs []string) []string {
 	if hasPnpm || !hasNpm {
 		return techs
 	}
-	// Walk up to find a pnpm workspace root.
 	dir, err := os.Getwd()
 	if err != nil {
 		return techs
@@ -420,25 +419,23 @@ func promotePnpmWorkspaceMember(techs []string) []string {
 	for {
 		parent := filepath.Dir(dir)
 		if parent == dir {
-			break
+			return techs
 		}
 		dir = parent
 		for _, indicator := range []string{"pnpm-workspace.yaml", "pnpm-lock.yaml"} {
 			if _, statErr := os.Stat(filepath.Join(dir, indicator)); statErr == nil {
 				log.Debug(fmt.Sprintf("Detected pnpm workspace root at %s via %s; promoting current directory from npm to pnpm.", dir, indicator))
-				result := make([]string, 0, len(techs))
+				promoted := make([]string, 0, len(techs))
 				for _, t := range techs {
 					if t == techutils.Npm.String() {
-						result = append(result, techutils.Pnpm.String())
-					} else {
-						result = append(result, t)
+						t = techutils.Pnpm.String()
 					}
+					promoted = append(promoted, t)
 				}
-				return result
+				return promoted
 			}
 		}
 	}
-	return techs
 }
 
 func (ca *CurationAuditCommand) doCurateAudit(results map[string]*CurationReport) error {
@@ -1003,12 +1000,14 @@ func (ca *CurationAuditCommand) setRepoFromNpmrc() error {
 func (ca *CurationAuditCommand) setRepoFromNpmrcForPnpm() error {
 	registryConfig, err := pnpmtech.GetNativePnpmRegistryConfig()
 	if err != nil {
-		return fmt.Errorf("pnpm: failed to read Artifactory details from .npmrc: %w\nEnsure the registry is configured in .npmrc (e.g. registry=https://<host>/artifactory/api/npm/<repo>/)", err)
+		log.Warn("Ensure the pnpm registry is configured in .npmrc (e.g. registry=https://<host>/artifactory/api/npm/<repo>/)")
+		return fmt.Errorf("pnpm: failed to read Artifactory details from .npmrc: %w", err)
 	}
 
 	var serverDetails *config.ServerDetails
 	if registryConfig.AuthToken != "" {
 		// .npmrc has an auth token that matches the registry — use it directly.
+		log.Debug("pnpm: using auth token from .npmrc")
 		serverDetails = &config.ServerDetails{
 			ArtifactoryUrl: registryConfig.ArtifactoryUrl,
 			AccessToken:    registryConfig.AuthToken,
@@ -1016,6 +1015,7 @@ func (ca *CurationAuditCommand) setRepoFromNpmrcForPnpm() error {
 	} else {
 		// No token in .npmrc — fall back to whatever 'jf c' has stored, overriding
 		// only the Artifactory URL so requests go to the correct registry.
+		log.Debug("pnpm: no token in .npmrc — using 'jf c' server credentials")
 		serverDetails, err = ca.ServerDetails()
 		if err != nil || serverDetails == nil {
 			return fmt.Errorf("pnpm: no auth token found in .npmrc and no 'jf c' server configured: %w", err)
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -39,7 +39,8 @@ import (
 	"github.com/jfrog/jfrog-cli-security/sca/bom/buildinfo"
 	"github.com/jfrog/jfrog-cli-security/sca/bom/buildinfo/technologies"
 	"github.com/jfrog/jfrog-cli-security/sca/bom/buildinfo/technologies/docker"
-	npmtech "github.com/jfrog/jfrog-cli-security/sca/bom/buildinfo/technologies/npm"
+	npmtech  "github.com/jfrog/jfrog-cli-security/sca/bom/buildinfo/technologies/npm"
+	pnpmtech "github.com/jfrog/jfrog-cli-security/sca/bom/buildinfo/technologies/pnpm"
 	"github.com/jfrog/jfrog-cli-security/sca/bom/buildinfo/technologies/python"
 	"github.com/jfrog/jfrog-cli-security/utils"
 	"github.com/jfrog/jfrog-cli-security/utils/formats"
@@ -91,6 +92,7 @@ var CurationOutputFormats = []string{string(outFormat.Table), string(outFormat.J
 var supportedTech = map[techutils.Technology]func(ca *CurationAuditCommand) (bool, error){
 	techutils.Npm:  func(ca *CurationAuditCommand) (bool, error) { return true, nil },
 	techutils.Yarn: func(ca *CurationAuditCommand) (bool, error) { return true, nil },
+	techutils.Pnpm: func(ca *CurationAuditCommand) (bool, error) { return true, nil },
 	techutils.Pip: func(ca *CurationAuditCommand) (bool, error) {
 		return ca.checkSupportByVersionOrEnv(techutils.Pip, MinArtiPassThroughSupport)
 	},
@@ -392,8 +394,55 @@ func getPolicyAndConditionId(policy, condition string) string {
 	return fmt.Sprintf("%s:%s", policy, condition)
 }
 
+// promotePnpmWorkspaceMember replaces "npm" with "pnpm" in the detected
+// technologies list when the current directory is a pnpm workspace member
+// (i.e., it has no pnpm indicators itself but an ancestor directory contains
+// pnpm-workspace.yaml or pnpm-lock.yaml).
+func promotePnpmWorkspaceMember(techs []string) []string {
+	hasPnpm := false
+	hasNpm := false
+	for _, t := range techs {
+		switch t {
+		case techutils.Pnpm.String():
+			hasPnpm = true
+		case techutils.Npm.String():
+			hasNpm = true
+		}
+	}
+	if hasPnpm || !hasNpm {
+		return techs
+	}
+	// Walk up to find a pnpm workspace root.
+	dir, err := os.Getwd()
+	if err != nil {
+		return techs
+	}
+	for {
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			break
+		}
+		dir = parent
+		for _, indicator := range []string{"pnpm-workspace.yaml", "pnpm-lock.yaml"} {
+			if _, statErr := os.Stat(filepath.Join(dir, indicator)); statErr == nil {
+				log.Debug(fmt.Sprintf("Detected pnpm workspace root at %s via %s; promoting current directory from npm to pnpm.", dir, indicator))
+				result := make([]string, 0, len(techs))
+				for _, t := range techs {
+					if t == techutils.Npm.String() {
+						result = append(result, techutils.Pnpm.String())
+					} else {
+						result = append(result, t)
+					}
+				}
+				return result
+			}
+		}
+	}
+	return techs
+}
+
 func (ca *CurationAuditCommand) doCurateAudit(results map[string]*CurationReport) error {
-	techs := techutils.DetectedTechnologiesListForCurationAudit()
+	techs := promotePnpmWorkspaceMember(techutils.DetectedTechnologiesListForCurationAudit())
 	if ca.DockerImageName() != "" {
 		log.Debug(fmt.Sprintf("Docker image name '%s' was provided, running Docker curation audit.", ca.DockerImageName()))
 		techs = []string{techutils.Docker.String()}
@@ -519,6 +568,8 @@ func (ca *CurationAuditCommand) getBuildInfoParamsByTech() (technologies.BuildIn
 		NpmLegacyPeerDeps:       ca.LegacyPeerDeps(),
 		// Yarn: always refresh yarn.lock when older than package.json (mirrors NpmOverwritePackageLock).
 		YarnOverwriteYarnLock: true,
+		// Pnpm params
+		MaxTreeDepth: ca.MaxTreeDepth(),
 		// Python params
 		PipRequirementsFile: ca.PipRequirementsFile(),
 		// Docker params
@@ -529,7 +580,7 @@ func (ca *CurationAuditCommand) getBuildInfoParamsByTech() (technologies.BuildIn
 }
 
 func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map[string]*CurationReport) error {
-	// --run-native is only supported for npm today; reject it early for all other techs.
+	// --run-native is only meaningful for npm/pnpm (.npmrc-based); reject it early for other techs.
 	if err := validateRunNativeForTech(tech, ca.RunNative()); err != nil {
 		return err
 	}
@@ -537,10 +588,14 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
 	if err != nil {
 		return errorutils.CheckErrorf("failed to get build info params for %s: %v", tech.String(), err)
 	}
-	// When --run-native is set for npm, the Artifactory details are already populated from .npmrc.
-	// Skip the npm.yaml config file lookup to avoid requiring 'jf npm-config'.
-	if ca.RunNative() && tech == techutils.Npm {
+	// When --run-native is set for npm, or for pnpm (always .npmrc-based), the Artifactory
+	// details are already populated from .npmrc. Skip the yaml config file lookup.
+	if (ca.RunNative() && tech == techutils.Npm) || tech == techutils.Pnpm {
 		params.IgnoreConfigFile = true
+	}
+	// Pnpm always resolves natively from .npmrc — --run-native is redundant and has no effect.
+	if ca.RunNative() && tech == techutils.Pnpm {
+		log.Warn("--run-native has no effect for pnpm; pnpm always resolves natively from .npmrc")
 	}
 	// For yarn with no yarn.yaml, fall back to npm.yaml — npm and yarn share the same Artifactory npm API.
 	resolverTech := resolveResolverTechForCuration(tech)
@@ -846,6 +901,12 @@ func (ca *CurationAuditCommand) SetRepo(tech techutils.Technology) error {
 		return ca.setRepoFromNpmrc()
 	}
 
+	// Pnpm always reads from .npmrc — there is no 'jf pnpm-config' command.
+	// pnpm shares the npm registry protocol, so the same .npmrc key/URL format applies.
+	if tech == techutils.Pnpm {
+		return ca.setRepoFromNpmrcForPnpm()
+	}
+
 	resolverParams, err := ca.getRepoParams(tech.GetProjectType())
 	if err != nil {
 		// npm and yarn share the same Artifactory npm API for curation, so their
@@ -873,8 +934,9 @@ func (ca *CurationAuditCommand) SetRepo(tech techutils.Technology) error {
 }
 
 // validateRunNativeForTech rejects --run-native for techs that don't implement
-// native-config semantics. Only npm is supported today; extend the allow-list
-// below when a new tech adds the matching native-config flow.
+// native-config semantics. npm uses it to read Artifactory details from .npmrc;
+// pnpm accepts it as a no-op (it always resolves from .npmrc). Extend the
+// allow-list below when a new tech adds the matching native-config flow.
 func validateRunNativeForTech(tech techutils.Technology, runNative bool) error {
 	if !runNative {
 		return nil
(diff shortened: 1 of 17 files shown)


sca/bom/buildinfo/technologies/pnpm/pnpm.go
```diff
@@ -144,12 +154,251 @@ func calculateDependencies(executablePath, workingDir string, params technologie
...
+		log.Debug("pnpm install --lockfile-only failed:\n" + string(out))
+		return wrapLockfileRegenError(out, runErr)
+	}
+
+	// Copy the generated lockfile back so parsePnpmLockFile can read it from workingDir.
+	generatedLock, readErr := fileutils.ReadFile(filepath.Join(tmpDir, "pnpm-lock.yaml"))
+	if readErr != nil {
+		return fmt.Errorf("lockfile not produced after 'pnpm install --lockfile-only': %w", readErr)
+	}
+	return os.WriteFile(filepath.Join(workingDir, "pnpm-lock.yaml"), generatedLock, 0644)
```
gineshkumar: `ensureLockfile` writes the regenerated lockfile back to the user's project directory (`workingDir`), but the `log.Debug` at the top of the function says "running in a temporary directory" — that's misleading, since the result ends up in the real project. For CI workspaces with read-only checkouts this causes a confusing permission error rather than a useful scan failure. More generally, a security scanning tool shouldn't silently modify project files.

Consider parsing the lockfile from `tmpDir` directly rather than copying it back:

```suggestion
	// parsePnpmLockFile reads from tmpDir — no write-back to the user's project needed.
	return parsePnpmLockFile(tmpDir)
```

If the write-back is intentional (to cache for future runs), at minimum add a `log.Info` so users know their lockfile was updated.

sca/bom/buildinfo/technologies/pnpm/pnpmlock.go
```diff
@@ -0,0 +1,243 @@
...
+
+// validateLockfileVersion rejects pnpm v5 lockfile format (5.x), which uses a different flat structure.
+func validateLockfileVersion(v string) error {
+	// Strip surrounding quotes if present (pnpm 9 writes lockfileVersion: '9.0').
+	v = strings.Trim(v, "'\"")
+	if v == "" {
+		return fmt.Errorf("pnpm-lock.yaml is missing lockfileVersion; run 'pnpm install --lockfile-only' to regenerate")
+	}
+	// Only reject clearly old formats (5.x). Anything >= 6.0 shares the same structure.
+	if strings.HasPrefix(v, "5.") {
```
gineshkumar: The comment says "Anything >= 6.0 shares the same structure" — this is incorrect. Lockfile versions 6.x/7.x/8.x (pnpm 7/8 era) use a `packages` block; the `snapshots` block only exists in v9.0+. The `pnpmLockFile` struct only maps `Snapshots`, so a v6/v7/v8 lockfile silently yields an empty `lf.Snapshots`. `walkSnapshot` returns `nil` for every dependency, and **all transitive dependencies are lost without any error or warning**.

A user with pnpm 10 installed but a pre-upgrade lockfile in v6 format will pass both the version check and this check, and get an incomplete dependency tree.

```suggestion
	// Only lockfileVersion '9.0' and later include the 'snapshots' block this parser uses.
	// Lockfile versions 5.x–8.x used a 'packages' block with a different structure.
	major, _, _ := strings.Cut(v, ".")
	majorNum, _ := strconv.Atoi(major)
	if majorNum < 9 {
		return fmt.Errorf("pnpm-lock.yaml lockfileVersion %q is not supported; run 'pnpm install --lockfile-only' to regenerate with a compatible lockfile format", v)
	}
```

sca/bom/buildinfo/technologies/pnpm/pnpm.go
```diff
@@ -144,12 +154,251 @@ func calculateDependencies(executablePath, workingDir string, params technologie
...
+// so both older and newer majors are rejected.
+func validateSupportedPnpmVersion(versionStr string) error {
+	// Version string may include extra lines (warnings on incompatible Node); take first token.
+	firstLine := strings.SplitN(versionStr, "\n", 2)[0]
+	parts := strings.SplitN(strings.TrimSpace(firstLine), ".", 2)
+	var major int
+	if _, scanErr := fmt.Sscanf(parts[0], "%d", &major); scanErr != nil {
+		return fmt.Errorf("could not parse pnpm version %q: %w", versionStr, scanErr)
+	}
+	if major != supportedPnpmMajorVersion {
```
gineshkumar, line 276: `major != supportedPnpmMajorVersion` rejects both older *and* newer versions. If pnpm 11 ships with the same `lockfileVersion: '9.0'` format (which is very likely — pnpm has kept that schema since v8), the lockfile parser will work correctly, but `jf ca` will refuse to run. Users who upgrade pnpm will be blocked until the codebase is updated.

The lockfile validator (`validateLockfileVersion`) already handles compatibility at the format level — the version guard should only reject versions below the minimum:

```suggestion
	if major < supportedPnpmMajorVersion {
		return fmt.Errorf("resolving pnpm dependencies from Artifactory requires pnpm %d.x or later. The current pnpm version is: %s", supportedPnpmMajorVersion, versionStr)
	}
```

If there's a specific reason pnpm 11 is expected to break the format, a `log.Warn` for `major > supportedPnpmMajorVersion` ("untested version, results may vary") would be safer than a hard reject.

sca/bom/buildinfo/technologies/pnpm/pnpmlock.go
```diff
@@ -0,0 +1,243 @@
...
+		return "", ref
+	}
+	return ref[:i], ref[i+1:]
+}
+
+// buildSnapshotKey constructs the key used to look up an entry in the snapshots map.
+// For scoped packages (@scope/pkg), rawRef is already a full key like "@scope/pkg@2.0.0(@peer@1.0)"
+// and is returned as-is. For unscoped packages, name and rawRef are joined as "name@rawRef".
+func buildSnapshotKey(name, rawRef string) string {
+	if strings.HasPrefix(rawRef, "@") {
```
gineshkumar: This branch is dead code — `rawRef` is never a scoped-package key. In both callers (`buildDepsMap` and `walkSnapshot`), `rawRef` is a version string such as `"2.0.0(@peer/dep@1.0.0)"` or `"1.0.0"`, which never starts with `@`. Scoped package names (e.g. `@scope/pkg`) go in the `name` parameter, not `rawRef`. The function comment above ("rawRef is already a full key like `@scope/pkg@2.0.0`") is also incorrect.

```suggestion
// buildSnapshotKey constructs the key used to look up an entry in the snapshots map.
// name is the bare package name (e.g. "xml" or "@scope/pkg"); rawRef is the version
// string from the importers or snapshots block (e.g. "1.0.1" or "2.0.0(@peer@1.0)").
func buildSnapshotKey(name, rawRef string) string {
	return name + "@" + rawRef
}
```

sca/bom/buildinfo/technologies/pnpm/pnpm.go
```diff
@@ -46,51 +53,56 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
...
-	return calculateDependencies(pnpmExecPath, dirForDependenciesCalculation, params)
+	return buildDependencyTreeFromPnpmLs(pnpmExecPath, currentDir, params)
 }
 
-func getPnpmExecPath() (pnpmExecPath string, err error) {
-	if pnpmExecPath, err = exec.LookPath("pnpm"); errorutils.CheckError(err) != nil {
+// buildDependencyTreeFromLockfile is the curation-audit path: parses pnpm-lock.yaml
+// directly without running pnpm ls or downloading tarballs.
+func buildDependencyTreeFromLockfile(pnpmExecPath, currentDir string, params technologies.BuildInfoBomGeneratorParams) (dependencyTrees []*xrayUtils.GraphNode, uniqueDeps []string, err error) {
+	if err = validatePnpmVersionForCuration(pnpmExecPath); err != nil {
```
gineshkumar: Nit / code quality: `pnpm --version` runs twice per curation audit. `getPnpmExecPath` (called at line 52) already invokes it for logging. `validatePnpmVersionForCuration` runs it again here for the version check. That's two subprocess spawns before any scanning starts.

Consider returning the version from `getPnpmExecPath` so it can be reused:

```go
func getPnpmExecPath() (pnpmExecPath, version string, err error) {
    // ...
    versionOut, versionErr := getPnpmCmd(pnpmExecPath, "", "--version").RunWithOutput()
    // ...
    version = strings.TrimSpace(string(versionOut))
    log.Debug("Pnpm version:", version)
    return
}

// validatePnpmVersionForCuration(version string) — no subprocess needed
```

commands/curation/curationaudit.go
```diff
@@ -824,6 +885,45 @@ func (ca *CurationAuditCommand) setRepoFromNpmrc() error {
...
+// URL parsing logic is identical. This is always called for pnpm — there is no 'jf pnpm-config'.
+//
+// Auth priority:
+//  1. Token from .npmrc — preferred, because it is scoped to the exact registry URL.
+//  2. Token from 'jf c' server config — used as fallback when .npmrc carries no token
+//     (e.g. user relies on a jf-managed credential store).
+func (ca *CurationAuditCommand) setRepoFromNpmrcForPnpm() error {
+	registryConfig, err := pnpmtech.GetNativePnpmRegistryConfig()
+	if err != nil {
+		return fmt.Errorf("pnpm: failed to read Artifactory details from .npmrc: %w\nEnsure the registry is configured in .npmrc (e.g. registry=https://<host>/artifactory/api/npm/<repo>/)", err)
```
gineshkumar: Error strings should not contain newlines — they compose in chains (e.g. `fmt.Errorf("outer: %w", err)`) and an embedded `\n` breaks the chain readability. The "Ensure the registry…" guidance belongs in a `log.Warn`, not in the error value.

```suggestion
		log.Warn("Ensure the pnpm registry is configured in .npmrc (e.g. registry=https://<host>/artifactory/api/npm/<repo>/)")
		return fmt.Errorf("pnpm: failed to read Artifactory details from .npmrc: %w", err)
```

commands/curation/curationaudit_test.go
```diff
@@ -1693,3 +1693,100 @@ func TestFetchNodesStatusConcurrentMapWrite(t *testing.T) {
...
+			sub := filepath.Join(root, "sub")
+			require.NoError(t, os.MkdirAll(sub, 0o755))
+
+			if tc.ancestorFile != "" {
+				require.NoError(t, os.WriteFile(filepath.Join(root, tc.ancestorFile), []byte{}, 0o644))
+			}
+
+			origDir, err := os.Getwd()
+			require.NoError(t, err)
+			require.NoError(t, os.Chdir(sub))
```
gineshkumar: Go 1.24 introduced `t.Chdir(dir)`, which automatically restores the working directory after the test (including on `t.Fatal` / panic). The manual save-and-restore pattern here can miss cleanup on early exits.

This project is on Go 1.25 — `t.Chdir` is available:

```suggestion
			t.Chdir(sub)
```

The `origDir` variable and the `defer` can be removed entirely.

commands/curation/curationaudit.go
```diff
@@ -824,6 +885,45 @@ func (ca *CurationAuditCommand) setRepoFromNpmrc() error {
...
+			return fmt.Errorf("pnpm: no auth token found in .npmrc and no 'jf c' server configured: %w", err)
+		}
+		serverDetails.ArtifactoryUrl = registryConfig.ArtifactoryUrl
+	}
+
+	repoConfig := (&project.RepositoryConfig{}).
+		SetTargetRepo(registryConfig.RepoName).
+		SetServerDetails(serverDetails)
+	ca.setPackageManagerConfig(repoConfig)
+	log.Info(fmt.Sprintf("pnpm: using Artifactory URL %q and repository %q from .npmrc", registryConfig.ArtifactoryUrl, registryConfig.RepoName))
```
gineshkumar, line 1030: The `log.Info` here logs which URL and repo are being used, but not *which auth source* was chosen. When a user has both a `.npmrc` token and a `jf c` server configured and the `.npmrc` token is expired or invalid, they'll get a 401 from Artifactory with no diagnostic indicating that the `.npmrc` token was preferred over the `jf c` credential.

Consider logging the auth source so failures are self-diagnosable:

```suggestion
	if registryConfig.AuthToken != "" {
		log.Debug("pnpm: using auth token from .npmrc")
	} else {
		log.Debug("pnpm: no token in .npmrc — using jf c server credentials")
	}
	log.Info(fmt.Sprintf("pnpm: using Artifactory URL %q and repository %q from .npmrc", registryConfig.ArtifactoryUrl, registryConfig.RepoName))
```

gineshkumar:
**Suggestion: add a real-world lockfile fixture**

The two existing fixtures (`pnpm-lock.yaml`, `pnpm-lock-complex.yaml`) are hand-crafted and share the same shape. Neither contains `patchedDependencies`, `overrides`, or `packageExtensionsChecksum` — top-level blocks that real pnpm 10 projects commonly have. If the parser is ever changed to use strict YAML decoding (`KnownFields(true)`), the existing tests won't catch the breakage.

**Proposed addition — two new files:**

`tests/testdata/projects/package-managers/pnpm/vite/package.json`
```json
{
  "name": "@vitejs/vite-monorepo",
  "version": "0.0.0",
  "private": true
}
```

`tests/testdata/projects/package-managers/pnpm/vite/pnpm-lock.yaml` (trimmed copy of vitejs/vite — 42 lines, structure taken verbatim):
```yaml
lockfileVersion: '9.0'

settings:
  autoInstallPeers: false
  dedupePeers: true
  excludeLinksFromLockfile: false

overrides:
  rolldown: 1.0.3

packageExtensionsChecksum: sha256-Tldxs3DhJEw/FFBonUidqhCBqApA0zxQnop3Y+BTO3U=

patchedDependencies:
  chokidar@3.6.0:
    hash: 8a4f9e2b397e6034b91a0508faae3cecb97f222313faa129d7cb0eb71e9d0e84
    path: patches/chokidar@3.6.0.patch
  dotenv-expand@13.0.0:
    hash: 49330a663821151418e003e822a82a6a61d2f0f8a6e3cab00c1c94815a112889
    path: patches/dotenv-expand@13.0.0.patch
  sirv@3.0.2:
    hash: c07c56eb72faea34341d465cde2314e89db472106ed378181e3447893af6bf95
    path: patches/sirv@3.0.2.patch

importers:
  .:
    devDependencies:
      '@types/node':
        specifier: ^24.12.4
        version: 24.12.4
      eslint:
        specifier: ^9.39.4
        version: 9.39.4(jiti@2.6.1)(ms@2.1.3)
  packages/vite:
    dependencies:
      rolldown:
        specifier: 1.0.3
        version: 1.0.3
  packages/create-vite:
    devDependencies:
      cross-spawn:
        specifier: ^7.0.6
        version: 7.0.6

snapshots:
  cross-spawn@7.0.6: {}
  rolldown@1.0.3: {}
  '@types/node@24.12.4': {}
  eslint@9.39.4(jiti@2.6.1)(ms@2.1.3): {}
```

**Test in `pnpmlock_test.go`:**
```go
func viteFixtureDir(t *testing.T) string {
    t.Helper()
    dir, err := filepath.Abs("../../../../..")
    require.NoError(t, err)
    return filepath.Join(dir, "tests", "testdata", "projects", "package-managers", "pnpm", "vite")
}

// TestParsePnpmLockFileVite is a smoke test using a trimmed copy of vitejs/vite's
// pnpm-lock.yaml. The fixture contains top-level blocks absent from the hand-crafted
// fixtures (patchedDependencies, overrides, packageExtensionsChecksum) and verifies
// the parser handles them without error. It would catch, for example, adding strict
// YAML decoding (KnownFields(true)) that would reject these unknown fields.
func TestParsePnpmLockFileVite(t *testing.T) {
    projects, err := parsePnpmLockFile(viteFixtureDir(t))
    require.NoError(t, err)
    // Fixture has 3 importers: root (.), packages/vite, packages/create-vite.
    assert.Greater(t, len(projects), 2)
}
```

Passes in 0ms. Happy to share the full diff if useful.

sca/bom/buildinfo/technologies/pnpm/pnpm.go
gauriy-tech: now all operations happening in tmp dir

sca/bom/buildinfo/technologies/pnpm/pnpm.go
gauriy-tech: RT does not support 11, only 10.x
Screenshot 2026-06-08 at 12 31 26 PM

sca/bom/buildinfo/technologies/pnpm/pnpmlock.go
gauriy-tech: Good catch on the comment — it was wrong, the real trigger is aliasing, not scoped packages. But the branch isn't dead code: for an aliased dependency ("x": "npm:<target>@<range>"), pnpm records the importers version as the alias target ref, which is itself the snapshot key. Verified with pnpm 10:

my-babel:
  specifier: npm:@babel/code-frame@^7.0.0
  version: '@babel/code-frame@7.29.7'   # rawRef, starts with @
snapshots:
  '@babel/code-frame@7.29.7':            # key == rawRef
name + "@" + rawRef would produce my-babel@@babel/code-frame@7.29.7 → snapshot miss → aliased package's transitives silently dropped.

I've kept the branch, fixed the comment, and also fixed a latent bug it surfaced: aliased unscoped targets (e.g. lodash.merge@4.6.2) don't start with @, so the old HasPrefix check missed them. The check now detects any alias target via @ in the version part (ignoring the peer-deps suffix), and I added tests for both alias shapes.

gauriy-tech:
Added — tests/testdata/projects/package-managers/pnpm/vite/{package.json,pnpm-lock.yaml} (trimmed vitejs/vite lockfile with overrides/patchedDependencies/packageExtensionsChecksum) plus TestParsePnpmLockFileVite. Guards against a future strict-decoding regression.

sca/bom/buildinfo/technologies/pnpm/pnpm.go
```diff
@@ -144,12 +158,265 @@ func calculateDependencies(executablePath, workingDir string, params technologie
...
+	return command
+}
+
+// resolveLockfileDir returns the directory whose pnpm-lock.yaml should be parsed, plus a
+// cleanup the caller must always invoke. An up-to-date lockfile returns workingDir and a
+// no-op; a missing/stale one is regenerated in a temp copy (returned for cleanup) so the
+// user's project is never modified and read-only checkouts still work.
+func resolveLockfileDir(pnpmExecPath, workingDir string) (lockfileDir string, cleanup func() error, err error) {
+	noop := func() error { return nil }
+	lockPath := filepath.Join(workingDir, "pnpm-lock.yaml")
```
attiasas, line 383: > 🔴 [Correctness]

promotePnpmWorkspaceMember walks up to find pnpm-workspace.yaml / pnpm-lock.yaml, but resolveLockfileDir only checks filepath.Join(workingDir, "pnpm-lock.yaml"). For a workspace package like packages/app/, lockExists will be false and the code will try pnpm install --lockfile-only on an isolated temp copy — likely failing or producing a wrong lockfile. Please walk up to the workspace root (same logic as promotePnpmWorkspaceMember) before resolving/regenerating the lockfile.

sca/bom/buildinfo/technologies/pnpm/pnpmlock.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	if len(lf.Importers) == 0 {
+		return nil, fmt.Errorf("pnpm-lock.yaml has no importers block; run 'pnpm install --lockfile-only' first")
+	}
+
+	var projects []pnpmLsProject
+	for importerPath, importer := range lf.Importers {
+		name, version := readPackageNameVersion(workingDir, importerPath)
+		project := pnpmLsProject{
```
attiasas: > 🔴 [Correctness]

parsePnpmLockFile iterates all lf.Importers and returns every workspace member. When jf ca runs from packages/vite/, auditTree will HEAD-check dependencies for all importers (root, packages/vite, packages/create-vite, etc.), not just the current package. After locating the root lockfile, filter to the importer matching the current working directory (e.g. "." at root, "packages/vite" for a member).

attiasas (commented):
Nice job. se my comments

sca/bom/buildinfo/technologies/pnpm/pnpmlock.go
gauriy-tech: Good catch — fixed. The behavior now mirrors npm exactly:

Root run (jf ca from the workspace root): we resolve the root lockfile and audit the whole workspace as a single tree — the root project plus each member nested as a direct child. This matches npm, where a root install resolves and audits the entire workspace from one lockfile.

Member-scoped run (jf ca --working-dirs=packages/vite): we walk up to the workspace root, parse the root lockfile, then filter to the importer matching the current directory (packages/vite) and audit only that member's dependencies. So auditTree only HEAD-checks the targeted package's deps, not every importer.

sca/bom/buildinfo/technologies/pnpm/pnpm.go
```diff
@@ -144,12 +158,265 @@ func calculateDependencies(executablePath, workingDir string, params technologie
...
+
+// validateSupportedPnpmVersion returns an error unless the installed pnpm major
+// version is exactly supportedPnpmMajorVersion. Curation supports only that major,
+// so both older and newer majors are rejected.
+func validateSupportedPnpmVersion(versionStr string) error {
+	// Version string may include extra lines (warnings on incompatible Node); take first token.
+	firstLine := strings.SplitN(versionStr, "\n", 2)[0]
+	parts := strings.SplitN(strings.TrimSpace(firstLine), ".", 2)
+	var major int
+	if _, scanErr := fmt.Sscanf(parts[0], "%d", &major); scanErr != nil {
```
gineshkumar: `fmt.Sscanf` is a C-ism for integer parsing. The idiomatic Go form is `strconv.Atoi` — it is also what this same PR uses in `pnpmlock.go:247` for the identical purpose, so this is an internal inconsistency.

```suggestion
	major, err := strconv.Atoi(parts[0])
	if err != nil {
		return fmt.Errorf("could not parse pnpm version %q: %w", versionStr, err)
	}
```

(`strconv` is already imported in this file.)

sca/bom/buildinfo/technologies/pnpm/pnpm.go
```diff
@@ -144,12 +158,265 @@ func calculateDependencies(executablePath, workingDir string, params technologie
...
+		case "--prod":
+			prodOnly = true
+		}
+	}
+	if !devOnly && !prodOnly {
+		return projects
+	}
+	for i := range projects {
+		if devOnly {
+			projects[i].Dependencies = nil
```
gineshkumar: `filterProjectsByScope` receives `projects []pnpmLsProject` as a slice value, but modifying via index (`projects[i].Dependencies = nil`) mutates the **caller's backing array** — it is not a separate copy.

This is safe today because the caller immediately overwrites `projects` and never reads the pre-filter state again. But the function name implies a non-destructive filter (consistent with how `filter` works in JS/every other language), so a future caller could reasonably reuse the original slice and get silently corrupted data.

Suggest returning a shallow copy with the fields cleared:

```suggestion
	result := make([]pnpmLsProject, len(projects))
	copy(result, projects)
	for i := range result {
		if devOnly {
			result[i].Dependencies = nil
		}
		if prodOnly {
			result[i].DevDependencies = nil
		}
	}
	return result
```

sca/bom/buildinfo/technologies/pnpm/pnpmlock.go
```diff
@@ -0,0 +1,255 @@
...
+type pnpmLockFile struct {
+	LockfileVersion string                      `yaml:"lockfileVersion"`
+	Importers       map[string]pnpmLockImporter `yaml:"importers"`
+	Snapshots       map[string]pnpmLockSnapshot `yaml:"snapshots"`
+}
+
+// pnpmLockImporter represents a workspace member (or the root project at ".").
+type pnpmLockImporter struct {
+	Dependencies    map[string]pnpmLockDep `yaml:"dependencies"`
+	DevDependencies map[string]pnpmLockDep `yaml:"devDependencies"`
```
gineshkumar: `pnpmLockImporter` maps to the `importers` block in `pnpm-lock.yaml`, but it only covers two of the three dependency sections:

```go
type pnpmLockImporter struct {
    Dependencies    map[string]pnpmLockDep `yaml:"dependencies"`
    DevDependencies map[string]pnpmLockDep `yaml:"devDependencies"`
    // optionalDependencies absent
}
```

A real-world pnpm-lock.yaml for a Next.js project with `sharp` (which Next.js officially recommends installing as optional for production image optimization) looks like:

```yaml
importers:
  .:
    optionalDependencies:
      sharp:
        specifier: ^0.33.5
        version: 0.33.5
```

With the current struct, `sharp` would be silently excluded from the curation scan. If a curation policy blocks it, the violation is never reported — and since `sharp` is optional, pnpm also silently skips it on install, so the user gets no signal at all.

The same applies to `fsevents` (macOS native file-watching), which nearly every pnpm project on macOS lists as optional.

Two questions:
1. Was the decision to skip `optionalDependencies` intentional (e.g., because optional packages may legitimately not install on this platform and reporting a block on them would be confusing)?
2. Related: `lockfileSpecifiersDrift` (the stale-lockfile check) also doesn't compare `optionalDependencies` — if a project only changes its optional deps, the lockfile is considered up-to-date and not regenerated. Is that the intended behavior?

gauriy-tech: Good catch — skipping optionalDependencies was not intentional, it's a gap. Fixed for both questions:

Parsing: pnpmLockImporter now has an OptionalDependencies field, and a new buildProdDepsMap helper folds optionalDependencies into the project's production dependencies (used in the root, each workspace member, and the single-importer/--working-dirs path). pnpm installs optional deps by default when the platform matches, and npm includes them in its tree, so treating them as prod deps is the right parity. A blocked sharp/fsevents will now surface.
Drift check: lockfileSpecifiersDrift now also compares optionalDependencies (both in the package.json struct and against the importer entry), so a project that only changes its optional deps is correctly detected as drifted and the lockfile is regenerated.
Added tests: TestParsePnpmLockFileOptionalDependencies (optional dep is curated as a prod dep) and a new drift case for an added optional dep.

gineshkumar (approved):

attiasas (changes requested):
Make sure all tests are passing. you have failing tests you added (see TestBuildDependencyTreePnpmLockfile...)
make sure your code is up to date with 'dev' branch

consider adding integration tests at: https://github.com/jfrog/jfrog-cli-security/blob/main/curation_test.go

attiasas (approved):
LGTM

PR #771, opened 2026-06-01
XRAY-144809 - CA command [Maven]: Add option to consider plugin deps

gauriy-tech:

-----

Full details about the RCA and Fix is described here https://jfrog-int.atlassian.net/browse/XRAY-144809?focusedCommentId=1115010

After Fix
Screenshot 2026-06-01 at 10 40 30 PM
Screenshot 2026-06-01 at 10 58 21 PM

Before Fix
Before_Fix


cli/docs/flags.go
```diff
@@ -350,6 +351,7 @@ var flagsMap = map[string]components.Flag{
 	CurationOutput:                components.NewStringFlag(OutputFormat, "Defines the output format of the command. Acceptable values are: table, json.", components.WithStrDefaultValue("table")),
 	SolutionPath:                  components.NewStringFlag(SolutionPath, "Path to the .NET solution file (.sln) to use when multiple solution files are present in the directory."),
 	IncludeCachedPackages:         components.NewBoolFlag(IncludeCachedPackages, "When set to true, the system will audit cached packages. This configuration is mandatory for Curation on-demand workflows, which rely on package caching."),
+	MvnIncludePluginDeps:          components.NewBoolFlag(MvnIncludePluginDeps, "[Maven] When set to true, Maven build-plugin transitive dependencies are included in the curation evaluation. Requires two additional Maven invocations (help:effective-pom, dependency:resolve-plugins) which may slow down the scan. By default only project dependencies are scanned."),
```
attiasas, line 354: make suer technical writer goes over the description text

sca/bom/buildinfo/technologies/java/mvn_test.go
```diff
@@ -439,6 +439,277 @@ func TestRemoveMavenConfig(t *testing.T) {
...
+  </build>
+</project>`
+	allowed := parseEffectivePomPluginCoordinates(xmlData)
+
+	assert.NotNil(t, allowed, "namespaced effective POM must produce a non-nil allow-list")
+	assert.Contains(t, allowed, "org.apache.maven.plugins:maven-resources-plugin")
+	assert.NotContains(t, allowed, "org.apache.maven.plugins:maven-deploy-plugin", "deploy-plugin must still be filtered when namespace is stripped")
+}
+
 func TestNewMavenDepTreeManagerPreservesAllParams(t *testing.T) {
```
gineshkumar, line 717: `TestNewMavenDepTreeManagerPreservesAllParams` explicitly asserts every field wired by `NewMavenDepTreeManager` — but `mvnIncludePluginDeps` (added by this PR at `mvn.go:73`) is missing from both the params setup and the assertions. If that wiring were accidentally removed, this test would still pass and the feature would silently default to `false`.

```suggestion
	params := &DepTreeParams{
		UseWrapper:              true,
		Server:                  server,
		DepsRepo:                "test-repo",
		IsMavenDepTreeInstalled: true,
		IsCurationCmd:           true,
		CurationCacheFolder:     "/tmp/cache",
		UseIncludedBuilds:       true,
		MvnIncludePluginDeps:    true,
	}
```

And after the existing assertions, add:

```go
assert.True(t, manager.mvnIncludePluginDeps)
```

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
 	}()
 	dependencyTree, uniqueDeps, err = getGraphFromDepTree(outputFilePaths)
+	if err != nil {
+		return
+	}
+	// Include Maven build-plugin transitive deps when requested.
+	// They are downloaded during mvn install but never appear in mvn dependency:tree,
+	// so without this step jf ca would miss curation violations that block the build.
+	// Skip if the tree is empty — no roots to attach to and no point running extra subprocesses.
+	if manager.mvnIncludePluginDeps && len(dependencyTree) > 0 {
```
gineshkumar, line 100: The injection loop (lines 98–109) contains two load-bearing invariants:
1. The deduplication guard (`if _, exists := uniqueDeps[gavID]; !exists`) — prevents double-injection
2. The module-root attachment loop — appends to every module root

Neither is exercised by any test because `resolvePluginDeps()` is a concrete method that calls Maven. Extracting the loop to a pure helper would make both invariants directly testable without mocking Maven:

```go
func injectPluginDeps(uniqueDeps map[string]*xray.DepTreeNode, dependencyTree []*xrayUtils.GraphNode, pluginDeps map[string]*xray.DepTreeNode) {
	for id, node := range pluginDeps {
		gavID := GavPackageTypeIdentifier + id
		if _, exists := uniqueDeps[gavID]; !exists {
			uniqueDeps[gavID] = node
			for _, moduleRoot := range dependencyTree {
				moduleRoot.Nodes = append(moduleRoot.Nodes, &xrayUtils.GraphNode{Id: gavID, Types: node.Types})
			}
		}
	}
}
```

The call site becomes:

```go
if manager.mvnIncludePluginDeps && len(dependencyTree) > 0 {
	injectPluginDeps(uniqueDeps, dependencyTree, manager.resolvePluginDeps())
}
```

A table test can then cover: empty `pluginDeps` (no-op), duplicate GAV (guard holds), new dep (attached to all module roots). The deduplication guard is the kind of one-liner that disappears in a cleanup without a test to catch it.

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+func (mdt *MavenDepTreeManager) resolvePluginDeps() map[string]*xray.DepTreeNode {
+	allowedPlugins := mdt.resolveInstallLifecyclePlugins()
+
+	goals := []string{"dependency:resolve-plugins", "-B"}
+	if mdt.isCurationCmd && mdt.curationCacheFolder != "" {
+		goals = append(goals, "-Dmaven.repo.local="+mdt.curationCacheFolder)
+	}
+	output, err := mdt.RunMvnCmd(goals)
+	if err != nil {
+		log.Warn("Failed to resolve Maven plugin dependencies; plugin deps will not be included in curation evaluation:", err.Error())
```
gineshkumar: This is the only new `log.Warn` in the plugin-deps code path that is missing the `[mvn-plugin-deps]` prefix. All 17 other new log sites include it consistently. An operator running `grep '\[mvn-plugin-deps\]'` to triage why zero plugin deps were included will see the count log (`[mvn-plugin-deps] 0 Maven plugin transitive deps included…`) but miss this failure message, leaving the root cause invisible without knowing to grep separately for "Failed to resolve Maven".

```suggestion
		log.Warn("[mvn-plugin-deps] Failed to resolve Maven plugin dependencies; plugin deps will not be included in curation evaluation:", err.Error())
```

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	parsed := parseMavenPluginDepsWithTrace(string(output), allowedPlugins)
+	log.Info(fmt.Sprintf("[mvn-plugin-deps] %d Maven plugin transitive deps included in curation evaluation after install-lifecycle filter", len(parsed)))
+	return parsed
+}
+
+// parseMavenPluginDepsWithTrace logs which top-level plugins were kept vs filtered as the
+// parser walks `mvn dependency:resolve-plugins` output. Use this when investigating which
+// allow-listed plugin is pulling a specific vulnerable transitive. The extra pass is cheap
+// compared to `mvn dependency:resolve-plugins` itself and only emits at debug level.
+func parseMavenPluginDepsWithTrace(output string, allowedPlugins map[string]struct{}) map[string]*xray.DepTreeNode {
```
gineshkumar: In Go, the `WithXxx` suffix is the established functional options / context-wrapping convention (`context.WithTimeout`, `context.WithCancel`, `http.NewRequestWithContext`). `parseMavenPluginDepsWithTrace` is neither — it is `parseMavenPluginDeps` with an extra debug-logging pass prepended. A contributor seeing this name will expect a tracing context or functional option argument.

The function has exactly one call site (line 140). Consider either:

- Renaming to `parseMavenPluginDepsDebugged` or `parseMavenPluginDepsVerbose`
- Or inlining the debug loop directly into `parseMavenPluginDeps` behind a `log.IsDebugLevel()` guard, removing the split entirely

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+		log.Warn("[mvn-plugin-deps] Failed to create temp file for effective POM; plugin filter disabled:", err.Error())
+		return nil
+	}
+	outputPath := outputFile.Name()
+	_ = outputFile.Close()
+	// Preserve the file on parse failure so callers can inspect why no plugins were extracted.
+	preserveFile := false
+	defer func() {
+		if preserveFile {
+			log.Info("[mvn-plugin-deps] effective POM preserved for inspection at: " + outputPath)
```
gineshkumar: This `log.Info` fires in the failure-preservation path: `preserveFile = true` is set only after `parseEffectivePomPluginCoordinates` returns nil (line 222–225), meaning something went wrong. The preceding `log.Warn` at line 224 ("effective POM parsed to empty allow-list — plugin filter disabled") is at Warn level, but this follow-on message — which tells the operator *where* to find the preserved file for inspection — is at Info.

An operator filtering on Warn+ will see the filter-disabled warning but not the file path, making the warning non-actionable.

```suggestion
			log.Warn("[mvn-plugin-deps] effective POM preserved for inspection at: " + outputPath)
```

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	if allowed == nil {
+		log.Warn(fmt.Sprintf("[mvn-plugin-deps] effective POM parsed to empty allow-list (file size %d bytes) - plugin filter disabled", info.Size()))
+		preserveFile = true
+	}
+	return allowed
+}
+
+// tailString returns at most n bytes from the end of s, useful for trimming verbose
+// mvn output to a manageable size in debug logs.
+func tailString(s string, n int) string {
```
gineshkumar, line 222: `tailString` slices by bytes (`s[len(s)-n:]`), not runes. If `len(s)-n` falls inside a multi-byte UTF-8 character, the result starts with a continuation byte and is invalid UTF-8.

This is reproducible:

```go
s := "xあy"  // 5 bytes: 78 E3 81 82 79 — "あ" is 3 bytes
got := tailString(s, 3)
// boundary = 5-3 = 2 → s[2:] = "\x81\x82y" (0x81 is a continuation byte)
// utf8.ValidString(got) == false
```

In practice this fires when Maven runs with a non-ASCII locale (e.g. `LANG=ja_JP.UTF-8`) and the error output used at line 202 (`tailString(string(mvnOutput), 2000)`) happens to be cut mid-kanji.

```suggestion
func tailString(s string, n int) string {
	runes := []rune(s)
	if len(runes) <= n {
		return s
	}
	return "..." + string(runes[len(runes)-n:])
}
```

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	log.Info(fmt.Sprintf("[mvn-plugin-deps] %d Maven plugin transitive deps included in curation evaluation after install-lifecycle filter", len(parsed)))
+	return parsed
+}
+
+// parseMavenPluginDepsWithTrace logs which top-level plugins were kept vs filtered as the
+// parser walks `mvn dependency:resolve-plugins` output. Use this when investigating which
+// allow-listed plugin is pulling a specific vulnerable transitive. The extra pass is cheap
+// compared to `mvn dependency:resolve-plugins` itself and only emits at debug level.
+func parseMavenPluginDepsWithTrace(output string, allowedPlugins map[string]struct{}) map[string]*xray.DepTreeNode {
+	for _, line := range strings.Split(output, "\n") {
```
gineshkumar: This function iterates the output once for debug logging, then `parseMavenPluginDeps` splits and iterates it again. Two passes, two `[]string` allocations on potentially large Maven output.

Move the debug logging inline into `parseMavenPluginDeps` and delete this wrapper:

```suggestion
func parseMavenPluginDepsWithTrace(output string, allowedPlugins map[string]struct{}) map[string]*xray.DepTreeNode {
	// Fold debug logging directly into parseMavenPluginDeps and remove this function.
	return parseMavenPluginDeps(output, allowedPlugins)
```

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	return allowed
+}
+
+// isPluginInInstallLifecycle returns true when the plugin's executions (or default phase)
+// fall within phases executed by `mvn install`.
+func isPluginInInstallLifecycle(coord string, executions []effectivePomExecution) bool {
+	defaultPostInstall := postInstallPluginsByDefault[coord]
+
+	// Collect phases explicitly bound by user executions (empty <phase> means "use plugin default").
+	var explicitPhases []string
```
gineshkumar: `explicitPhases` is collected into a `[]string` only to check `len == 0` and iterate. No need to materialise the slice — a single pass with a flag does the same thing with zero allocation:

```suggestion
	hasExplicit := false
	for _, ex := range executions {
		if ex.Phase == "" {
			continue
		}
		hasExplicit = true
		if !postInstallPhases[ex.Phase] {
			return true
		}
	}
	if !hasExplicit {
		return !postInstallPluginsByDefault[coord]
	}
	return false
```

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	mvnOutput, err := mdt.RunMvnCmd(goals)
+	if err != nil {
+		log.Warn("[mvn-plugin-deps] mvn help:effective-pom failed - plugin filter disabled, all plugin deps will be reported. Reason: " + err.Error())
+		if len(mvnOutput) > 0 {
+			log.Debug("[mvn-plugin-deps] mvn output (tail):\n" + tailString(string(mvnOutput), 2000))
+		}
+		return nil
+	}
+
+	info, statErr := os.Stat(outputPath)
```
gineshkumar: `os.Stat` followed immediately by `os.ReadFile` is two syscalls when one is enough. `os.ReadFile` returns an error if the file is missing, and `len(data) == 0` catches the empty case. Replace the three-way check:

```suggestion
	data, err := os.ReadFile(outputPath)
	if errors.Is(err, fs.ErrNotExist) {
		log.Warn("[mvn-plugin-deps] effective POM output file missing after mvn run - plugin filter disabled. Reason:", err.Error())
		return nil
	}
	if err != nil {
		log.Warn("[mvn-plugin-deps] failed to read effective POM output - plugin filter disabled. Reason:", err.Error())
		return nil
	}
	if len(data) == 0 {
		log.Warn("[mvn-plugin-deps] effective POM output file is empty - plugin filter disabled. The maven-help-plugin version may not honor -Doutput=")
		return nil
	}
```

Also remove the now-unused `info` / `statErr` variables and the `os.Stat` import if it becomes unused.

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+func tailString(s string, n int) string {
+	if len(s) <= n {
+		return s
+	}
+	return "..." + s[len(s)-n:]
+}
+
+// postInstallPhases are lifecycle phases NOT executed by `mvn install`.
+// A plugin whose only executions target these phases is excluded from the allow-list.
+var postInstallPhases = map[string]bool{
```
gineshkumar: `allowedPlugins` uses `map[string]struct{}` (the Go set idiom) but `postInstallPhases` and `postInstallPluginsByDefault` use `map[string]bool`. Same concept, two representations in the same file.

```suggestion
var postInstallPhases = map[string]struct{}{
	"pre-site":    {},
	"site":        {},
	"post-site":   {},
	"site-deploy": {},
	"deploy":      {},
}
```

Update the lookup in `isPluginInInstallLifecycle` from `postInstallPhases[phase]` to `_, ok := postInstallPhases[phase]; ok` (or just check presence). Same for `postInstallPluginsByDefault`.

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	}()
+
+	goals := []string{"help:effective-pom", "-B", "-Doutput=" + outputPath}
+	if mdt.isCurationCmd && mdt.curationCacheFolder != "" {
+		goals = append(goals, "-Dmaven.repo.local="+mdt.curationCacheFolder)
+	}
+	log.Debug("[mvn-plugin-deps] running 'mvn " + strings.Join(goals, " ") + "' to build the install-lifecycle plugin allow-list")
+	mvnOutput, err := mdt.RunMvnCmd(goals)
+	if err != nil {
+		log.Warn("[mvn-plugin-deps] mvn help:effective-pom failed - plugin filter disabled, all plugin deps will be reported. Reason: " + err.Error())
```
gineshkumar: String concatenation always allocates, even when debug logging is disabled. The JFrog log functions accept variadic args — use them:

```suggestion
		log.Warn("[mvn-plugin-deps] mvn help:effective-pom failed - plugin filter disabled, all plugin deps will be reported. Reason:", err.Error())
```

Same pattern applies to the other concatenated `log.Warn` calls in this function (lines 209, 219).

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +88,344 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+//
+//	[INFO]    g:a:maven-plugin:version:scope   (top-level plugin — switches the active filter)
+//	[INFO]       g:a:jar:version               (transitive dep, no classifier)
+//	[INFO]       g:a:jar:classifier:version    (transitive dep with classifier — version is last)
+func parseMavenPluginDeps(output string, allowedPlugins map[string]struct{}) map[string]*xray.DepTreeNode {
+	deps := map[string]*xray.DepTreeNode{}
+	// includeCurrent gates whether transitive deps under the most recently seen top-level
+	// plugin should be collected. nil allow-list means "include all".
+	includeCurrent := allowedPlugins == nil
+	for _, line := range strings.Split(output, "\n") {
```
gineshkumar: Go 1.26.3 has `strings.SplitSeq` (added in 1.24). It returns an iterator — no `[]string` allocated, same semantics as `strings.Split(s, sep)` in a range loop:

```suggestion
	for line := range strings.SplitSeq(output, "\n") {
```

sca/bom/buildinfo/technologies/java/mvn_test.go
```diff
@@ -439,6 +439,277 @@ func TestRemoveMavenConfig(t *testing.T) {
 	assert.FileExists(t, mavenConfigPath)
 }
 
+func TestParseMavenPluginDeps(t *testing.T) {
```
gineshkumar, line 445: All new test functions cover pure functions with no shared state. None call `t.Parallel()`. Add it to all 11 new test functions so they run concurrently:

```suggestion
func TestParseMavenPluginDeps(t *testing.T) {
	t.Parallel()
```

sca/bom/buildinfo/technologies/java/mvn_test.go
```diff
@@ -439,6 +439,277 @@ func TestRemoveMavenConfig(t *testing.T) {
...
+	deps := parseMavenPluginDeps(mvnOutput, allowed)
+
+	assert.Contains(t, deps, "org.apache.maven.plugins:maven-resources-plugin:3.4.0")
+	assert.Contains(t, deps, "org.codehaus.plexus:plexus-utils:4.0.2")
+	assert.Contains(t, deps, "org.ow2.asm:asm:9.7")
+	assert.NotContains(t, deps, "org.eclipse.sisu:org.eclipse.sisu.plexus:0.3.5", "site-plugin transitive dep must be filtered out")
+	assert.NotContains(t, deps, "org.sonatype.sisu:sisu-guice:3.2.3", "site-plugin transitive dep must be filtered out")
+}
+
+func TestParseEffectivePomPluginCoordinatesIncludesInstallLifecycle(t *testing.T) {
```
gineshkumar: Six `TestParseEffectivePom*` functions share identical structure: build XML, call `parseEffectivePomPluginCoordinates`, assert Contains/NotContains. Collapse into one table-driven test:

```go
func TestParseEffectivePomPluginCoordinates(t *testing.T) {
	t.Parallel()
	tests := []struct {
		name     string
		xmlData  string
		included []string
		excluded []string
	}{
		{
			name:     "install-lifecycle plugins included",
			xmlData:  `... your xml ...`,
			included: []string{"org.apache.maven.plugins:maven-compiler-plugin"},
			excluded: []string{"org.apache.maven.plugins:maven-deploy-plugin"},
		},
		// ... remaining cases
	}
	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			got := parseEffectivePomPluginCoordinates(tc.xmlData)
			for _, k := range tc.included {
				assert.Contains(t, got, k)
			}
			for _, k := range tc.excluded {
				assert.NotContains(t, got, k)
			}
		})
	}
}
```

~250 lines → ~100.

sca/bom/buildinfo/technologies/java/mvn.go
gauriy-tech: Good catch on the byte-vs-rune cut. Taking a slightly different fix:

func tailString(s string, n int) string {
	if len(s) <= n { return s }
	start := len(s) - n
	for start < len(s) && !utf8.RuneStart(s[start]) { start++ }
	return "..." + s[start:]
}
Reason: preserves the byte budget semantics of n (the call site uses 2000 as a ~2KB log cap) and avoids the O(n) []rune(s) allocation. The new TestTailStringValidUTF8 includes your "xあy" repro and asserts utf8.ValidString on the output.

sca/bom/buildinfo/technologies/java/mvn.go
gauriy-tech: Same fix as the earlier "WithXxx suffix" comment — wrapper deleted, debug logging folded into parseMavenPluginDeps, and the now-unused mavenPluginLineRe regex removed. Single pass over the output, single []string allocation.

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +90,334 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	if allowedPlugins != nil {
+		log.Debug(fmt.Sprintf("[mvn-plugin-deps] effective-pom install-lifecycle allow-list (%d plugins):", len(allowedPlugins)))
+		for coord := range allowedPlugins {
+			log.Debug("[mvn-plugin-deps]   allowed:", coord)
+		}
+	} else {
+		log.Debug("[mvn-plugin-deps] effective-pom allow-list unavailable - reporting every plugin dep without lifecycle filter")
+	}
+	parsed := parseMavenPluginDeps(string(output), allowedPlugins)
+	log.Info(fmt.Sprintf("[mvn-plugin-deps] %d Maven plugin transitive deps included in curation evaluation after install-lifecycle filter", len(parsed)))
```
gineshkumar: When `allowedPlugins` is nil (effective-pom resolution failed), `parseMavenPluginDeps` returns **all** plugin deps — no lifecycle filter was applied. The current message `"after install-lifecycle filter"` is inaccurate in that path.

```suggestion
	if allowedPlugins != nil {
		log.Info(fmt.Sprintf("[mvn-plugin-deps] %d plugin transitive deps included after install-lifecycle filter", len(parsed)))
	} else {
		log.Info(fmt.Sprintf("[mvn-plugin-deps] %d plugin transitive deps included (lifecycle filter unavailable — all reported)", len(parsed)))
	}
```

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -83,9 +90,334 @@ func buildMavenDependencyTree(params *DepTreeParams) (dependencyTree []*xrayUtil
...
+	for start < len(s) && !utf8.RuneStart(s[start]) {
+		start++
+	}
+	return "..." + s[start:]
+}
+
+// postInstallPhases are lifecycle phases NOT executed by `mvn install`.
+// A plugin whose only executions target these phases is excluded from the allow-list.
+// Covers the post-install Default phase (deploy), the Site lifecycle, and the Clean lifecycle.
+var postInstallPhases = map[string]struct{}{
```
gineshkumar: Only `deploy` is truly "post-install" (Default lifecycle). `pre-clean/clean/post-clean` and `pre-site/site/post-site/site-deploy` belong to separate lifecycles — they aren't post-install phases, they just don't run during `mvn install`. A future maintainer reading the name might assume it's exhaustive for the Default lifecycle only.

```suggestion
// phasesNotRunByInstall is the set of lifecycle phases that mvn install never executes.
// It covers: the single Default-lifecycle phase after install (deploy), the entire Site
// lifecycle, and the entire Clean lifecycle. A plugin whose only executions target these
// phases is excluded from the allow-list.
var phasesNotRunByInstall = map[string]struct{}{
```

Update the two lookup sites accordingly:
```go
if _, skip := phasesNotRunByInstall[ex.Phase]; !skip {
```

gineshkumar (approved):

gauriy-tech:

attiasas (commented):
Nice job!
Please add an integration test case with a project replicating that.
I would consider also adding this option to `jf audit` command. most of the logic is the same, users may want to include this in their scan as well

attiasas:
Also, why this is not implemented in `maven-dep-tree`? I may be a better solution
https://github.com/jfrog/maven-dep-tree

cli/docs/flags.go
gauriy-tech: it is approved

gauriy-tech:
This is needed by a customer, we have consider it to take as a long term solution in https://github.com/jfrog/maven-dep-tree in upcoming quarter

gauriy-tech:
Keeping it for curation scope for this PR , will connect with audit team to discuss enabling for their workflow

gauriy-tech:
Added an integration test case maven tree - one blocked plugin dependency in TestDoCurationAudit (commands/curation/curationaudit_test.go), with a new fixture under tests/testdata/projects/package-managers/maven/maven-curation-plugin-deps/.

What it tests (real mvn run against the mock curation server):

jf ca --mvn-include-plugin-deps detects a curation-blocked artifact that only enters the build via a Maven build plugin's transitive closure — the exact "0 blocked in jf ca but build is blocked" scenario.
Fixture pins maven-jar-plugin:3.4.1, whose fixed closure pulls org.ow2.asm:asm:9.8 (via plexus-archiver:4.9.2). asm is not in mvn dependency:tree, so without the flag it's invisible.
Verifies the end-to-end path: flag wiring → dependency:resolve-plugins → parse → inject into dep graph → curation HEAD probe → blocked PackageStatus (org.ow2.asm:asm:9.8, maven, {pol1, cond1}).

attiasas:
> This is needed by a customer, we have consider it to take as a long term solution in https://github.com/jfrog/maven-dep-tree in upcoming quarter

Should be documented that this is a "temp" fix. I whould rather you add it to the actual maven-dep-tree repo and we will add this only if its urgent until actually implemented. should be documented and tracked

gauriy-tech:
This is needed by a customer, we have consider it to take as a long term solution in https://github.com/jfrog/maven-dep-tree in upcoming quarter - https://jfrog-int.atlassian.net/browse/XRAY-145307

attiasas (approved):
LGTM for now. pending to see logic moved to maven-dep-tree

PR #777, opened 2026-06-16
XRAY-145623 - pip jf ca failure due to CVS

Phavya-jfrog:

-----

**Problem**
When CVS is active, Artifactory removes blocked package versions from the pip simple index. This causes pip install to fail during jf ca dependency resolution — and the command aborts with no output instead of reporting which packages are blocked.

**Solution**
Detect CVS-induced pip failures and run a metadata-API fallback to recover curation policy details and render a partial report.

### DIFF
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -96,6 +96,7 @@ const (
 )
 
 var CurationOutputFormats = []string{string(outFormat.Table), string(outFormat.Json)}
+var osGetwd = os.Getwd
 
 var supportedTech = map[techutils.Technology]func(ca *CurationAuditCommand) (bool, error){
 	techutils.Npm:  func(ca *CurationAuditCommand) (bool, error) { return true, nil },
@@ -377,6 +378,7 @@ func convertResultsToSummary(results map[string]*CurationReport) formats.Results
 			CuratedPackages: &formats.CuratedPackages{
 				PackageCount: packagesStatus.totalNumberOfPackages,
 				Blocked:      getBlocked(packagesStatus.packagesStatus),
+				IsPartial:    packagesStatus.isPartial,
 			},
 		})
 	}
@@ -1240,9 +1242,10 @@ func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, t
 		// Fallback produced nothing — surface the original error (current behaviour).
 		return cvsErr
 	}
-	workPath, wdErr := os.Getwd()
+	workPath, wdErr := osGetwd()
 	if wdErr != nil {
-		return cvsErr
+		log.Warn(fmt.Sprintf("curation-blocked resolution fallback: could not determine working directory (%v) — reporting under fallback key", wdErr))
+		workPath = "unknown-project"
 	}
 	results[filepath.Base(workPath)] = &CurationReport{
 		packagesStatus:        packagesStatus,
@@ -1395,6 +1398,12 @@ func (nc *treeAnalyzer) fetchCvsBlockedStatus(pins []python.PinnedRequirement) [
 				pin.Name, resolvedVersion, headErr))
 			continue
 		}
+		// Package is accessible — CVS cache may be stale; not currently blocked.
+		if headErr == nil && headResp != nil && headResp.StatusCode != http.StatusForbidden {
+			log.Debug(fmt.Sprintf("curation-blocked resolution fallback: HEAD probe returned %d for %s==%s — not CVS-blocked, skipping",
+				headResp.StatusCode, pin.Name, resolvedVersion))
+			continue
+		}
 
 		// ── Step 3b: 403 from HEAD → recover policy details via GET with waiver
 		var pkStatus *PackageStatus
@@ -1409,12 +1418,15 @@ func (nc *treeAnalyzer) fetchCvsBlockedStatus(pins []python.PinnedRequirement) [
 		depRelation := directRelation
 		if effectiveParent(pin) != pin.Name {
 			depRelation = indirectRelation
+		} else if pin.Version == "" && pin.VersionRange == "" {
+			// Name-only entry from ResolutionImpossible — parent attribution is
+			// unknown but these are always transitive deps by definition.
+			depRelation = indirectRelation
 		}
 		if pkStatus == nil {
-			// HEAD returned non-403 or GET probe errored — CVS stripped the
-			// version from the index but policy details aren't available via
-			// this path; record with unknown reason so the package is never
-			// silently dropped.
+			// HEAD returned 403 but GET probe errored — CVS stripped the version
+			// from the index but policy details aren't available via this path;
+			// record with unknown reason so the package is never silently dropped.
 			statuses = append(statuses, &PackageStatus{
 				PackageName:       pin.Name,
 				PackageVersion:    resolvedVersion,
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -1220,11 +1220,11 @@ func (nc *treeAnalyzer) fetchNodeStatus(node xrayUtils.GraphNode, p *sync.Map) e
 func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, tech techutils.Technology, results map[string]*CurationReport) error {
 	rtManager, serverDetails, err := ca.getRtManagerAndAuth(tech)
 	if err != nil {
-		return fmt.Errorf("CVS fallback: failed to get Artifactory manager (%w); pip error: %w", err, cvsErr)
+		return fmt.Errorf("curation-blocked resolution fallback: failed to get Artifactory manager (%w); pip error: %w", err, cvsErr)
 	}
 	rtAuth, err := serverDetails.CreateArtAuthConfig()
 	if err != nil {
-		return fmt.Errorf("CVS fallback: failed to create auth config (%w); pip error: %w", err, cvsErr)
+		return fmt.Errorf("curation-blocked resolution fallback: failed to create auth config (%w); pip error: %w", err, cvsErr)
 	}
 	analyzer := treeAnalyzer{
 		rtManager:            rtManager,
@@ -1355,7 +1355,7 @@ func (nc *treeAnalyzer) fetchCvsBlockedStatus(pins []python.PinnedRequirement) [
 			// Use the unfiltered all-versions metadata API to find the newest match.
 			allVersions, err := nc.lookupPypiAllVersions(pin.Name)
 			if err != nil {
-				log.Debug(fmt.Sprintf("CVS fallback: all-versions lookup failed for %s%s: %v",
+				log.Debug(fmt.Sprintf("curation-blocked resolution fallback: all-versions lookup failed for %s%s: %v",
 					pin.Name, pin.VersionRange, err))
 				continue
 			}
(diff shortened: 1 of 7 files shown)


Phavya-jfrog:
Screenshot 2026-06-17 at 10 12 49 AM
Screenshot 2026-06-17 at 10 13 08 AM

commands/curation/curationaudit.go
```diff
@@ -1185,6 +1213,262 @@ func (nc *treeAnalyzer) fetchNodeStatus(node xrayUtils.GraphNode, p *sync.Map) e
...
+		}
+
+		// ── Step 3a: HEAD probe — detect whether the version is download-blocked
+		headDetails := nc.httpClientDetails.Clone()
+		headResp, _, headErr := nc.rtManager.Client().SendHead(dlURL, headDetails)
+		if headErr != nil && (headResp == nil || headResp.StatusCode != http.StatusForbidden) {
+			log.Debug(fmt.Sprintf("curation-blocked resolution fallback: HEAD probe failed for %s==%s: %v",
+				pin.Name, resolvedVersion, headErr))
+			continue
+		}
```
gineshkumar, line 1400: **Bug + missing test: HEAD 200 response produces a false-positive blocked entry**

The guard at line 1393 skips a package only when a *transport-level error* occurs (`headErr != nil`). When HEAD succeeds with a non-403 status (e.g. 200), the code falls through to the `pkStatus == nil` branch and adds the package to results with `BlockingReasonUnknown` — marking an accessible, non-blocked package as "Blocked by curation (response could not be parsed)."

**Real business scenario: stale CVS simple-index cache**

CVS filters the pip simple index to hide blocked versions. When a policy changes (waiver granted, policy updated, version rotated out of scope), there can be a window where:

1. Pip reads the *cached* simple index — version is still missing → `isCvsVersionFilteredOutput` fires, `CvsBlockedError` is raised.
2. The fallback's HEAD probe goes directly to **RT → Curation → Catalog** — Curation sees the updated policy and returns **200**.
3. The code reaches `pkStatus == nil` and records the package as blocked.

The user sees a "blocked" row in the curation table for a package that is no longer blocked — a false alarm that could trigger unnecessary waivers or pipeline failures.

**Fix — add an explicit accessible-package guard:**

```suggestion
		headResp, _, headErr := nc.rtManager.Client().SendHead(dlURL, headDetails)
		if headErr != nil && (headResp == nil || headResp.StatusCode != http.StatusForbidden) {
			log.Debug(fmt.Sprintf("curation-blocked resolution fallback: HEAD probe failed for %s==%s: %v",
				pin.Name, resolvedVersion, headErr))
			continue
		}
		// Package is accessible — CVS cache may be stale; not currently blocked.
		if headErr == nil && headResp != nil && headResp.StatusCode != http.StatusForbidden {
			log.Debug(fmt.Sprintf("curation-blocked resolution fallback: HEAD probe returned %d for %s==%s — not CVS-blocked, skipping",
				headResp.StatusCode, pin.Name, resolvedVersion))
			continue
		}
```

**Missing test**

`TestFetchCvsBlockedStatusHeadErrorNoFalsePositive` already documents the intent for transport errors. A parallel test is needed for the HEAD-200 case:

```go
func TestFetchCvsBlockedStatusHeadOKNoFalsePositive(t *testing.T) {
    // Scenario: CVS simple-index cache is stale — version was missing at pip
    // resolve time but the artifact is now accessible via the direct URL.
    serverMock, _, rtManager := coreCommonTests.CreateRtRestsMockServer(t, func(w http.ResponseWriter, r *http.Request) {
        switch {
        case r.Method == http.MethodGet && strings.Contains(r.URL.Path, "/pypi/langchain-core/json"):
            w.WriteHeader(http.StatusOK)
            _, _ = w.Write([]byte(`{"releases":{"1.4.7":[]}}`))
        case r.Method == http.MethodGet && strings.Contains(r.URL.Path, "/pypi/langchain-core/1.4.7/json"):
            w.WriteHeader(http.StatusOK)
            _, _ = fmt.Fprintf(w, `{"urls":[{"packagetype":"bdist_wheel","url":"../../packages/ab/cd/langchain_core-1.4.7-py3-none-any.whl"}]}`)
        case r.Method == http.MethodHead:
            w.WriteHeader(http.StatusOK) // stale cache cleared; package is accessible
        }
    })
    defer serverMock.Close()

    analyzer := treeAnalyzer{ /* ... same setup as existing tests ... */ }
    pins := []python.PinnedRequirement{
        {Name: "langchain-core", Version: "1.4.7", ParentName: "langchain-core", ParentVersion: "1.4.7"},
    }
    statuses := analyzer.fetchCvsBlockedStatus(pins)
    require.Empty(t, statuses, "accessible package must not appear as blocked (stale CVS cache scenario)")
}
```

commands/curation/curationaudit.go
```diff
@@ -1185,6 +1213,267 @@ func (nc *treeAnalyzer) fetchNodeStatus(node xrayUtils.GraphNode, p *sync.Map) e
...
+
+		// ── Step 3b: 403 from HEAD → recover policy details via GET with waiver
+		var pkStatus *PackageStatus
+		if headResp != nil && headResp.StatusCode == http.StatusForbidden {
+			var getErr error
+			pkStatus, getErr = nc.getBlockedPackageDetails(dlURL, pin.Name, resolvedVersion)
+			if getErr != nil {
+				log.Debug(fmt.Sprintf("curation-blocked resolution fallback: GET probe failed for %s==%s: %v",
+					pin.Name, resolvedVersion, getErr))
```
gineshkumar, line 1415: **ResolutionImpossible packages mislabeled as `directRelation`**

When pip fails with ResolutionImpossible and multiple direct deps exist in requirements.txt, `parseCvsFailedPackages` cannot attribute the stripped transitive package to a specific parent, so it self-attributes:

```go
pr := PinnedRequirement{Name: name, ParentName: name}  // python_cvs_fallback.go:135
```

Here `effectiveParent(pin) == pin.Name`, so the condition below always produces `directRelation`:

```go
depRelation := directRelation
if effectiveParent(pin) != pin.Name {
    depRelation = indirectRelation
}
```

These packages appear in the ResolutionImpossible error **because** they are deep transitive deps that CVS stripped — they are never direct. The user sees them labeled "direct dependency" in the curation table, which is wrong.

**Distinguishing condition:** a ResolutionImpossible entry has `pin.Version == ""` AND `pin.VersionRange == ""`. A genuine direct exact-pin always has `pin.Version != ""`.

```suggestion
		depRelation := directRelation
		if effectiveParent(pin) != pin.Name {
			depRelation = indirectRelation
		} else if pin.Version == "" && pin.VersionRange == "" {
			// Name-only entry from ResolutionImpossible — parent attribution is
			// unknown but these are always transitive deps by definition.
			depRelation = indirectRelation
		}
```

**Missing test** — extend `TestFetchCvsBlockedStatusSetsDepRelation` with a third case:

```go
// ResolutionImpossible: name-only, self-attributed → must be indirect.
pins = []python.PinnedRequirement{
    {Name: blockedPkg, ParentName: blockedPkg}, // no Version, no VersionRange
}
statuses = analyzer.fetchCvsBlockedStatus(pins)
require.Len(t, statuses, 1)
assert.Equal(t, indirectRelation, statuses[0].DepRelation,
    "ResolutionImpossible CVS-fallback row must be indirect (parent unknown)")
```

commands/curation/curationaudit.go
```diff
@@ -1185,6 +1213,267 @@ func (nc *treeAnalyzer) fetchNodeStatus(node xrayUtils.GraphNode, p *sync.Map) e
...
+	}
+	packagesStatus := analyzer.fetchCvsBlockedStatus(cvsErr.Packages)
+	if len(packagesStatus) == 0 {
+		// Fallback produced nothing — surface the original error (current behaviour).
+		return cvsErr
+	}
+	workPath, wdErr := os.Getwd()
+	if wdErr != nil {
+		return cvsErr
+	}
```
gineshkumar: **`runCvsFallback` silently discards recovered `packagesStatus` on `os.Getwd()` failure**

`packagesStatus` is non-empty at this point — the `len == 0` guard at line 1239 already returned early. When `os.Getwd()` fails, the function returns `cvsErr`, throwing away every policy detail the metadata-API fallback just recovered. `wdErr` is also swallowed with no log entry, so the failure is invisible.

```suggestion
		workPath, wdErr := os.Getwd()
		if wdErr != nil {
			log.Warn(fmt.Sprintf("curation-blocked resolution fallback: could not determine working directory (%v) — reporting under fallback key", wdErr))
			workPath = "unknown-project"
		}
```

**Missing test** — `os.Getwd()` fails when the process's current directory has been deleted. Reproduce it by creating a temp dir, `chdir`-ing into it, removing it, then calling `runCvsFallback`:

```go
func TestRunCvsFallbackGetWdFailurePreservesResults(t *testing.T) {
    tmp := t.TempDir()
    require.NoError(t, os.Chdir(tmp))
    require.NoError(t, os.Remove(tmp)) // delete the dir we're standing in

    t.Cleanup(func() { _ = os.Chdir(os.TempDir()) }) // restore for other tests

    // build a minimal CvsBlockedError whose packages are already "resolved"
    // by a mock server — so fetchCvsBlockedStatus returns ≥1 status.
    // ... (same mock-server setup as TestFetchCvsBlockedStatusTransitive) ...

    results := map[string]*CurationReport{}
    err := ca.runCvsFallback(cvsErr, techutils.Pip, results)

    assert.NoError(t, err, "os.Getwd failure must not surface as an error")
    assert.Len(t, results, 1, "recovered packagesStatus must be stored even when Getwd fails")
}
```

commands/curation/curationaudit.go
```diff
@@ -1185,6 +1213,267 @@ func (nc *treeAnalyzer) fetchNodeStatus(node xrayUtils.GraphNode, p *sync.Map) e
...
+	results[filepath.Base(workPath)] = &CurationReport{
+		packagesStatus:        packagesStatus,
+		totalNumberOfPackages: len(packagesStatus),
```
gineshkumar, line 1252: **`PackageCount` in partial CVS fallback reports reflects only the blocked packages recovered, not the full dependency tree**

In the normal audit path `totalNumberOfPackages` is set to the full dep-tree size. Here it is set to `len(packagesStatus)` — the number of CVS-blocked packages the fallback managed to recover.

`convertResultsToSummary` surfaces this as `PackageCount`, so a project with 50 dependencies where 3 are CVS-blocked would report **"3 packages analyzed"**.  

The `isPartial: true` flag and `cvsPartialReportWarning` warn the interactive user, but neither propagates to the numeric stat.

commands/curation/curationaudit.go
gineshkumar: @Phavya-jfrog Please verify the suggestions before implementing.

Phavya-jfrog:

sca/bom/buildinfo/technologies/python/python_cvs_fallback.go
```diff
@@ -44,19 +150,106 @@ func normalizePyPIName(name string) string {
...
 	b.WriteString("Curation audit failed: one or more pinned package versions were unavailable during dependency resolution, so the corresponding curation policy violations could not be evaluated.")
 	if len(pins) > 0 {
 		b.WriteString("\n\nAffected package(s):\n")
 		for _, p := range pins {
-			fmt.Fprintf(&b, " - %s==%s\n", p.Name, p.Version)
+			if p.VersionRange != "" {
+				fmt.Fprintf(&b, " - %s%s\n", p.Name, p.VersionRange)
+			} else {
+				fmt.Fprintf(&b, " - %s==%s\n", p.Name, p.Version)
+			}
```
gineshkumar, line 166: **`formatCvsBlockedRequirementsMessage` renders `pkgname==` for ResolutionImpossible entries**

When `parseCvsFailedPackages` produces a ResolutionImpossible entry (`Version == ""` AND `VersionRange == ""`), the formatter reaches the `else` branch and emits `" - langgraph-sdk==\n"` — a trailing `==` with an empty version string.

This surfaces in `CvsBlockedError.Error()`, which is called when `runCvsFallback` fails to recover any statuses and returns `cvsErr`. The malformed line appears in logs and in the user-facing error message.

```suggestion
			if p.VersionRange != "" {
				fmt.Fprintf(&b, " - %s%s\n", p.Name, p.VersionRange)
			} else if p.Version != "" {
				fmt.Fprintf(&b, " - %s==%s\n", p.Name, p.Version)
			} else {
				fmt.Fprintf(&b, " - %s (version unknown)\n", p.Name)
			}
```

commands/curation/curationaudit.go
Phavya-jfrog: Verified and implemented

gineshkumar (approved):

attiasas (approved):
LGTM

PR #779, opened 2026-06-17
Add AI help coverage check to static analysis

ehl-jf:
Adds the AI-help-coverage check to the Static Analysis workflow. It fails CI when a visible command ships without AI-oriented help, so agents never fall back to the short human description.

The action auto-detects this repo's command entry point and checks only its commands. This repo is fully covered, so no commands are skipped.


attiasas (approved):
LGTM

PR #783, opened 2026-06-23
XRAY-145721 - Added fallback logic for Poetry

Phavya-jfrog:

-----
**Poetry fallback:** When a pinned Poetry package is blocked by CVS, instead of failing with no output, jf ca now uses the Artifactory PyPI metadata API to recover the policy violation and render a partial curation table — same as the existing pip fallback.

**Waiver suppression:** Waiver prompts are now skipped for partial reports (pip and poetry) since the full dependency tree wasn't resolved.

### DIFF
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -86,7 +86,7 @@ const (
 	MinXrayPassThroughSupport = "3.92.0"
 	MinArtiGradleGemSupport   = "7.63.5"
 
-	// cvsPartialReportWarning is shown when pip resolution failed because CVS
+	// cvsPartialReportWarning is shown when pip or poetry resolution failed because CVS
 	// stripped a required version from the simple index, but the metadata-API
 	// fallback succeeded in recovering at least one policy violation.
 	cvsPartialReportWarning = "The curation audit was unable to fully resolve the dependency tree because one or more pinned package versions " +
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -92,7 +92,7 @@ const (
 	cvsPartialReportWarning = "The curation audit was unable to fully resolve the dependency tree because one or more pinned package versions " +
 		"are blocked by the curation policy. Details of the policy violations are shown in the table below.\n" +
 		"Dependency analysis cannot proceed until these issues are addressed.\n" +
-		"Once you apply a waiver or switch to an approved version and re-run the audit, additional results will be available."
+		"Once you switch to an approved version and re-run the audit, additional results will be available."
 )
 
 var CurationOutputFormats = []string{string(outFormat.Table), string(outFormat.Json)}
@@ -361,6 +361,12 @@ func (ca *CurationAuditCommand) Run() (err error) {
 
 	for projectPath, packagesStatus := range results {
 		err = errors.Join(err, printResult(ca.OutputFormat(), projectPath, packagesStatus.packagesStatus))
+
+		// A partial report comes from the CVS fallback: the dependency tree could
+		// not be fully resolved. Never offers a waiver when the full tree wasn't built
+		if packagesStatus.isPartial {
+			continue
+		}
 
 		for _, ps := range packagesStatus.packagesStatus {
 			if ps.WaiverAllowed && !utils.IsCI() {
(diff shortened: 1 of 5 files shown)


commands/curation/curationaudit.go
```diff
@@ -92,7 +92,7 @@ const (
 	cvsPartialReportWarning = "The curation audit was unable to fully resolve the dependency tree because one or more pinned package versions " +
 		"are blocked by the curation policy. Details of the policy violations are shown in the table below.\n" +
 		"Dependency analysis cannot proceed until these issues are addressed.\n" +
-		"Once you apply a waiver or switch to an approved version and re-run the audit, additional results will be available."
+		"Once you switch to an approved version and re-run the audit, additional results will be available."
```
gineshkumar, line 95: **[Re: commands/curation/curationaudit.go:89]**

**[R-1] Low — Stale comment: says "pip" only, now applies to Poetry too**

The doc comment says _"shown when **pip** resolution failed because CVS stripped a required version"_ but `runCvsFallback` is now also called for Poetry (line 641). The constant text itself is accurate; only this comment wasn't updated.

```go
// cvsPartialReportWarning is shown when pip or poetry resolution failed because CVS
// stripped a required version from the simple index, but the metadata-API
// fallback succeeded in recovering at least one policy violation.
```

The neighbour at line 1227 already says "pip or poetry" correctly — small inconsistency to close.

commands/curation/curationaudit.go
```diff
@@ -362,6 +362,12 @@ func (ca *CurationAuditCommand) Run() (err error) {
 	for projectPath, packagesStatus := range results {
 		err = errors.Join(err, printResult(ca.OutputFormat(), projectPath, packagesStatus.packagesStatus))
 
+		// A partial report comes from the CVS fallback: the dependency tree could
+		// not be fully resolved. Never offers a waiver when the full tree wasn't built
+		if packagesStatus.isPartial {
```
gineshkumar, line 367: **[R-2] Recommendation — Missing test for waiver-skip invariant on partial reports**

The `isPartial` guard here enforces: _partial (CVS-fallback) reports never prompt for waivers_. That invariant is not tested. The nearest `isPartial` test (in `TestConvertResultsToSummary` around line 1480) targets `convertResultsToSummary`, not the `Run()` waiver loop. `requestWaiver` has exactly one call site (line 374) — if this `continue` were accidentally removed, the user would be prompted for waivers on an incomplete dependency tree.

Suggested coverage: add a table-driven case with `isPartial=true` and at least one package where `WaiverAllowed=true`, then assert `requestWaiver` is not called (either mock the method or use a spy on the command).

Counter-argument: the guard is a single `continue` — acceptable to defer.

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -636,8 +636,7 @@ func wrapPoetryCurationErr(lockErr error) error {
 		return nil
 	}
 	if isCvsVersionFilteredOutput(lockErr.Error()) {
-		pins := parseCvsFailedPackages(lockErr.Error())
-		lockErr = errors.Join(lockErr, errors.New(formatCvsBlockedRequirementsMessage(pins)))
+		return &CvsBlockedError{Packages: parseCvsFailedPackages(lockErr.Error()), Cause: lockErr}
 	}
 	if msgToUser := technologies.GetMsgToUserForCurationBlock(true, techutils.Poetry, lockErr.Error()); msgToUser != "" {
```
gineshkumar, line 641: **[R-3] Recommendation — `wrapPoetryCurationErr` msgToUser branch not covered by tests**

Two new tests cover:
- CVS-filtered → `*CvsBlockedError` (`TestWrapPoetryCurationErrReturnsCvsBlockedError`)
- Plain pass-through (`TestWrapPoetryCurationErrNonCvsPassesThrough`)

Neither covers the branch at this line — where `isCvsVersionFilteredOutput` is false but `GetMsgToUserForCurationBlock(true, Poetry, …)` returns a non-empty user-facing message (e.g., Poetry emitting an Artifactory 403-style block during the lock step that isn't caught by the CVS regex).

The PR's early-return for CVS-filtered errors makes this branch more isolated, so a future regression here would be silent.

Suggested addition: `TestWrapPoetryCurationErrMsgToUserPath` — craft an error string that `IsForbiddenOutput(poetry, …)` recognises; assert the returned error wraps both the original lockErr and the msgToUser text.

Counter-argument: this path is unchanged by the PR; acceptable to defer.

gineshkumar (approved):

commands/curation/curationaudit.go
Phavya-jfrog: Agreed, will add a dedicated test for this in a follow-up. For now deferring since the change is a single continue.

attiasas (commented):
LGTM, please add description to the PR about the changes and align to dev

attiasas (approved):

PR #786, opened 2026-06-24
XRAY-145929 - Onboard hugging face for jf curation-audit

gauriy-tech:

-----

Adds huggingfaceml as a first-class package type in jf curation-audit, enabling curation policy enforcement for Hugging Face models used in Python and notebook projects.

Auto-discovery: When HF_ENDPOINT is set and .py/.ipynb files are present, jf ca statically scans source for HF call sites and probes each discovered model against Artifactory curation. Dynamic references and local paths are surfaced as warnings after the table.

Spot-check mode: --hugging-face-model=<org/model>[:<revision>] audits only the explicitly named models, skipping auto-discovery and all other package managers. Supports comma-separated values.

Datasets are detected but not audited — Catalog does not currently score datasets.

No impact on jf audit, jf scan, or any existing package manager workflow.

Screenshot 2026-07-08 at 2 11 37 PM


commands/curation/curationaudit.go
```diff
@@ -566,6 +639,78 @@ func (ca *CurationAuditCommand) doCurateAudit(results map[string]*CurationReport
...
+		return errorutils.CheckErrorf(
+			"--docker-image and --hugging-face-model cannot be used together; run separate curation-audit commands for each",
+		)
+	}
+
+	// Detect --hugging-face-model= (explicitly empty value).
+	// An empty flag value is indistinguishable from "flag not passed" once GetStringFlagValue
+	// returns ""; scanning os.Args is the only reliable way to catch the explicit empty case.
+	if ca.HuggingFaceModel() == "" {
+		for _, arg := range os.Args {
```
gineshkumar: **Detect the empty `--hugging-face-model` value at the wiring layer, not by scanning `os.Args`.**

Scanning the global `os.Args` inside a command method is fragile and lives in the wrong layer — `validateCurationAuditFlags` has no access to the flag context, so it reaches around it. The idiomatic API is already used throughout the wiring file: `c.IsFlagSet(...)` (`cli/scancommands.go:323,393`, `cli/utils.go:73,122`).

Detect the explicit-empty case in `getCurationCommand` (`cli/scancommands.go:755`), where the `components.Context` is in scope:

```go
if c.IsFlagSet(flags.HuggingFaceModel) && c.GetStringFlagValue(flags.HuggingFaceModel) == "" {
    return nil, errorutils.CheckErrorf("--hugging-face-model value cannot be empty; ...")
}
```

Related test gap: the tests call `doCurateAudit` directly, so this `os.Args` branch never matches under test — the empty-value error path is currently uncovered. Moving the check to the context layer makes it unit-testable via a fake `Context`.

Runtime behavior is correct today; this is maintainability + testability.

sca/bom/buildinfo/technologies/huggingface/huggingface.go
```diff
@@ -0,0 +1,235 @@
...
+	// Flat graph: one root node whose children are the unique model/dataset refs.
+	root := &xrayUtils.GraphNode{Id: hfRootNodeName(workingDir), Nodes: children}
+	return []*xrayUtils.GraphNode{root}, uniqueIDs, warnings, nil
+}
+
+// hfRootNodeName returns the synthetic root node ID for the HF dependency graph.
+// Using the working directory basename (rather than a hardcoded "huggingface-project")
+// ensures that --working-dirs=a,b produces distinct keys in the results map
+// ("a:main" vs "b:main") so the second run does not overwrite the first.
+func hfRootNodeName(workingDir string) string {
```
gineshkumar: **Question: `hfRootNodeName` basename collision vs. the comment's claim.**

The comment at `curationaudit.go:371-374` states that setting `OriginPath` makes `--working-dirs` entries produce "distinct keys … so the second run does not overwrite the first." But `hfRootNodeName` returns only `filepath.Base(dir)`, and for HF the version is always forced to `main`, so the results-map key (`curationaudit.go:984`) reduces to `<basename>:main`.

Two working dirs with the same basename but different parents (e.g. `services/a/model` and `services/b/model`) would still collide — the second silently overwrites the first.

Is this scenario in scope for `--working-dirs`? If so, consider keying on a path that disambiguates same-basename dirs; if not, the comment slightly overstates the guarantee and could be softened. Not blocking.

sca/bom/buildinfo/technologies/huggingface/discovery/python.go
```diff
@@ -0,0 +1,584 @@
...
+		}
+	}
+	return false
+}
+
+func buildSnippet(argsRaw string) string {
+	if len(argsRaw) > 60 {
+		return argsRaw[:60] + "..."
+	}
+	return argsRaw
```
gineshkumar: `buildSnippet` copies raw call arguments into user-visible warnings, so values such as `token="..."` can be exposed in CI logs. Keep the file, line, and reason, but redact argument values.

---

sca/bom/buildinfo/technologies/huggingface/discovery/scanner.go
```diff
@@ -0,0 +1,193 @@
...
+			if IsExcludedWalkDir(d.Name()) {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		switch strings.ToLower(filepath.Ext(path)) {
+		case ".py":
+			disc, unres, ferr := scanPyFile(path, root)
+			if ferr != nil {
+				log.Debug(fmt.Sprintf("huggingface scanner: skipping %s: %v", path, ferr))
```
gineshkumar, line 82: Was it intentional for read and parse failures to be debug-only and treated as success? This can make the command appear to have complete HF coverage after skipping source files; would a user-visible partial-scan warning be more appropriate?

---

sca/bom/buildinfo/technologies/huggingface/huggingface.go
```diff
@@ -0,0 +1,235 @@
...
+	if name := filepath.Base(dir); name != "" && name != "." && name != "/" {
+		return name
+	}
+	return "huggingface-project"
+}
+
+// GetHuggingFaceRepositoryConfig resolves the Artifactory repository from HF_ENDPOINT
+// and verifies it exists, mirroring docker.GetDockerRepositoryConfig.
+func GetHuggingFaceRepositoryConfig() (*project.RepositoryConfig, error) {
+	serverDetails, err := config.GetDefaultServerConf()
```
gineshkumar: The command already resolves the user-selected `--server-id`, but this reloads the default server and can probe a different Artifactory instance. Pass the selected server details into this function and validate that `HF_ENDPOINT` targets the same host.

---

sca/bom/buildinfo/technologies/huggingface/discovery/python.go
```diff
@@ -0,0 +1,584 @@
...
+// filesystem path rather than a Hugging Face Hub repo id. Hub ids are either a
+// bare name or "namespace/name" (a single slash, no path prefixes); anything
+// with a leading path marker, a backslash, a drive letter, or more than one
+// slash is treated as a local path.
+// localOutputPrefixes are first-segment names that look like Hub org/model IDs (single
+// slash, no leading path marker) but are canonically local output or checkpoint directories
+// in ML training scripts. They are never used as Hub organisation names in practice.
+// This guards against "output/gpt2-finetuned"-style paths that isLocalPath would otherwise
+// classify as a valid two-segment Hub id.
+var localOutputPrefixes = []string{
```
gineshkumar, line 671: Hugging Face's repository-ID rules allow `namespace/repo_name` with an alphanumeric, `_`, `-`, or `.` namespace, so names such as `output/model` are valid syntax and this list silently excludes them from curation. Could ambiguous one-slash values use filesystem evidence or be surfaced as unresolved instead? ([Official validator documentation](https://huggingface.co/docs/huggingface_hub/main/en/package_reference/utilities#huggingface_hub.utils.validate_repo_id))

---

utils/techutils/techutils.go
```diff
@@ -59,14 +59,15 @@ const (
...
-	Docker   Technology = "docker"
-	Oci      Technology = "oci"
-	Rpm      Technology = "rpm"
-	Debian   Technology = "deb"
-	Composer Technology = "composer"
-	Alpine   Technology = "alpine"
-	Cpp      Technology = "cpp"
-	NoTech   Technology = ""
+	Docker        Technology = "docker"
+	HuggingFaceMl Technology = "huggingfaceml"
```
gineshkumar: Go initialisms stay uppercase, so this exported identifier should be `HuggingFaceML` before the new API surface spreads. _(Source: go-coding-standards.mdc)_

---

sca/bom/buildinfo/technologies/huggingface/huggingface.go
```diff
@@ -0,0 +1,235 @@
...
+	originalRef := modelRef
+	info := &ModelInfo{Revision: DefaultRevision}
+
+	// Split off the optional revision. A revision never contains '/', so only treat a
+	// trailing ":<suffix>" as a revision when the suffix has no path separator.
+	// A bare trailing colon (e.g. "org/model:") is treated as no revision given,
+	// so DefaultRevision is preserved rather than overwritten with an empty string.
+	// A leading colon (e.g. ":main") sets idx==0; idx >= 0 covers it so modelRef
+	// becomes "" and the empty-check below fires with a clear error.
+	if idx := strings.LastIndex(modelRef, ":"); idx >= 0 && !strings.Contains(modelRef[idx+1:], "/") {
```
gineshkumar: `refs/pr/3` is a valid Hugging Face revision, but the slash guard keeps it inside `RepoID`; the resulting malformed probe then becomes a debug-only 404 skip. Split the colon suffix regardless of `/` and path-escape the revision when building the URL. ([Official Hugging Face revision documentation](https://huggingface.co/docs/huggingface_hub/main/en/guides/download#from-specific-version))

---

sca/bom/buildinfo/technologies/huggingface/discovery/python.go
```diff
@@ -0,0 +1,584 @@
...
+		unresolved = append(unresolved, u...)
+	}
+	return
+}
+
+// ---- constant table -------------------------------------------------------
+
+// simpleAssignRe matches: NAME = "literal"  or  NAME = 'literal'
+// Anchored to avoid matching mid-expression assignments.
+var simpleAssignRe = regexp.MustCompile(`(?m)^[ \t]*([A-Za-z_][A-Za-z0-9_]*)\s*=\s*["']([^"']+)["'][ \t]*(?:#.*)?$`)
```
gineshkumar, line 119: A common deployment pattern keeps a checked-in fallback model but overrides it through an environment variable; this test currently audits the fallback instead of marking the runtime model unresolved:

```go
func TestParsePythonSource_EnvOverrideInvalidatesLiteral(t *testing.T) {
	t.Parallel()
	src := `
import os
from transformers import AutoModel

MODEL_ID = "sentence-transformers/all-MiniLM-L6-v2"
MODEL_ID = os.getenv("HF_MODEL_ID", MODEL_ID)
model = AutoModel.from_pretrained(MODEL_ID)
`
	discovered, unresolved := ParsePythonSource(src, "serve.py", nil)
	assert.Empty(t, discovered, "the runtime model is selected by HF_MODEL_ID")
	require.Len(t, unresolved, 1)
	assert.Equal(t, "non-literal repo_id", unresolved[0].Reason)
}
```

---

sca/bom/buildinfo/technologies/huggingface/huggingface.go
```diff
@@ -0,0 +1,235 @@
...
+// ModelInfo holds the parsed components of a --hugging-face-model flag value
+// ("<repo-id>[:<revision>]"). The Artifactory repository is derived from HF_ENDPOINT.
+type ModelInfo struct {
+	RepoID   string
+	Revision string
+}
+
+// ParseModelReference parses "<repo-id>[:<revision>]" into a ModelInfo.
+// Revision defaults to "main" when not specified.
+func ParseModelReference(modelRef string) (*ModelInfo, error) {
```
gineshkumar, line 61: Could we centralize Hugging Face repo/revision handling in a small `huggingface/reference` package instead of maintaining separate splitters and heuristics? Mirroring [`validate_repo_id`](https://huggingface.co/docs/huggingface_hub/main/en/package_reference/utilities#huggingface_hub.utils.validate_repo_id) and [`hf_hub_url`](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/file_download.py) would give flag parsing, discovery, graph IDs, and URL generation one upstream-aligned contract.

Suggested core:

```go
package reference

import (
	"fmt"
	"net/url"
	"regexp"
	"strings"
)

type RepoType string

const (
	PackagePrefix     = "huggingfaceml://"
	DatasetNodeMarker = "dataset|"
	DefaultRevision   = "main"
	maxRepoIDLength   = 96
	RepoTypeModel     RepoType = "model"
	RepoTypeDataset   RepoType = "dataset"
)

var repoPartPattern = regexp.MustCompile(
	`^[A-Za-z0-9_](?:[A-Za-z0-9._-]{0,94}[A-Za-z0-9_])?$`,
)

type Reference struct {
	RepoID   string
	Revision string
	RepoType RepoType
}

func Parse(raw string) (Reference, error) {
	raw = strings.TrimSpace(strings.TrimPrefix(raw, PackagePrefix))
	repoType := RepoTypeModel
	if strings.HasPrefix(raw, DatasetNodeMarker) {
		repoType = RepoTypeDataset
		raw = strings.TrimPrefix(raw, DatasetNodeMarker)
	}
	repoID, revision := raw, DefaultRevision
	if i := strings.LastIndex(raw, ":"); i >= 0 {
		repoID = raw[:i]
		if raw[i+1:] != "" {
			revision = raw[i+1:]
		}
	}
	if err := ValidateRepoID(repoID); err != nil {
		return Reference{}, err
	}
	return Reference{RepoID: repoID, Revision: revision, RepoType: repoType}, nil
}

// ValidateRepoID mirrors Hugging Face's documented repository-ID rules.
func ValidateRepoID(repoID string) error {
	if len(repoID) == 0 || len(repoID) > maxRepoIDLength {
		return fmt.Errorf("invalid Hugging Face repo ID %q: length must be 1-%d", repoID, maxRepoIDLength)
	}
	if strings.Count(repoID, "/") > 1 {
		return fmt.Errorf("invalid Hugging Face repo ID %q: expected repo or namespace/repo", repoID)
	}
	if strings.Contains(repoID, "--") || strings.Contains(repoID, "..") ||
		strings.HasSuffix(repoID, ".git") {
		return fmt.Errorf("invalid Hugging Face repo ID %q", repoID)
	}
	for _, part := range strings.Split(repoID, "/") {
		if !repoPartPattern.MatchString(part) {
			return fmt.Errorf("invalid Hugging Face repo ID %q", repoID)
		}
	}
	return nil
}

func EscapeRevision(revision string) string {
	return url.PathEscape(revision)
}

func (r Reference) NodeID() string {
	marker := ""
	if r.RepoType == RepoTypeDataset {
		marker = DatasetNodeMarker
	}
	return PackagePrefix + marker + r.RepoID + ":" + r.Revision
}

func (r Reference) APIPath() string {
	repoTypePath := "api/models"
	if r.RepoType == RepoTypeDataset {
		repoTypePath = "api/datasets"
	}
	parts := strings.Split(r.RepoID, "/")
	for i := range parts {
		parts[i] = url.PathEscape(parts[i])
	}
	return fmt.Sprintf("%s/%s/revision/%s",
		repoTypePath, strings.Join(parts, "/"), EscapeRevision(r.Revision))
}
```

Suggested specification-derived tests:

```go
func TestValidateRepoID_OfficialVectors(t *testing.T) {
	t.Parallel()
	tests := []struct {
		name    string
		repoID  string
		wantErr bool
	}{
		{name: "single name", repoID: "gpt2"},
		{name: "namespace", repoID: "openai-community/gpt2"},
		{name: "valid output namespace", repoID: "output/model"},
		{name: "mixed allowed characters", repoID: "Foo-BAR_foo.bar123"},
		{name: "too many path segments", repoID: "models/org/model", wantErr: true},
		{name: "leading period", repoID: ".model", wantErr: true},
		{name: "double hyphen", repoID: "org/model--v2", wantErr: true},
		{name: "double period", repoID: "org/model..v2", wantErr: true},
		{name: "git suffix", repoID: "org/model.git", wantErr: true},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
			err := ValidateRepoID(tt.repoID)
			if tt.wantErr {
				require.Error(t, err)
				return
			}
			require.NoError(t, err)
		})
	}
}

func TestParse_SpecialRevision(t *testing.T) {
	t.Parallel()
	ref, err := Parse("org/model:refs/pr/3")
	require.NoError(t, err)
	assert.Equal(t, "org/model", ref.RepoID)
	assert.Equal(t, "refs/pr/3", ref.Revision)
	assert.Equal(t, RepoTypeModel, ref.RepoType)
	assert.Equal(t, "huggingfaceml://org/model:refs/pr/3", ref.NodeID())
}

func TestParse_DatasetReferenceAndAPIPath(t *testing.T) {
	t.Parallel()
	ref, err := Parse("huggingfaceml://dataset|stanfordnlp/squad:refs/pr/3")
	require.NoError(t, err)
	assert.Equal(t, RepoTypeDataset, ref.RepoType)
	assert.Equal(t,
		"huggingfaceml://dataset|stanfordnlp/squad:refs/pr/3",
		ref.NodeID())
	assert.Equal(t,
		"api/datasets/stanfordnlp/squad/revision/refs%2Fpr%2F3",
		ref.APIPath())
}

func TestReference_APIPathEscapesRevision(t *testing.T) {
	t.Parallel()
	ref := Reference{
		RepoID: "org/model", Revision: "fix#1", RepoType: RepoTypeModel,
	}
	assert.Equal(t, "api/models/org/model/revision/fix%231", ref.APIPath())
}
```

This would replace:

- `HuggingFacePackagePrefix`, `DatasetNodeMarker`, `DefaultRevision`, `ModelInfo`, and `ParseModelReference` in `huggingface.go`;
- the duplicate split, dataset-marker handling, and URL-path construction in `getHuggingFaceNameAndVersion`;
- the separate `DefaultRevision` / `RepoType` definitions in `discovery/result.go`; and
- duplicated node-ID concatenation in both explicit and auto-discovery paths.

`ParseModelReferences` can remain as the comma-list wrapper around `reference.Parse`, while discovery can use `reference.ValidateRepoID` before filesystem-path classification. A syntactically valid ID such as `output/model` is still ambiguous with a local path, so that decision must use filesystem existence rather than namespace-name heuristics.

This refactor does not address source-regex accuracy, warning redaction, skipped-file reporting, selected-server handling, or result-key collisions; those remain separate fixes.

---

commands/curation/curationaudit.go
```diff
@@ -1378,6 +1568,11 @@ func (nc *treeAnalyzer) fetchNodeStatus(node xrayUtils.GraphNode, p *sync.Map) e
 			}
 			return nil
 		}
+		// HF 404: model not cached / unknown revision — not a curation block; skip like NuGet does.
+		if resp != nil && resp.StatusCode == http.StatusNotFound && nc.tech == techutils.HuggingFaceML {
+			log.Debug(fmt.Sprintf("Hugging Face: %s:%s not resolvable at %s (HTTP 404) — skipping", name, version, packageUrl))
```
gineshkumar: An HF 404 means the requested model/revision was not audited, but this branch only logs at Debug and continues; the node remains counted and the command later prints `Found 0 blocked packages`. Consider returning an error for explicit spot-check mode and recording a user-visible partial warning for auto-discovery.

---

sca/bom/buildinfo/technologies/huggingface/discovery/python.go
```diff
@@ -0,0 +1,686 @@
...
+		return classLocalPath
+	}
+	looksLikeLocalOutput := false
+	for _, prefix := range localOutputPrefixes {
+		if strings.HasPrefix(s, prefix) {
+			looksLikeLocalOutput = true
+			break
+		}
+	}
+	if !looksLikeLocalOutput {
```
gineshkumar: Filesystem disambiguation only runs for the hard-coded `localOutputPrefixes`, so an existing local directory such as `models/gpt2` bypasses `os.Stat` and is audited as the Hub ID `models/gpt2`. Please check root-relative filesystem existence for every relative literal before applying the prefix/Hub heuristic.

This regression test demonstrates the real local-directory case:

```go
func TestParsePythonSource_ExistingUnlistedLocalPath(t *testing.T) {
	t.Parallel()
	root := t.TempDir()
	require.NoError(t, os.MkdirAll(filepath.Join(root, "models", "gpt2"), 0o755))

	discovered, unresolved := ParsePythonSource(
		`from_pretrained("models/gpt2")`, "app.py", nil, root)
	assert.Empty(t, discovered)
	require.Len(t, unresolved, 1)
	assert.Equal(t, "local filesystem path", unresolved[0].Reason)
}
```

---

sca/bom/buildinfo/technologies/huggingface/discovery/python.go
```diff
@@ -0,0 +1,686 @@
...
+		revision = DefaultRevision
+		revDefaulted = true
+	}
+
+	// Resolve repo_type
+	repoType := cp.defaultRepoType
+	if cp.repoTypeArgName != "" {
+		if v, ok := keyword[cp.repoTypeArgName]; ok {
+			rv := resolveArgValue(v, constTable)
+			if rv.literal == "dataset" {
```
gineshkumar: When `repo_type` is present but dynamic or unsupported, this branch leaves the default `model` type and emits a model node even though runtime may resolve a dataset. Please mark non-literal/unsupported values unresolved; only explicit `model` or `dataset` values should be emitted.

This regression test shows that a runtime-selected type is currently emitted as a model:

```go
func TestParsePythonSource_DynamicRepoTypeIsUnresolved(t *testing.T) {
	t.Parallel()
	discovered, unresolved := ParsePythonSource(
		`snapshot_download(repo_id="org/data", repo_type=args.repo_type)`,
		"app.py", nil)
	assert.Empty(t, discovered)
	require.Len(t, unresolved, 1)
	assert.Equal(t, "non-literal repo_type", unresolved[0].Reason)
}
```

---

sca/bom/buildinfo/technologies/huggingface/discovery/scanner.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// FormatSkippedFilesWarning returns a user-visible warning listing source files
+// that could not be read or parsed (scan is partial). Empty when none were skipped.
+func FormatSkippedFilesWarning(skipped []SkippedFile) string {
+	if len(skipped) == 0 {
+		return ""
+	}
+	var sb strings.Builder
+	fmt.Fprintf(&sb, "%d source file(s) could not be scanned for Hugging Face references and were skipped — the scan below is PARTIAL:\n", len(skipped))
```
gineshkumar: This warning is emitted after the result tables, so “the scan below” points in the wrong direction.

```suggestion
fmt.Fprintf(&sb, "%d source file(s) could not be scanned for Hugging Face references and were skipped — the scan results are PARTIAL:\n", len(skipped))
```

---

commands/curation/curationaudit.go
gineshkumar: The new 404 warning fixes visibility, but this report still stores the full graph count; `convertResultsToSummary` therefore reports unresolved models as resolved/approved packages. Please subtract `len(analyzer.hfUnresolvedNodes)` from `totalNumberOfPackages` and carry a separate HF partial-summary flag/reason—reusing `CurationReport.isPartial` directly would also emit the CVS-specific warning. A mixed auto-discovery test (one HEAD 200, one HEAD 404) should assert one approved package and a partial summary.

---

gauriy-tech: Done
Screenshot 2026-07-20 at 12 10 43 PM

gineshkumar:
**[Re: commands/curation/curationaudit.go:934]** Reusing `getHuggingFaceNameAndVersion` to extract the project label from the tree root (a directory name, not a `repo_id:revision`) makes every HF audit's project name show a spurious `:main` suffix in both the table title and the JSON summary `Target` field — `SplitRepoIDAndRevision` finds no `:` in a bare dir name and the revision unconditionally defaults to `main`. _(Source: getUrlNameAndVersionByTech dispatch, line 2010)_

```go
	_, projectName, projectScope, projectVersion := getUrlNameAndVersionByTech(tech, rootNode, nil, "", "")
	if tech == techutils.HuggingFaceML {
		// rootNode.Id is a directory/project name, not a "repo_id:revision" model
		// reference — getHuggingFaceNameAndVersion's revision-defaulting logic only
		// applies to real model nodes.
		projectName, projectVersion = rootNode.Id, ""
	}
```

---

gineshkumar (approved):

cli/docs/scan/curation/help.go
```diff
@@ -14,7 +14,7 @@ When to use:
 
 Prerequisites:
 - A configured JFrog Platform server (jf c add) with JFrog Curation entitlement.
-- Project must use a supported package manager (npm, yarn, pip, maven, gradle, nuget, go) resolved through a curation-configured remote.
+- Project must use a supported package manager (npm, yarn, pip, maven, gradle, nuget, go) resolved through a curation-configured remote. Docker images and Hugging Face models are audited via dedicated flags (datasets are detected but not audited).
```
attiasas: Make sure text approved by the tech writer

cli/docs/flags.go
```diff
@@ -370,7 +371,8 @@ var flagsMap = map[string]components.Flag{
 	UseConfigProfile: components.NewBoolFlag(UseConfigProfile, "Set to false to override config profile for the audit.", components.WithBoolDefaultValue(true), components.SetHiddenBoolFlag()),
 
 	// Docker flags
-	DockerImageName: components.NewStringFlag(DockerImageName, "Specifies the Docker image name to audit. Uses the same format as the Docker CLI, including Artifactory-hosted images."),
+	DockerImageName:  components.NewStringFlag(DockerImageName, "Specifies the Docker image name to audit. Uses the same format as the Docker CLI, including Artifactory-hosted images."),
+	HuggingFaceModel: components.NewStringFlag(HuggingFaceModel, "Hugging Face models to audit, as '<repo-id>[:<revision>]' (revision defaults to 'main'). Multiple entries are comma-separated (e.g. 'mcpotato/42-eicar-street:main,bert-base-uncased'). Datasets are not audited (curation does not currently cover datasets). Requires HF_ENDPOINT set to the Artifactory Hugging Face repository URL."),
```
attiasas, line 377: Similar, make sure text approved

sca/bom/buildinfo/technologies/huggingface/discovery/python.go
```diff
@@ -0,0 +1,695 @@
...
+		}
+		if end < 0 {
+			continue // unmatched paren — skip
+		}
+		spans = append(spans, callSpan{name: name, argsRaw: line[argsStart:end]})
+	}
+	return spans
+}
+
+func matchLogicalLine(line string, startLine int, filename string, constTable map[string]string, root string) (discovered []DiscoveredModel, unresolved []UnresolvedSite) {
```
attiasas, line 330: Only skips lines starting with #; findCallSpans then runs on the raw line. So:
```
result = compute() # from_pretrained("org/fake-model")
```

Is audited as a real model (empirically reproduced). For a curation gate this produces false-positive blocks on commented-out code.

commands/curation/curationaudit.go
```diff
@@ -2057,6 +2320,43 @@ func getDockerNameAndVersion(id, artiUrl, repo string) (downloadUrls []string, n
...
 }
 
+// getHuggingFaceNameAndVersion extracts the model id and revision from a node id of the
+// form "huggingfaceml://<repo_id>:<revision>" and builds the model-info probe URL.
+//
+// The probe targets the model metadata endpoint, which the curation service blocks
+// (HEAD returns 403) for a malicious revision — independent of any specific file:
+//
+//	{artiUrl}/api/huggingfaceml/{repo}/api/models/{repo_id}/revision/{revision}
+func getHuggingFaceNameAndVersion(id, artiUrl, repo string) (downloadUrls []string, name, version string) {
```
attiasas, line 2341: repo_id (name) is not URL-escaped while revision is.
ParseModelReference does zero charset validation, so ?, #, spaces, or .. in a model ref produce malformed/injected/traversal URLs in the HEAD probe.

commands/curation/curationaudit.go
```diff
@@ -556,6 +664,13 @@ func (ca *CurationAuditCommand) doCurateAudit(results map[string]*CurationReport
 		}
 
 		if err := ca.auditTree(techutils.Technology(tech), results); err != nil {
+			if techutils.Technology(tech) == techutils.HuggingFaceML && ca.HuggingFaceModel() == "" {
+				// HF auto-discovery is additive — config/connectivity failures must not abort other techs.
+				log.Warn(fmt.Sprintf("Hugging Face curation audit skipped: %v", err))
```
attiasas, line 673: When HF is the only tech and HF_ENDPOINT/repo is misconfigured, the user gets a silent success

Is this ok?

sca/bom/buildinfo/technologies/huggingface/discovery/python.go
```diff
@@ -0,0 +1,695 @@
...
+		}
+		return argValue{isDynamic: true}
+	}
+
+	return argValue{isDynamic: true}
+}
+
+// unquotePythonString extracts the value from a Python string literal token,
+// including triple-quoted """...""" and ”'...”' forms.
+func unquotePythonString(s string) (string, bool) {
```
attiasas, line 619: accepts multi-literal tokens
"org/" + "model" and implicit string concatenation yield garbage repo IDs (e.g. org/" + "model) that get probed

commands/curation/curationaudit.go
```diff
@@ -708,11 +889,31 @@ func (ca *CurationAuditCommand) getBuildInfoParamsByTech() (technologies.BuildIn
 		PipRequirementsFile: ca.PipRequirementsFile(),
 		// Docker params
 		DockerImageName: ca.DockerImageName(),
+		// Hugging Face params
+		HuggingFaceModel: ca.HuggingFaceModel(),
+		WorkingDirectory: ca.OriginPath,
```
attiasas:
```suggestion
		HFWorkingDirectory: ca.OriginPath,
```

similar to `HFProjectName`

attiasas (commented):
Nice work, check out my comments!

commands/curation/curationaudit.go
gauriy-tech: Fixed the same
Screenshot 2026-07-28 at 11 01 01 AM

attiasas (approved):

PR #798, opened 2026-07-07
XRAY-140550 - Onboarding uv for jf curation-audit

Phavya-jfrog:

Onboards uv as a supported package manager for jf curation-audit, plus support for auditing standalone PEP 723 inline scripts. Registry details are read from pyproject.toml's [[tool.uv.index]] or ~/.config/uv/uv.toml (however Artifactory's "Set Me Up" configures it) — no jf uv-config step needed. Resolution always goes through Artifactory's curation pass-through endpoint: uv.lock is regenerated in a temp dir unless it's both up-to-date and verified as already resolved from Artifactory, in which case it's reused as-is.

New --script <path> flag audits a single PEP 723 inline script directly instead of scanning the working directory. When a uv project has an un-audited sibling script, jf ca surfaces a one-line hint to use --script.

Changes
- Add techutils.Uv to curation-audit's supported technologies, with a minimum-version gate (uv >= 0.6.17).
- Add GetNativeUvRegistryConfig/setRepoFromUvToml for registry detection, and generateUvLockInTempDir/generateUvLock to resolve uv.lock through the curation pass-through endpoint.
- Add parseUvLock, buildUvDepTree, and buildUvDownloadUrlsMap to build the dependency tree and download URLs from the lock file, including multi-version marker forks and workspace members.
- Add buildDependencyTreeForScript/generateUvScriptLockForCuration and the --script flag for auditing PEP 723 inline scripts standalone.
- Add WrapUvCurationErr/classifyUvCurationLockError so CVS-stripped versions route through the existing metadata-API fallback table instead of a bare error.
- Fix Poetry's CVS-fallback to recover range-constrained transitive dependencies via poetryDependsOnParentRegex (previously only exact pins were caught).
- Fix promotePipToUv so a pip-exclusive file (requirements.txt, Pipfile, etc.) always wins over a stray uv.lock, and vice versa.

-----


commands/curation/curationaudit.go
```diff
@@ -1258,6 +1390,35 @@ func (ca *CurationAuditCommand) setRepoFromYarnrcForYarnV4(yarnExecPath, working
...
+
+	base, sdErr := ca.ServerDetails()
+	if sdErr != nil {
+		return fmt.Errorf("uv: failed to read 'jf c' server configuration: %w", sdErr)
+	}
+	if base == nil {
+		return errorutils.CheckErrorf("uv: no 'jf c' server configured")
+	}
+	copied := *base
+	copied.ArtifactoryUrl = registryConfig.ArtifactoryUrl
```
gineshkumar, line 1882: A project-local `pyproject.toml` can replace the configured Artifactory origin while retaining its credentials. Running `jf ca` can therefore send credentials for the configured JPD to another host.

---

sca/bom/buildinfo/technologies/uv/uv.go
```diff
@@ -0,0 +1,1063 @@
...
+	}
+
+	artiIndexUrl, artifactoryUrl, repoName := "", "", ""
+	if params.ServerDetails != nil && params.DependenciesRepository != "" {
+		artifactoryUrl = params.ServerDetails.GetArtifactoryUrl()
+		repoName = params.DependenciesRepository
+		var buildErr error
+		artiIndexUrl, buildErr = buildUvCurationIndexUrl(params.ServerDetails, params.DependenciesRepository)
+		if buildErr != nil {
+			log.Warn(fmt.Sprintf("uv: failed to build curation index URL: %v — will fall back to public PyPI", buildErr))
```
gineshkumar: If curation-index URL construction fails, this continues with an empty `UV_DEFAULT_INDEX`, so `uv lock` resolves through ambient/public configuration and the audit can bypass curation. Consider returning `buildErr` here so the audit fails closed.

This regression test currently fails with `An error is expected but got nil`:

```go
func TestScriptAuditFailsClosedWhenCurationIndexCannotBeBuilt(t *testing.T) {
	fakeBinDir := t.TempDir()
	writeFakeUvExecutableForScriptLock(t, fakeBinDir, "https://ambient.example", "repo")

	scriptPath := filepath.Join(t.TempDir(), "demo.py")
	require.NoError(t, os.WriteFile(scriptPath, []byte(pep723ScriptContent), 0o644))

	params := technologies.BuildInfoBomGeneratorParams{
		ScriptPath:             scriptPath,
		ServerDetails:          &config.ServerDetails{ArtifactoryUrl: ":// malformed"},
		DependenciesRepository: "repo",
	}

	_, _, _, err := buildDependencyTreeForScript(params)
	require.Error(t, err, "curation audit must not run uv against an ambient index")
}
```

---

sca/bom/buildinfo/technologies/uv/uv.go
```diff
@@ -0,0 +1,1063 @@
...
+func maskPassword(s, rawIndexUrl string) string {
+	u, err := url.Parse(rawIndexUrl)
+	if err != nil {
+		return s
+	}
+	pw, hasPw := u.User.Password()
+	if !hasPw || pw == "" {
+		return s
+	}
+	return strings.ReplaceAll(s, pw, "***")
```
gineshkumar: `u.User.Password()` returns the decoded password, while uv can echo its percent-encoded userinfo, leaving credentials such as `p%40ss%2Fword` in logs and returned errors. Consider redacting the encoded userinfo as well as the decoded secret.

This regression test currently fails because the encoded password remains in `masked`:

```go
func TestMaskPasswordMasksPercentEncodedPassword(t *testing.T) {
	u := &url.URL{
		Scheme: "https",
		Host:   "host.example",
		Path:   "/artifactory/api/pypi/repo/simple",
		User:   url.UserPassword("user", "p@ss/word"),
	}
	encodedSecret := "p%40ss%2Fword"
	out := "uv failed while fetching " + u.String()

	masked := maskPassword(out, u.String())

	assert.NotContains(t, masked, encodedSecret)
}
```

---

gineshkumar:
Could you add a description covering the uv project and PEP 723 script flows, registry/auth behavior, validation performed, and the Poetry CVS fallback changes? The current body is only the unchecked template, so reviewers cannot verify the intended scope or design trade-offs.

---

commands/curation/curationaudit.go
```diff
@@ -1258,6 +1391,56 @@ func (ca *CurationAuditCommand) setRepoFromYarnrcForYarnV4(yarnExecPath, working
...
+func sameArtifactoryHost(configuredUrl, discoveredUrl string) bool {
+	configured, err := url.Parse(configuredUrl)
+	if err != nil || configured.Host == "" {
+		return false
+	}
+	discovered, err := url.Parse(discoveredUrl)
+	if err != nil || discovered.Host == "" {
+		return false
+	}
+	return strings.EqualFold(configured.Host, discovered.Host)
```
gineshkumar: The host-only match still allows an `https → http` downgrade to the *same* host — `copied.ArtifactoryUrl` (line 1419) adopts the `pyproject.toml`/`uv.toml` scheme, and `buildUvCurationIndexUrl` then embeds `user:password` into that URL as `UV_DEFAULT_INDEX`, so a project-local `pyproject.toml` declaring `http://<same-host>/...` can force the configured server's credentials to be sent in cleartext. +1 to the origin-replacement concern above — matching the scheme closes the residual (and update the "ignoring scheme" doc comment on line 1430 accordingly).

```suggestion
	return strings.EqualFold(configured.Host, discovered.Host) &&
		strings.EqualFold(configured.Scheme, discovered.Scheme)
```

---

sca/bom/buildinfo/technologies/uv/uv.go
```diff
@@ -0,0 +1,1079 @@
...
+			continue
+		}
+		compId := python.PythonPackageTypeIdentifier + python.NormalizePypiName(pkg.Name) + ":" + pkg.Version
+		urls[compId] = direct
+		log.Debug(fmt.Sprintf("uv: %s:%s -> %s", pkg.Name, pkg.Version, direct))
+	}
+
+	expected := len(packages) - skipped
+	resolved := len(urls)
+	if resolved < expected {
```
gineshkumar: This warning is effectively unreachable: `expected` is `len(packages) - skipped`, and every non-skipped package is added to `urls`, so `resolved == expected` except in the degenerate duplicate-component-id case — the "N package(s) will not be HEAD-checked" signal will essentially never fire. Consider dropping it, or deriving `expected` from packages that had a candidate download URL but failed the Artifactory-prefix check (a separate counter), so the warning reflects real coverage gaps.

---

sca/bom/buildinfo/technologies/uv/uv.go
```diff
@@ -0,0 +1,1079 @@
...
+			strings.Contains(line, "path ="), strings.Contains(line, "git ="), strings.Contains(line, "directory ="):
+			continue // not resolved from a package index — curation doesn't apply
+		case strings.Contains(line, "registry"):
+			registry, found := extractQuotedField(line, "registry")
+			if !found || !allowed[registry] {
+				return false
+			}
+		default:
+			return false
+		}
```
gineshkumar, line 525: These source-kind checks match `virtual`/`editable` as bare substrings of the whole line, so a `registry` URL that happens to contain either word is misrouted into the "curation doesn't apply" branch and skips the allow-set check — it's never verified against the repo the user configured. Low severity: it only triggers when a source URL literally contains `virtual`/`editable` (a repo named that way, or a foreign registry in the lock), and a re-resolution backstops correctness. Anchoring each case to its `<key> =` (and checking `registry` first, since uv always emits `source = { registry = "..." }`) keeps the intended classification without the misrouting.

```suggestion
		switch {
		case strings.Contains(line, "registry ="):
			registry, found := extractQuotedField(line, "registry")
			if !found || !allowed[registry] {
				return false
			}
		case strings.Contains(line, "virtual ="), strings.Contains(line, "editable ="),
			strings.Contains(line, "path ="), strings.Contains(line, "git ="), strings.Contains(line, "directory ="):
			continue // not resolved from a package index — curation doesn't apply
		default:
			return false
		}
```

---

Phavya-jfrog: Fixed — applied exactly as suggested: 
registry = is now checked first, and every case is anchored to "=" instead of matching bare substrings, so a registry URL containing "virtual"/"editable" is no longer misrouted past the allow-set check.

gineshkumar (approved):

_typos.toml
```diff
@@ -2,6 +2,7 @@
 testng = "testng"
 afe = "afe"
 Bouned = "Bouned"
+certifi = "certifi"
```
attiasas: ?

cli/docs/scan/curation/help.go
```diff
@@ -24,12 +24,14 @@ Common patterns:
...
   $ jf curation-audit --docker-image=my-image:tag
   $ HF_ENDPOINT=https://my.jfrog.io/artifactory/api/huggingfaceml/my-hf-repo jf curation-audit --hugging-face-model=org/model:main
+  $ jf ca --script=script.py
 
 Gotchas:
 - The user/token must be entitled for Curation; otherwise the command exits with an entitlement notice.
 - Requires the project's package manager binary on PATH (npm, mvn, etc.).
 - Run from the project root or pass --working-dirs.
 - For Maven multi-module: --use-wrapper if mvnw is used.
+- --script (uv only) audits one PEP 723 inline-script .py file directly; --script and --working-dirs cannot be used together, run separate commands for each.
```
attiasas, line 34: make sure all text changes are approved

cli/docs/flags.go
```diff
@@ -356,6 +357,7 @@ var flagsMap = map[string]components.Flag{
 	MvnIncludePluginDeps:          components.NewBoolFlag(MvnIncludePluginDeps, "[Maven] When set to true, Maven build-plugin transitive dependencies are resolved and included in the curation evaluation. By default only project dependencies are scanned."),
 	LegacyPeerDeps:                components.NewBoolFlag(LegacyPeerDeps, "[npm] Pass --legacy-peer-deps to npm install to bypass peer-dependency version conflicts."),
 	RunNative:                     components.NewBoolFlag(RunNative, "[npm] Use the native npm client for dependency resolution. Reads Artifactory URL and repository from the project's .npmrc registry — no 'jf npm-config' required. Respects .npmrc and Volta configuration."),
+	Script:                        components.NewStringFlag(Script, "[uv] Path to a PEP 723 inline-script .py file to audit standalone, instead of scanning the working directory for a pyproject.toml/uv.lock project."),
```
attiasas, line 360: make sure new flag is approved and reported to a tech writer

sca/bom/buildinfo/technologies/uv/uv.go
```diff
@@ -0,0 +1,1084 @@
...
+			}
+			continue
+		}
+	}
+	flush()
+	return urls
+}
+
+// parseArtifactoryPypiUrl splits an Artifactory PyPI URL into its base URL and repo name.
+func parseArtifactoryPypiUrl(rawUrl string) (artiUrl, repoName string, err error) {
```
attiasas, line 201: Index URLs are sliced with rawUrl[:idx], so https://user:token@host/artifactory/api/pypi/repo/simple becomes ArtifactoryUrl https://user:token@host/artifactory. That value is logged at Info (:82, :101)

Artifactory “Set Me Up” PyPI/uv URLs commonly embed tokens. This logs secrets and makes buildUvDownloadUrlsMap’s strings.HasPrefix(direct, artiBase) fail when lock download URLs lack userinfo — packages are skipped as “non-Artifactory” and never HEAD-checked. Pipenv already does this correctly via url.Parse and fmt.Sprintf("%s://%s%s", scheme, host, path) in python.go:1266-1283

Parse with url.Parse; set ArtifactoryUrl to scheme+host+path only (strip userinfo). Keep using jf c credentials (as setRepoFromUvToml already intends). Add a test with an embedded-credential index URL.

commands/curation/curationaudit.go
```diff
@@ -625,32 +635,115 @@ func promoteYarnWorkspaceMember(techs []string) []string {
...
-	if ca.HuggingFaceModel() != "" {
+	if uvSignal == "" {
+		return techs
+	}
+	log.Info(uvSignal + " — treating project as uv.")
+	techs = removeTech(techs, techutils.Pip.String())
+	if !slices.Contains(techs, techutils.Uv.String()) {
+		techs = append(techs, techutils.Uv.String())
+	}
+	return techs
```
attiasas, line 725: Presence of user-level uv config is treated as a project-type signal. A pip project detected only via bare pyproject.toml (no requirements.txt / setup.py) is flipped to uv if the developer has a personal Set Me Up uv.toml.

Yarn explicitly guards against this (curationaudit.go:881-885: global config must not promote without a project pin). Global uv config is for registry auth, not “this directory is a uv project.” Mis-promotion will fail or audit the wrong resolver.

Drop ~/.config/uv/uv.toml from promotePipToUv signals. Keep it only in GetNativeUvRegistryConfig. Require project-local signals (uv.lock, [tool.uv], [[tool.uv.index]]).

sca/bom/buildinfo/technologies/uv/uv.go
```diff
@@ -0,0 +1,1084 @@
...
+	return strings.Join(out, "\n")
+}
+
+// generateUvLock runs `uv lock` in workDir for curation-audit. When scriptName is set,
+// it runs `uv lock --script <scriptName>` instead, locking that single PEP 723 inline
+// script rather than the project in workDir.
+// When artiIndexUrl is set, UV_DEFAULT_INDEX is overridden so uv resolves through Artifactory;
+// CVS failures (version stripped from the simple index) become CvsBlockedError.
+// When artiIndexUrl is empty (index URL could not be built), uv lock runs with its own config.
+func generateUvLock(workDir, artiIndexUrl, scriptName string) error {
```
attiasas, line 648: Project [[tool.uv.index]] is neutralized and UV_DEFAULT_INDEX is set, but ~/.config/uv/uv.toml is still loaded. Extra indexes there can still participate in resolution.

A secondary public index can satisfy CVS-stripped packages that Artifactory removed, bypassing curation during lock generation — the same class of risk called out for project indexes (XRAY-146949).

Point UV_CONFIG_FILE at an empty/minimal temp config, or otherwise prevent user-config discovery while still allowing project pyproject.toml dependency reading. Re-verify lock sources afterward (already partially done).

sca/bom/buildinfo/technologies/uv/uv.go
```diff
@@ -0,0 +1,1084 @@
...
+	out, err := exec.Command("uv", "--version").CombinedOutput()
+	if err != nil {
+		return errorutils.CheckErrorf("JFrog CLI uv curation requires uv %s or higher to be installed.", CurationUvMinimumVersion)
+	}
+	raw := strings.TrimSpace(string(out))
+	versionStr := parseUvVersionFromOutput(raw)
+	if versionStr == "" {
+		log.Debug(fmt.Sprintf("uv: could not parse version from %q — skipping minimum version check", raw))
+		return nil
+	}
```
attiasas, line 788: Output-format changes or unexpected binaries silently skip the >= 0.6.17 requirement the feature depends on.
(Fail closed (same as missing uv), or accept only known formats.)

cli/scancommands.go
```diff
@@ -728,6 +728,13 @@ func ShouldRunCurationAfterFailure(c *components.Context, tech techutils.Technol
 }
 
 func getCurationCommand(c *components.Context) (*curation.CurationAuditCommand, error) {
+	scriptPath := c.GetStringFlagValue(flags.Script)
+	if scriptPath != "" && c.GetStringFlagValue(flags.WorkingDirs) != "" {
+		return nil, errorutils.CheckErrorf("--script and --working-dirs cannot be used together; run separate curation-audit commands for each")
+	}
+	if scriptPath != "" && c.GetStringFlagValue(flags.DockerImageName) != "" {
+		return nil, errorutils.CheckErrorf("--script and --docker-image cannot be used together; run separate curation-audit commands for each")
+	}
```
attiasas, line 737: Script conflicts with working-dirs/docker are rejected, but HF is silently ignored when --script is set.

sca/bom/buildinfo/technologies/uv/uv_test.go
```diff
@@ -0,0 +1,1723 @@
...
+[[tool.uv.index]]
+name = "two"
+url = "https://host/artifactory/api/pypi/uv-test-repo/simple"
+`
+		got := stripNonExplicitPyprojectIndexes(content)
+		assert.NotContains(t, got, "tool.uv.index")
+	})
+}
+
+func TestParseArtifactoryPypiUrl(t *testing.T) {
```
attiasas, line 461: No unit test for credential-bearing index URLs

attiasas (changes requested):
Nice work, 
Check out my comments, make sure you are up to date with `dev` and all tests are passings (changes are covered)

cli/docs/scan/curation/help.go
Phavya-jfrog: Got approval

cli/docs/flags.go
Phavya-jfrog: Got approval

commands/curation/curationaudit.go
Phavya-jfrog: Kept this intentionally: uv has no jf uv-config — it's a fully native approach, unlike pip/npm/yarn where JFrog CLI manages its own config file. Since there's no project-level uv config marker beyond uv.lock/[tool.uv], we treat ~/.config/uv/uv.toml as signaling uv intent when nothing else says otherwise. If any pip-exclusive file is present — requirements.txt, setup.py, setup.cfg, Pipfile, poetry.lock — the project is still classified as pip regardless.

attiasas (approved):
LGTM

PR #805, opened 2026-07-21
XRAY-156157 - Obtain configuration by workspace

VidrevichMichael:

-----

Add command line flag "workspace". It allows to define more than one configuration for repo.

Depends on:
* https://github.com/jfrog/jfrog-client-go/pull/1358

### DIFF
--- a/cli/docs/flags.go
+++ b/cli/docs/flags.go
@@ -180,6 +180,7 @@ const (
 	GitThreads      = gitPrefix + Threads
 
 	UseConfigProfile = "use-config-profile"
+	Workspace        = "workspace, ws"
 )
 
 // Mapping between security commands (key) and their flags (key).
@@ -220,7 +221,7 @@ var commandFlags = map[string][]string{
 		// Violations params
 		scanProjectKey, Watches, Snippet, ScanVuln, Fail,
 		// Scan params
-		Threads, ExclusionsAudit, WorkingDirs,
+		Threads, ExclusionsAudit, WorkingDirs, Workspace,
 		auditSca, auditIac, auditSast, auditSecrets, auditWithoutCA, SecretValidation, Sbom, UseConfigProfile,
 		// Output params
 		Licenses, OutputFormat, ExtendedTable, OutputDir, UploadRtRepoPath,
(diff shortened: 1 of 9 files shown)


cli/docs/flags.go
```diff
@@ -180,6 +180,7 @@ const (
 	GitThreads      = gitPrefix + Threads
 
 	UseConfigProfile = "use-config-profile"
+	Workspace        = "workspace, ws"
```
Jordanh1996: fable:
```
🔴 Critical — the flag value is silently dropped

cli/docs/flags.go declares:

Workspace = "workspace, ws"

The "name, alias" syntax is a urfave/cli v1 convention, but the jfrog-cli-core components framework doesn't support it. The conversion layer registers the flag with urfave (which splits it into --workspace and --ws, so parsing and help output look fine), but then reads the value back with the full compound string: fillFlagMaps calls baseContext.String(f.Name) → flagSet.Lookup("workspace, ws") → nil, and IsSet("workspace, ws") is also false (conversionlayer.go:452-459, getValueForStringFlag at conversionlayer.go:480). I confirmed empirically with urfave/cli v1.22.17:

go run . ga --workspace my-ws
→ String("workspace, ws")="" IsSet=false | String("workspace")="my-ws"

So c.GetStringFlagValue(flags.Workspace) in cli/gitcommands.go:62 always returns "" and the entire feature is a no-op from the command line. Notably, no other flag in flags.go uses alias syntax — there's no precedent because it doesn't work.

Fix: Workspace = "workspace". If a short alias is genuinely wanted, it needs framework support first; drop ws for now.

Why tests didn't catch it: the new test calls GetConfigProfileByUrl(...) directly, below the flag layer. A test exercising the components Context → SetWorkspaceName path would have caught this — worth adding alongside the fix.
```

cli/docs/flags.go
```diff
@@ -368,6 +369,7 @@ var flagsMap = map[string]components.Flag{
 	Port:         components.NewStringFlag(Port, "Specifies the port to run the SAST server on.", components.SetMandatory()),
 
 	UseConfigProfile: components.NewBoolFlag(UseConfigProfile, "Set to false to override config profile for the audit.", components.WithBoolDefaultValue(true), components.SetHiddenBoolFlag()),
+	Workspace:        components.NewStringFlag(Workspace, "Workspace name used together with the repository URL to fetch the applicable config profile from the platform."),
```
attiasas, line 372: Make sure the tech writer goes over and approve text

commands/git/audit/gitauditparams.go
```diff
@@ -20,6 +20,7 @@ type GitAuditParams struct {
 	serverDetails    *config.ServerDetails
 	useConfigProfile bool
 	configProfile    *services.ConfigProfile
+	workspaceName    string
```
attiasas: Please move this attribute to be in `results.ResultContext`
It will be easy to use/display later or append it to the results when converting to other formats

attiasas (commented):
LGTM! nice work take a look at my comments
1. Please add description to the PR.
2. I think we may want to append `workspace` meta information to the outputs in different formats (I think our PM wanted to add property to the CDX format about it, please validate)

attiasas (approved):
LGTM

PR #807, opened 2026-07-23
Update checkout to V7 and add input, remove unused workflow

attiasas:
Fixing:

```
Run actions/checkout@v5
Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

Tests/CI will not run if not marked by needed label (only maintainers can add)...


eranturgeman (approved):
LGTM, just 2 things:
1) what is the github.sha addition. why is it needed?
2) make sure that the 'persist-credentials' is not breaking anything. it basically prevents GITHUB_TOKEN to be kept in the .git config file, which means that if a later step needs to perform some git operations and so far it counted on finding the token in the .git config file- it will fail

Issue #404, opened 2025-04-03
jfrog audit --mvn not work with disabled repository id "central" on pom.xml


phuc98ute:
Hi,

Environment like below:
- Run on github runner.
- jfrog-cli latest version v2.74.1
- jfrog-cli-sercurity version v1.16.1

I am working on the maven project with config to disable maven central with config on pom.xml like below:
`

    <repositories>
        <repository>
            <id>central</id>
            <url>http://repo1.maven.org/maven2</url>
            <releases>
                <enabled>false</enabled>
            </releases>
            <snapshots>
                <enabled>false</enabled>
            </snapshots>
        </repository>
    </repositories>
    <pluginRepositories>
        <pluginRepository>
            <id>central</id>
            <url>http://repo1.maven.org/maven2</url>
            <releases>
                <enabled>false</enabled>
            </releases>
            <snapshots>
                <enabled>false</enabled>
            </snapshots>
        </pluginRepository>
    </pluginRepositories>
`

the command jf audit --mvn always return error:
`
Error:  Plugin org.apache.maven.plugins:maven-install-plugin:3.1.1 or one of its dependencies could not be resolved:
Error:  	Could not find artifact org.apache.maven.plugins:maven-install-plugin:jar:3.1.1
Error:  -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginResolutionException

12:07:47 [Info] Trace ID for JFrog Platform logs: 693742722d1f63ff
Error: 7 [Error] target '/runner/_work/insshd-azure-library-mvn-parent/insshd-azure-library-mvn-parent [maven]' errors:
failed to build dependency tree: failed while building 'maven' dependency tree: failed running command 'mvn org.apache.maven.plugins:maven-install-plugin:3.1.1:install-file -Dfile=/tmp/jfrog.cli.temp.-1743509261-2764527690/maven-dep-tree.jar -B -s /tmp/jfrog.cli.temp.-1743509261-2764527690/settings.xml': exit status 1
`

After deep dive into logic of frog-cli-security, I believe that the logic cause the issue like detailed below:
[...]

PR #808, opened 2026-07-23
XRAY-135041 slnx support added for dotnet

Phavya-jfrog:

-----

`jf curation-audit` silently skipped .NET projects with only a `.slnx` file, since technology detection never recognized it.

- Added `.slnx` to `Nuget`/`Dotnet` indicators and package descriptors.
- Fixed `--solution-path` not being passed into `solution.Load()` (was hardcoded to `""`), which made `solution.Load()` auto-discover and merge projects from every solution file instead of honoring the explicit flag.
- Updated `--solution-path` flag description to mention `.slnx`.

**Upstream PR:**
build-info-go PR -- https://github.com/jfrog/build-info-go/pull/404

Screenshot 2026-07-24 at 11 34 51 AM


utils/techutils/techutils.go
```diff
@@ -246,8 +246,8 @@ var technologiesData = map[Technology]TechData{
 	},
 	Nuget: {
 		formal:             "NuGet",
-		indicators:         []string{".sln", ".csproj"},
-		packageDescriptors: []string{".sln", ".csproj"},
+		indicators:         []string{".sln", ".slnx", ".csproj"},
```
gineshkumar, line 251: build-info-go cannot parse `.slnx`, so adding it to `indicators` makes a nested solution resolve **zero** dependencies — and nested is what `dotnet new sln` produces by default on SDK 9.0.200+.

Verified on `root/My.slnx` + `root/Lib/Lib.csproj`:

| command | before | after |
|---|---|---|
| `jf ca` | .NET not detected, skipped | detected → `found no dependencies for the audited project using 'nuget' as the package manager` (`curationaudit.go:760`) |
| `jf audit` (recursive, default) | working dir `Lib` → `trees=1 deps=4` | working dir `.` → `trees=0 deps=0`, and `buildinfobom.go:169` drops it at Debug level, so in a multi-tech repo the .NET packages vanish silently |

Upstream cause: auto-discovery filters `filepath.Ext(filePath) == ".sln"` (`solution/solution.go:338`); `parseSlnFile`'s regex `Project\("(.*\..*proj)` cannot match `.slnx` XML (`solution.go:372`); the `loadSingleProjectFromDir` fallback lists `*proj` **non-recursively** (`solution.go:268` → `utils/fileutils.go:189`), so it sees 0 projects at the solution root. The only `.slnx` layout that works today is the solution file and `.csproj` in the *same* directory, with `--solution-path` — which no fixture in this repo uses.

Could this wait on `.slnx` parsing in build-info-go? Detection alone turns a skipped project into a failing one.

---

sca/bom/buildinfo/technologies/nuget/nuget_test.go
```diff
@@ -286,3 +278,46 @@ func TestRunDotnetRestoreWithRealSolutionFile(t *testing.T) {
 		assert.NotContains(t, err.Error(), "this folder contains more than one project or solution file")
 	}
 }
+
+// TestBuildDependencyTreeWithSlnxSolution proves a '.slnx'-only project (no '.sln' fallback)
```
gineshkumar, line 282: This test fails on the PR head — both assertions get empty values:

```
[Warn] expecting 1 'proj' file but fuond 0 files in path: /var/.../jfrog.cli.temp.-...
    Error: Should NOT be empty, but was []
--- FAIL: TestBuildDependencyTreeWithSlnxSolution (3.35s)
```

`dotnet restore` succeeds; the failure is that `solution.Load` returns no projects for the `slnx-single` fixture, because the `.csproj` lives in `ClassLibrary1/` and the directory fallback only lists `*proj` at the solution root. The doc comment above therefore claims the opposite of what the test proves.

The red result isn't visible in CI: `Pretest` fails with *"Refusing to check out fork pull request code from a 'pull_request_target' workflow"*, so every test job on this PR reports `skipping`.

---

sca/bom/buildinfo/technologies/nuget/nuget.go
```diff
@@ -158,10 +158,21 @@ func runDotnetRestoreAndLoadSolution(params technologies.BuildInfoBomGeneratorPa
...
 		return
 	}
-	sol, err = solution.Load(tmpWd, "", exclusionPattern, log.Logger)
+	sol, err = solution.Load(tmpWd, solutionFileName(params.SolutionFilePath), exclusionPattern, log.Logger)
 	return
 }
 
+// solutionFileName returns the base file name of the solution file explicitly provided via the
+// SolutionFilePath param (e.g. the '--solution-path' flag), or an empty string if none was
+// provided. An empty string tells solution.Load to auto-discover '.sln'/'.slnx' files under the
```
gineshkumar, line 174: `solution.Load` does not auto-discover `.slnx` — its filter is `filepath.Ext(filePath) == ".sln"` (build-info-go `solution/solution.go:338`), so this comment sends the next reader looking for behavior that doesn't exist.

```suggestion
// provided. An empty string tells solution.Load to auto-discover '.sln' files under the
```

---

sca/bom/buildinfo/technologies/nuget/nuget.go
```diff
@@ -158,10 +158,21 @@ func runDotnetRestoreAndLoadSolution(params technologies.BuildInfoBomGeneratorPa
...
+// given directory, which is ambiguous when more than one solution file is present.
+func solutionFileName(solutionFilePath string) string {
+	if solutionFilePath == "" {
+		return ""
+	}
+	return filepath.Base(solutionFilePath)
+}
```
gineshkumar, line 181: `filepath.Base` discards the directory, and `solution.Load` re-joins the name onto its own `path` (`solution/solution.go:335`), so `--solution-path sub/My.sln` now resolves to `<wd>/My.sln` and `os.ReadFile` fails with a bare `no such file or directory` that never mentions the flag. Before this PR `solution.Load` received `""` and auto-discovered instead, so the mismatch surfaced only in the `dotnet restore` args.

Worth failing early with the flag named, so the user knows which input was wrong:

```suggestion
func solutionFileName(solutionFilePath string) string {
	if solutionFilePath == "" {
		return ""
	}
	if dir := filepath.Dir(solutionFilePath); dir != "." && dir != "" {
		log.Warn(fmt.Sprintf("'--solution-path' is resolved relative to the scanned directory; ignoring directory part of %q", solutionFilePath))
	}
	return filepath.Base(solutionFilePath)
}
```

---

sca/bom/buildinfo/technologies/nuget/nuget_test.go
```diff
@@ -286,3 +278,46 @@ func TestRunDotnetRestoreWithRealSolutionFile(t *testing.T) {
...
+	toolType := bidotnet.ConvertNameToToolType("dotnet")
+	for _, slnName := range []string{"TestSolution.sln", "TestAppOnly.sln"} {
+		t.Run(slnName, func(t *testing.T) {
+			params := technologies.BuildInfoBomGeneratorParams{
+				SolutionFilePath: filepath.Join(workDir, slnName),
+			}
+			err := runDotnetRestore(workDir, params, toolType, []string{})
+			if err != nil {
+				assert.NotContains(t, err.Error(), "this folder contains more than one project or solution file")
+			}
```
gineshkumar, line 364: `if err != nil { assert.NotContains(...) }` passes on every error that isn't the ambiguity message, so this test would still be green if `dotnet restore` failed for an unrelated reason. Asserting no error makes it prove what the comment claims:

```suggestion
			err := runDotnetRestore(workDir, params, toolType, []string{})
			assert.NoError(t, err)
```

Also missing: a case that passes a `.slnx` through `--solution-path` and asserts dependencies actually resolve — every `.slnx` assertion in this PR either checks string manipulation only (`TestSolutionFilePathParameter`) or is currently failing.

---

utils/techutils/techutils.go
gineshkumar: Correcting the version qualifier above: I verified `dotnet new sln` defaults to `.slnx` on SDK **10.0.302**, which is what I had available — not on 9.0.200. I conflated when `.slnx` became supported with when it became the default, so read that line as "on recent SDKs" until someone checks the exact boundary.

The nested-layout point is unaffected: `dotnet sln add` writes a subdirectory project path regardless of solution format, and every existing NuGet/.NET fixture in this repo that has a solution file nests its projects one level down.

---

Phavya-jfrog: Fixed upstream in build-info-go - https://github.com/jfrog/build-info-go/pull/404. Just need to bump the dependency here before merging; no other changes needed in jfrog-cli-security.

sca/bom/buildinfo/technologies/nuget/nuget_test.go
Phavya-jfrog: Test will pass once build-info-go (https://github.com/jfrog/build-info-go/pull/404) is merged and the dependency is bumped here in go.mod. No other changes needed.

sca/bom/buildinfo/technologies/nuget/nuget.go
Phavya-jfrog: no fix needed — same as the other two.

sca/bom/buildinfo/technologies/nuget/nuget.go
gineshkumar: You're right, and I over-severitized this. I verified on both heads with `sub/My.sln` + `sub/Lib/Lib.csproj`: `origin/dev` fails with `MSBUILD : error MSB1009: Project file does not exist.`, this branch fails with `open <wd>/My.sln: no such file or directory`. Same zero-dependency outcome — only the failure point and the message moved. This should have been **Low**, not Medium, and my phrase "tolerated" for the base-branch behaviour was simply wrong.

The guard as applied has a problem, though, and it's mine: `filepath.Dir` returns a real directory for **every** absolute path, so `dir != "."` is true whenever `--solution-path` is absolute — including the documented case where that directory *is* the scanned directory and nothing is being ignored. The ubuntu Unit Tests log on `e80d4eb` shows it firing three times on a correct invocation:

```
[Warn] '--solution-path' is resolved relative to the scanned directory; ignoring directory part of "/tmp/jfrog.cli.temp.-1785221798-1262357816/TestSolution.slnx"
[Warn] '--solution-path' ... (same, ×3)
[Info] Using solution file: TestSolution.slnx
```

Three times because `solutionFileName` is called at three sites (:49, :161, :282). Nothing was wrongly ignored there — that tmp directory is the scanned directory.

A comparison inside `solutionFileName` can't fix it cleanly either: it's called once with `wd` and once with `tmpWd`, so it has no stable base. Validating once in `BuildDependencyTree`, where `wd` is known, gives one actionable error and no false positives:

```go
if p := params.SolutionFilePath; p != "" {
	if _, statErr := os.Stat(filepath.Join(wd, filepath.Base(p))); statErr != nil {
		err = fmt.Errorf("--solution-path %q: no solution file named %q in %s "+
			"(the flag is resolved relative to the scanned directory)", p, filepath.Base(p), wd)
		return
	}
}
```

On your two options I'd take (1) — keep non-support, but fail clearly rather than warn. Option (2) is cheaper than it looks (`filepath.Rel` plus a containment check, and it wouldn't break these tests, which pass absolute paths equal to the cwd), but it reads as a clean standalone PR rather than scope on this one.

---

sca/bom/buildinfo/technologies/nuget/nuget.go
gineshkumar: Withdrawing this one — please don't apply the suggestion. Once build-info-go #404 lands, `getSlnFiles` matches both extensions, so `solution.Load` genuinely will auto-discover `.slnx` and this comment becomes accurate. My edit would make it wrong.

---

sca/bom/buildinfo/technologies/nuget/nuget_test.go
gineshkumar: CI now confirms this — `Pretest` passes on `e80d4eb` (the `dev` merge brought the checkout fix), so Unit Tests actually ran and are red on ubuntu, macOS and Windows:

```
--- FAIL: TestBuildDependencyTreeWithSlnxSolution (0.91s)
    Messages: expected at least one dependency tree to be built from the '.slnx' solution
--- FAIL: TestBuildDependencyTreeWithExplicitSlnxSolutionPath
    Messages: expected at least one dependency tree to be built from the '.slnx' solution via --solution-path
```

Worth noting so you don't debug them separately: the new explicit-path test fails for the **same** reason, not a distinct one. `--solution-path` can't help while the parser is missing — `getSlnFiles` locates the `.slnx`, `parseSlnFile`'s regex can't match `<Project Path="…" />`, so zero projects parse and the directory fallback then lists `*proj` non-recursively at the solution root, where this fixture has none.

Both should go green on the build-info-go bump, since `parseSlnxFile` returns projects and the non-recursive fallback is never reached. No change needed here beyond the bump.

---

sca/bom/buildinfo/technologies/nuget/nuget.go
```diff
@@ -46,7 +46,14 @@ func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (depen
 	if err != nil {
 		return
 	}
-	sol, err := solution.Load(wd, "", params.ExclusionPattern, log.Logger)
+	if p := params.SolutionFilePath; p != "" {
+		if _, statErr := os.Stat(filepath.Join(wd, filepath.Base(p))); statErr != nil {
```
gineshkumar: `os.Stat` succeeds on directories, so `--solution-path some/dir/` reduces to `dir` and passes this check whenever `wd/dir` exists — then fails inside `parseSlnFile` with `read …: is a directory`, which is the confusing error the check exists to prevent. My suggestion missed this; a mode check closes it.

```suggestion
		if info, statErr := os.Stat(filepath.Join(wd, filepath.Base(p))); statErr != nil || info.IsDir() {
```

The branch also has no test either way. This one is hermetic — the check returns before any solution load or `restore`, so no `dotnet` involved — and the second case fails today and passes with the fix above:

```go
// TestSolutionFilePathValidation covers the '--solution-path' pre-flight check in
// BuildDependencyTree. The check runs before any solution load or 'restore', so these
// cases never invoke the dotnet CLI.
func TestSolutionFilePathValidation(t *testing.T) {
	testCases := []struct {
		name             string
		solutionFilePath string
		setup            func(t *testing.T, dir string)
	}{
		{
			name:             "base name not present in scanned directory",
			solutionFilePath: filepath.Join("sub", "Missing.sln"),
		},
		{
			name:             "path resolves to a directory, not a solution file",
			solutionFilePath: filepath.Join("sub", "NotASolution"),
			setup: func(t *testing.T, dir string) {
				require.NoError(t, os.Mkdir(filepath.Join(dir, "NotASolution"), 0o750))
			},
		},
	}

	for _, test := range testCases {
		t.Run(test.name, func(t *testing.T) {
			dir := t.TempDir()
			if test.setup != nil {
				test.setup(t, dir)
			}
			t.Chdir(dir)

			dependencyTrees, uniqueDeps, err := BuildDependencyTree(technologies.BuildInfoBomGeneratorParams{
				SolutionFilePath: test.solutionFilePath,
			})
			require.Error(t, err)
			assert.Contains(t, err.Error(), "--solution-path", "the error must name the flag the user passed")
			assert.Contains(t, err.Error(), filepath.Base(test.solutionFilePath))
			assert.Nil(t, dependencyTrees)
			assert.Nil(t, uniqueDeps)
		})
	}
}
```

Traced rather than executed — Artifactory is returning 401 for every JFrog module here, so I could not run it locally. Worth a run before taking it.

---

gineshkumar (approved):

go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/gookit/color v1.6.1
 	github.com/hashicorp/go-hclog v1.6.3
 	github.com/hashicorp/go-plugin v1.6.3
-	github.com/jfrog/build-info-go v1.13.1-0.20260728083052-16a97012811d
-	github.com/jfrog/froggit-go v1.23.1
+	github.com/jfrog/build-info-go v1.13.1-0.20260728131356-58a6ccca0214
+  github.com/jfrog/froggit-go v1.23.1
```
attiasas: whitespace corruption — froggit-go line uses spaces instead of a tab; restore the tab

go.mod
```diff
@@ -14,8 +14,8 @@ require (
 	github.com/gookit/color v1.6.1
 	github.com/hashicorp/go-hclog v1.6.3
 	github.com/hashicorp/go-plugin v1.6.3
-	github.com/jfrog/build-info-go v1.13.1-0.20260728083052-16a97012811d
-	github.com/jfrog/froggit-go v1.23.1
+	github.com/jfrog/build-info-go v1.13.1-0.20260728131356-58a6ccca0214
```
attiasas: approve + merge only after `build-info` PR is merged and `go.mod` is updated to point to head

attiasas (commented):
LGTM, Nice work!
Lets wait for build-info PR to merge before continue here

attiasas (approved):
LGTM

PR #819, opened 2026-07-31
XRAY-156190 - Implement all packages blocked in packument fallback logic for npm

gauriy-tech:

-----

## Summary

Adds a fallback for `jf ca` (npm): when `npm install --package-lock-only` crashes mid-resolution due to a curation-blocked package, this reconstructs a partial report from npm's own debug log instead of aborting with zero findings. Classifies each unresolved entry as whole-package-blocked, non-registry (git URL/local ref/alias — never probed), unresolvable range (advisory only), or genuine ETARGET (still probed). Scoped to npm only — yarn/pnpm untouched.

Before
Screenshot 2026-08-03 at 2 21 53 PM

After
Screenshot 2026-08-03 at 2 22 02 PM

### DIFF
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -62,11 +62,13 @@ import (
 
 const (
 	// The "blocked" represents the unapproved status that can be returned by the Curation Service for dependencies..
-	blocked                = "blocked"
-	BlockingReasonPolicy   = "Policy violations"
-	BlockingReasonNotFound = "Package pending update"
-	BlockingReasonOnDemand = "Package pending — Curation on-demand scan in progress"
-	BlockingReasonUnknown  = "Blocked by curation (response could not be parsed)"
+	blocked                    = "blocked"
+	notEvaluated               = "not evaluated"
+	BlockingReasonPolicy       = "Policy violations"
+	BlockingReasonNotFound     = "Package pending update"
+	BlockingReasonOnDemand     = "Package pending — Curation on-demand scan in progress"
+	BlockingReasonUnknown      = "Blocked by curation (response could not be parsed)"
+	BlockingReasonNotEvaluated = "Not evaluated — non-registry dependency"
 
 	directRelation   = "direct"
 	indirectRelation = "indirect"
@@ -105,10 +107,20 @@ const (
 
 	// hfUnresolvedReportKey is used when an HF scan found only dynamic references — no table, just warnings.
 	hfUnresolvedReportKey = "huggingface (unresolved references)"
+
+	// npmLogPartialReportWarning is shown when runNpmLogFallback recovers a partial report.
+	// npmLogOriginalErr (the real underlying npm error) goes to log.Debug only, not here —
+	// npm's own error text can include its full captured stdout/stderr, far too noisy for a
+	// console warning.
+	npmLogPartialReportWarning = "npm install could not fully resolve the dependency tree because one or more packages " +
+		"are blocked by the curation policy. This report was reconstructed from npm's debug log after the installation " +
+		"failed, so it may be incomplete. Dependencies of blocked packages could not be analyzed because their metadata " +
+		"was unavailable."
 )
 
 var CurationOutputFormats = []string{string(outFormat.Table), string(outFormat.Json)}
 var osGetwd = os.Getwd
+var npmGetConfigValue = npmtech.GetNpmConfigValue
 
 var supportedTech = map[techutils.Technology]func(ca *CurationAuditCommand) (bool, error){
 	techutils.Npm:  func(ca *CurationAuditCommand) (bool, error) { return true, nil },
@@ -289,6 +301,12 @@ type CurationReport struct {
 	// Kept separate from isPartial, which also triggers the CVS-specific warning
 	// and skips the waiver flow.
 	hfPartial bool
+	// npmLogPartial marks a report produced by runNpmLogFallback. Set alongside isPartial
+	// (to reuse its waiver-skip gate), but kept distinct for its own warning text and
+	// partialReason.
+	npmLogPartial bool
+	// npmLogOriginalErr is the real install error, surfaced alongside npmLogPartialReportWarning.
+	npmLogOriginalErr string
 	// warnings holds user-facing messages from tree-build (e.g. unresolved HF references).
 	warnings []string
 	// huggingFaceReport marks reports produced by the Hugging Face audit path (including
@@ -426,7 +444,12 @@ func (ca *CurationAuditCommand) Run() (err error) {
 	}
 	for projectPath, report := range results {
 		if report.isPartial {
-			log.Warn(fmt.Sprintf("[%s] %s", projectPath, cvsPartialReportWarning))
+			warningText := cvsPartialReportWarning
+			if report.npmLogPartial {
+				warningText = npmLogPartialReportWarning
+				log.Debug(fmt.Sprintf("[%s] underlying npm error: %s", projectPath, report.npmLogOriginalErr))
+			}
+			log.Warn(fmt.Sprintf("[%s] %s", projectPath, warningText))
 		}
 	}
 
@@ -475,20 +498,31 @@ func (ca *CurationAuditCommand) Run() (err error) {
 func convertResultsToSummary(results map[string]*CurationReport) formats.ResultsSummary {
 	summaryResults := formats.ResultsSummary{}
 	for projectPath, packagesStatus := range results {
-		// hfPartial reports must still be recorded, not treated as "HF wasn't attempted".
-		if isWarningsOnlyReport(packagesStatus) && !packagesStatus.hfPartial {
+		// hfPartial/npmLogPartial reports must still be recorded, not treated as "not attempted".
+		if isWarningsOnlyReport(packagesStatus) && !packagesStatus.hfPartial && !packagesStatus.npmLogPartial {
 			continue
 		}
 		var partialReason string
 		switch {
+		// npmLogPartial is checked before isPartial: an npm-log-fallback report sets both
+		// (isPartial to reuse the existing waiver-skip gate below), so npmLogPartial must
+		// win here or it would incorrectly report "cvs_fallback".
+		case packagesStatus.npmLogPartial:
+			// jfrog-ignore: categorical tag, not hardcoded credentials
+			partialReason = "npm_log_fallback"
 		case packagesStatus.isPartial:
 			partialReason = "cvs_fallback"
 		case packagesStatus.hfPartial:
 			partialReason = "hf_unresolved"
 		}
+		// npmLogPartial's totalNumberOfPackages includes non-remediable "not evaluated" rows, so use the blocked-row count instead.
+		packageCount := packagesStatus.totalNumberOfPackages
+		if packagesStatus.npmLogPartial {
+			packageCount = countBlockedPackages(packagesStatus.packagesStatus)
+		}
 		summaryResults.Scans = append(summaryResults.Scans, formats.ScanSummary{Target: projectPath,
 			CuratedPackages: &formats.CuratedPackages{
-				PackageCount:  packagesStatus.totalNumberOfPackages,
+				PackageCount:  packageCount,
 				Blocked:       getBlocked(packagesStatus.packagesStatus),
 				IsPartial:     packagesStatus.isPartial || packagesStatus.hfPartial,
 				PartialReason: partialReason,
@@ -498,9 +532,23 @@ func convertResultsToSummary(results map[string]*CurationReport) formats.Results
 	return summaryResults
 }
 
+// countBlockedPackages counts only rows with Action == blocked, excluding "not evaluated" rows.
+func countBlockedPackages(packagesStatus []*PackageStatus) int {
+	count := 0
+	for _, ps := range packagesStatus {
+		if ps.Action == blocked {
+			count++
+		}
+	}
+	return count
+}
+
 func getBlocked(pkgStatus []*PackageStatus) []formats.BlockedPackages {
 	blockedMap := map[string]formats.BlockedPackages{}
 	for _, pkg := range pkgStatus {
+		if pkg.Action != blocked {
+			continue
+		}
 		for _, policy := range pkg.Policy {
 			polAndCondKey := getPolicyAndConditionId(policy.Policy, policy.Condition)
 			if _, ok := blockedMap[polAndCondKey]; !ok {
@@ -949,6 +997,42 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
 	if err != nil {
 		return err
 	}
+	// Discovered before installing, not after: npm writes a log on every invocation, so running
+	// this after a crash would make its own log newer than the crash's. Failure isn't fatal.
+	//
+	// Uses logs-dir, not cache: <cache>/_logs is only npm's default, overridable independently.
+	// "npm config get logs-dir" prints the literal string "null" when unset, hence the fallback.
+	var npmLogsDir string
+	var npmLogsMaxWasZero bool
+	var npmLogBaselineKey int64
+	var npmConfigLookupErr error
+	if tech == techutils.Npm {
+		if workDir, wdErr := osGetwd(); wdErr != nil {
+			npmConfigLookupErr = fmt.Errorf("failed to determine working directory for npm config lookup: %w", wdErr)
+		} else if logsDirValue, logsDirErr := npmGetConfigValue(workDir, "logs-dir"); logsDirErr != nil {
+			npmConfigLookupErr = fmt.Errorf("failed to determine npm logs-dir: %w", logsDirErr)
+		} else if logsMaxValue, logsMaxErr := npmGetConfigValue(workDir, "logs-max"); logsMaxErr != nil {
+			npmConfigLookupErr = fmt.Errorf("failed to determine npm logs-max config: %w", logsMaxErr)
+		} else if logsDirValue != "" && logsDirValue != "null" {
+			npmLogsDir, npmLogsMaxWasZero = logsDirValue, logsMaxValue == "0"
+		} else if cacheDir, cacheErr := npmGetConfigValue(workDir, "cache"); cacheErr != nil {
+			npmConfigLookupErr = fmt.Errorf("failed to determine npm cache directory: %w", cacheErr)
+		} else {
+			npmLogsDir, npmLogsMaxWasZero = filepath.Join(cacheDir, "_logs"), logsMaxValue == "0"
+		}
+		if npmConfigLookupErr != nil {
+			log.Debug(fmt.Sprintf("npm curation debug-log fallback will be unavailable if the install fails: %v", npmConfigLookupErr))
+		}
+		if npmLogsMaxWasZero {
+			// Only override when the ambient config would otherwise suppress the debug log
+			// entirely — never force a value that prunes a user's own nonzero retention setting.
+			params.NpmForceLogsMax = "1"
+		}
+		if npmLogsDir != "" {
+			// Snapshotted last so the baseline includes the config-lookup calls' own logs too.
+			npmLogBaselineKey, _ = npmDebugLogNewestKey(npmLogsDir)
+		}
+	}
 	depTreeResult, err := buildinfo.GetTechDependencyTree(params, serverDetails, tech)
 	if err != nil {
 		// When CVS strips a pinned version from the simple index, pip can't
@@ -959,7 +1043,18 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
 		if (tech == techutils.Pip || tech == techutils.Poetry) && errors.As(err, &cvsErr) {
 			return ca.runCvsFallback(cvsErr, tech, results)
 		}
+		// npm's install error is a plain opaque error, nothing to type-match on — trigger the
+		// debug-log fallback unconditionally, unless the config lookup above already failed.
+		if tech == techutils.Npm {
+			if npmConfigLookupErr != nil {
+				return err
+			}
+			return ca.runNpmLogFallback(npmLogsDir, tech, results, err, npmLogsMaxWasZero, npmLogBaselineKey)
+		}
 		return err
+	}
+	if tech == techutils.Npm {
+		defer cleanupForcedNpmDebugLog(npmLogsDir, npmLogsMaxWasZero, npmLogBaselineKey)
 	}
 	// Validate the graph isn't empty.
 	if len(depTreeResult.FullDepTrees) == 0 {
@@ -1212,7 +1307,7 @@ func printResult(format outFormat.OutputFormat, projectPath string, packagesStat
 	if format == "" {
 		format = outFormat.Table
 	}
-	log.Output(fmt.Sprintf("Found %v blocked packages for project %s", len(packagesStatus), projectPath))
+	log.Output(fmt.Sprintf("Found %v blocked packages for project %s", countBlockedPackages(packagesStatus), projectPath))
 	switch format {
 	case outFormat.Json:
 		if len(packagesStatus) > 0 {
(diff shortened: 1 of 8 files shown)


gauriy-tech:

commands/curation/curationaudit.go
```diff
@@ -1861,6 +1912,218 @@ func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, t
...
+				continue
+			}
+			seenRangeWarnings[rangeKey] = true
+			advisoryWarnings = append(advisoryWarnings, fmt.Sprintf(
+				"Package '%s' (requested as '%s') could not be resolved to a specific version, so its curation status could not be checked. This is not necessarily a curation block — verify the dependency resolves correctly.",
+				entry.Name, entry.Specifier))
+			continue
+		}
+
+		key, keyErr := npmPackageKey(tech, analyzer.url, analyzer.repo, entry.Name, entry.Version)
```
Phavya-jfrog: **Dedup/storage key collision for same-name entries with different specifiers (High)**

`key` is derived from `entry.Name` + `entry.Version`, but `entry.Version` is blank for every blank-version category (`npmEntryETARGET`, `npmEntryNonRegistrySpecifier`). If the same package name appears twice with different specifiers from different parents (e.g. `foo` wanted as exact `1.0.0` by one parent and exact `2.0.0` by another, both genuinely blocked), both entries collapse onto the same key here, in the `preProcessMap.Store` calls below, and in `buildGraphFromLogEntries`'s `npmNodeId(entry.Name, entry.Version)` (npmlogparser.go:265).

The result isn't a dropped row — both parents' edges still produce a row (via `fillGraphRelations`'s per-edge dedup on `name+version+parent`), but both rows get the *same* (first-probed) version/policy. The second parent's real blocked version is silently discarded and misreported as the first's.

Verified with a test: two ETARGET entries for `widget` (versions `2.0.0` and `3.0.0`, different parents) — both rows report `PackageVersion=2.0.0`; `3.0.0` never appears. The same collision reproduces for two `npmEntryNonRegistrySpecifier` entries with the same name but different specifiers (e.g. a git URL vs. a `file:` path).

Suggested fix: disambiguate the key/graph-node-id for every blank-version category whose row content varies by specifier (`npmEntryETARGET` *and* `npmEntryNonRegistrySpecifier` — not just ETARGET), e.g.:

```suggestion
		key, keyErr := npmPackageKey(tech, analyzer.url, analyzer.repo, entry.Name, effectiveGraphVersion(entry, blockedPackages))
```

with a helper:
```go
func effectiveGraphVersion(entry npmLogEntry, blockedPackages map[string]npmBlockedInfo) string {
	if entry.Version != "" {
		return entry.Version
	}
	switch classifyBlankVersionEntry(entry, blockedPackages) {
	case npmEntryETARGET, npmEntryNonRegistrySpecifier:
		return entry.Specifier
	default:
		return entry.Version // whole-package-blocked: content doesn't vary by specifier, safe to collapse
	}
}
```
threaded through `buildGraphFromLogEntries`'s `npmNodeId` call too, so the graph node identity and the map key stay consistent.

<sub>🤖 Posted by an automated PR review (xray-pr-review).</sub>

commands/curation/npmlogparser.go
```diff
@@ -0,0 +1,312 @@
...
+	Condition string
+}
+
+var (
+	// Matches: 42 notice All versions blocked - {policy:X,condition:Y}
+	npmNoticeBlockedRegex = regexp.MustCompile(`^\d+\s+notice\s+All versions blocked - \{policy:([^,]+),condition:([^}]+)\}`)
+	// Matches: 43 http fetch GET 403 https://.../api/npm/<repo>/<name> ...
+	npmFetch403Regex = regexp.MustCompile(`^\d+\s+http fetch GET 403\s+\S*/api/npm/[^/\s]+/(\S+)\s`)
+	// Matches: 110 silly placeDep ROOT accepts@2.0.0 OK for: express@5.2.1 want: ^2.0.0
+	npmPlaceDepRegex = regexp.MustCompile(
```
Phavya-jfrog, line 66: **Regex drops any `want:` specifier containing a space (Medium)**

The trailing group `want:\s+(\S+)\s*$` requires the specifier to contain no whitespace. Two legitimate npm version-range syntaxes contain a literal space in the `want:` clause of a real `silly placeDep` debug-log line: OR-ranges (`1.0.0 || 2.0.0`) and hyphen-ranges (`1.0.0 - 2.0.0`). For both, the *entire line* fails to match, so the dependency is silently dropped before it ever becomes an `npmLogEntry` — no row, and (unlike the `npmEntryUnresolvableRange` path) no advisory warning either.

Verified directly: `npmPlaceDepRegex.FindStringSubmatch(...)` returns `nil` for both `"... want: 1.0.0 || 2.0.0"` and `"... want: 1.0.0 - 2.0.0"`.

```suggestion
		`^\d+\s+silly\s+placeDep\s+(\S+)\s+((?:@[^@\s]+/)?[^@\s]+)@(\S*)\s+OK for:\s+((?:@[^@\s]+/)?[^@\s]+)@(\S+)\s+want:\s+(.+?)\s*$`)
```

<sub>🤖 Posted by an automated PR review (xray-pr-review).</sub>

commands/curation/curationaudit.go
```diff
@@ -1861,6 +1912,218 @@ func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, t
...
+		}
+	}
+
+	if len(packagesStatus) == 0 && len(advisoryWarnings) == 0 {
+		return originalErr
+	}
+
+	results[uniqueReportKey(results, filepath.Base(workPath), tech)] = &CurationReport{
+		packagesStatus:        packagesStatus,
+		totalNumberOfPackages: len(packagesStatus),
```
Phavya-jfrog, line 2136: **"Found %v blocked packages" header and `totalNumberOfPackages` overcount when non-registry rows are present (Medium)**

`totalNumberOfPackages: len(packagesStatus)` here counts every row, including the new `nonRegistrySpecifierStatus` rows (`Action: notEvaluated`, not `blocked`). The same unconditional `len(packagesStatus)` is used in `printResult`'s `"Found %v blocked packages for project %s"` header (curationaudit.go:1299, not shown in this diff since it's pre-existing code, but it flows the same slice). Before this PR every row actually was `blocked`, so the count was accurate; now a report with 1 truly blocked package and 2 git-URL/local-reference "not evaluated" rows will print "Found 3 blocked packages" and report a `PackageCount` of 3 in the summary — both misleading to a user trying to gauge how many packages actually need remediation.

Suggested fix: count only rows where `Action == blocked` at both call sites:

```go
blockedCount := 0
for _, ps := range packagesStatus {
	if ps.Action == blocked {
		blockedCount++
	}
}
```

<sub>🤖 Posted by an automated PR review (xray-pr-review).</sub>

commands/curation/npmlogparser.go
```diff
@@ -0,0 +1,312 @@
...
+type npmBlockedInfo struct {
+	Policy    string
+	Condition string
+}
+
+var (
+	// Matches: 42 notice All versions blocked - {policy:X,condition:Y}
+	npmNoticeBlockedRegex = regexp.MustCompile(`^\d+\s+notice\s+All versions blocked - \{policy:([^,]+),condition:([^}]+)\}`)
+	// Matches: 43 http fetch GET 403 https://.../api/npm/<repo>/<name> ...
+	npmFetch403Regex = regexp.MustCompile(`^\d+\s+http fetch GET 403\s+\S*/api/npm/[^/\s]+/(\S+)\s`)
```
Phavya-jfrog: **Regex also matches tarball-fetch 403 lines, desyncing the `pendingNotices` FIFO queue (High)**

`(\S+)` isn't anchored against `/`, so it also matches a *tarball*-fetch 403 line (e.g. `.../api/npm/my-npm-remote/foo/-/foo-1.0.0.tgz`), capturing the garbage segment `"foo/-/foo-1.0.0.tgz"` as if it were a packument match. Since npm fetches packuments and tarballs concurrently, and the code's own comment two lines up already acknowledges that "concurrent fetches can leave multiple notices pending at once," a per-version-blocked package's tarball 403 landing between a whole-package-block `notice` line and its real matching packument-403 line will incorrectly consume the pending notice and advance the FIFO — the real whole-package block is then lost or misattributed to the wrong package.

Verified with a synthetic interleaved log: a `notice` for package A, followed by an unrelated tarball-403 for package B, followed by A's real packument-403 — `blockedPackages` ends up missing `A` entirely and contains a garbage key for `B`.

```suggestion
	npmFetch403Regex = regexp.MustCompile(`^\d+\s+http fetch GET 403\s+\S*/api/npm/[^/\s]+/((?:@[^/\s]+/)?[^/\s]+)\s`)
```
(disallows embedded `/` in the captured name, so tarball paths — which always contain `/-/` — no longer match)

<sub>🤖 Posted by an automated PR review (xray-pr-review).</sub>

commands/curation/npmlogparser.go
```diff
@@ -0,0 +1,312 @@
...
+	var newestModTime int64
+	for _, de := range dirEntries {
+		if de.IsDir() || !strings.HasSuffix(de.Name(), ".log") {
+			continue
+		}
+		info, infoErr := de.Info()
+		if infoErr != nil {
+			continue
+		}
+		if modTime := info.ModTime().UnixNano(); newestPath == "" || modTime > newestModTime {
```
Phavya-jfrog: **mtime tie-break favors the older file, backwards from intent (Low)**

Strict `>` means that on an exact `ModTime()` tie, the first-encountered file (in `os.ReadDir`'s lexical order, which coincides with chronological order since these filenames are ISO-8601-prefixed) wins — i.e. the *older* file, not the newer one this function is meant to find. This needs either coarse filesystem mtime resolution or two log files written within the same tick to trigger, so it's narrow, but reproducible deterministically (e.g. via `os.Chtimes` forcing a tie in a test) and inverts the function's stated purpose.

Suggested fix: parse the embedded ISO-8601 timestamp from the filename as the primary sort key, falling back to `ModTime()` only if the filename can't be parsed.

<sub>🤖 Posted by an automated PR review (xray-pr-review).</sub>

Phavya-jfrog (approved):

commands/curation/npmlogparser.go
gineshkumar: **Whole-package blocks are silently lost for scoped packages.**

`npmFetch403Regex` keys `blockedPackages` on the raw URL path segment, but npm (via pacote/npm-package-arg `escapedName`) percent-encodes the scope separator in packument requests — so the log line is `.../api/npm/<repo>/@babel%2fcore`, and the key becomes `@babel%2fcore`. `classifyBlankVersionEntry` then looks up `blockedPackages[entry.Name]`, where `entry.Name` comes from `npmPlaceDepRegex` as `@babel/core`. The lookup misses.

Verified by running both regexes against a real-shaped line:

```
403 line  -> key    = "@babel%2fcore"
placeDep  -> lookup = "@babel/core"     => MISS
```

Downstream effect for the common `"@babel/core": "^7.0.0"` shape: the entry falls through to `classifyNpmSpecifier("^7.0.0")`, which returns `probeable=true, resolvedVer="7.0.0"`, so `entry.Specifier != resolvedVer` hits and the entry is classified `npmEntryUnresolvableRange`. It is dropped from the blocked table and instead emits:

> "…could not be resolved to a specific version… **This is not necessarily a curation block** — verify the dependency resolves correctly."

So a scoped package that *is* blocked by policy is reported as probably-not-a-block. Scoped + caret range is the dominant npm dependency shape (`@types/*`, `@babel/*`, `@angular/*`, `@aws-sdk/*`), and this is the feature's headline case.

Suggested fix — normalize the captured name before storing:

```go
name := m[1]
if decoded, decErr := url.PathUnescape(name); decErr == nil {
    name = decoded
}
blockedPackages[name] = pendingNotices[0]
```

(`%2F` uppercase also occurs, so decode rather than string-replace.)

No test in the PR uses a scoped package — the fixture and all unit cases use unscoped names only, which is why this passes CI. Worth adding a scoped whole-package-blocked case to `TestParseNpmDebugLog` and to the `mixed-crash` fixture.

commands/curation/curationaudit.go
```diff
@@ -977,6 +1000,25 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
...
 	}
+	// Discovered before installing, not after: npm writes a log on every invocation, so running
+	// this after a crash would make its own log newer than the crash's. Failure isn't fatal.
+	var npmCacheDir string
+	var npmLogsMaxWasZero bool
+	var npmConfigLookupErr error
+	if tech == techutils.Npm {
+		if workDir, wdErr := osGetwd(); wdErr != nil {
+			npmConfigLookupErr = fmt.Errorf("failed to determine working directory for npm config lookup: %w", wdErr)
+		} else if cacheDir, cacheErr := npmGetConfigValue(workDir, "cache"); cacheErr != nil {
```
gineshkumar: **The fallback reads `<cache>/_logs`, but npm's log location is the `logs-dir` config, not the cache.**

`auditTree` resolves the log directory from `npm config get cache` and `findNewestNpmDebugLog` joins `_logs` onto it. npm has a dedicated config for this:

> `logs-dir` — Default: "A directory named `_logs` inside the cache". The location of npm's log directory.

`<cache>/_logs` is only the *default*. Any user or CI image that sets `logs-dir` (common when the cache lives on a fast ephemeral volume but logs are collected elsewhere) gets a directory that doesn't exist, `findNewestNpmDebugLog` returns `""`, and the whole feature silently no-ops — the user sees the original opaque `npm install` error with no indication that recovery was attempted and missed.

Suggested fix: query `logs-dir` and use it directly, falling back to `<cache>/_logs` only if the key is unsupported on the installed npm (`logs-dir` predates the npm 7 floor this code already assumes, so a plain switch is likely fine). That also removes the `_logs` join from `findNewestNpmDebugLog`.

sca/bom/buildinfo/technologies/npm/npm.go
```diff
@@ -189,6 +204,11 @@ func createTreeDepsParam(params *technologies.BuildInfoBomGeneratorParams) biuti
 	if params.NpmLegacyPeerDeps {
 		installCommandArgs = appendUniqueFlag(installCommandArgs, LegacyPeerDepsFlag)
 	}
+	if params.IsCurationCmd {
+		// Force a debug log for the crash-recovery fallback to read, without touching --cache.
+		// Appended last so this flag can't be shadowed by an earlier one.
+		installCommandArgs = append(installCommandArgs, "--logs-max", "10")
```
gineshkumar: **Forcing `--logs-max 10` overrides user config, and the `logs-max=0` case is only half-handled.**

`--logs-max 10` is appended for every curation npm install, unconditionally. Two consequences:

1. **`logs-max=0` users get logs written and left behind on the success path.** npm's docs: "If set to 0, no log files will be written for the current run." The cleanup that respects this (`os.Remove(logFilePath)` when `logsMaxWasZero`) lives inside `runNpmLogFallback`, which only runs when the install *fails*. On the overwhelmingly common successful audit, the forced log is written and never removed. `logs-max=0` is often set deliberately — npm debug logs record full registry URLs and whatever was on the npm argv, so this is a compliance-relevant behaviour change, not just stray disk use.

2. **`logs-max` above 10 gets pruned down.** npm's log cleanup deletes oldest-first to satisfy `logs-max`, so a user with `logs-max=50` has their retained npm logs cut to ~10 as a side effect of running `jf ca`.

`auditTree` already reads the real value into `npmLogsMaxWasZero`, so the information needed to avoid both is available — it just isn't threaded to `createTreeDepsParam`. Options: pass the resolved value through `BuildInfoBomGeneratorParams` and only override when it is `0` (and then use `--logs-max 1`, since one log is all the fallback needs); or keep the override and clean up unconditionally when `logsMaxWasZero`, including on the success path.

commands/curation/npmlogparser.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	if scanErr := scanner.Err(); scanErr != nil {
+		return nil, nil, logFile, errorutils.CheckErrorf("failed to scan npm debug log %q: %v", logFile, scanErr)
+	}
+	return entries, blockedPackages, logFile, nil
+}
+
+// findNewestNpmDebugLog returns the most recently modified *-debug-0.log under <logDir>/_logs/,
+// or "" if none exist.
+func findNewestNpmDebugLog(logDir string) (string, error) {
```
gineshkumar: **Only the newest log *chunk* is read — npm splits a single run's log at 50,000 lines.**

npm's `log-file.js` rotates within one run at `#MAX_LOGS_PER_FILE = 50_000`, producing `<ts>-debug-0.log`, `<ts>-debug-1.log`, … for the same invocation. `findNewestNpmDebugLog` returns exactly one file (newest mtime), so on a run that spills past 50k lines the parser sees only the last chunk.

That is the worst possible truncation for this design: `placeDep` lines are emitted in resolution order, so chunk 0 holds the `ROOT`-location entries. Losing it means `rootIdentityFromLogEntries` finds no `ROOT` entry, `directDepNames` comes out empty (every block reported as transitive with a wrong parent), and `buildGraphFromLogEntries` builds a tree whose root id matches nothing in `treeMap` — so `GraphsRelations` walks nothing and the report comes back empty, returning `originalErr` as if nothing were recoverable.

A debug log is not only `placeDep` lines (every `http fetch`, `silly`, `timing` line counts), so 50k is reachable on a large monorepo — which is exactly the population most likely to trip a curation block in the first place.

Suggested fix: in `findNewestNpmDebugLog`, group by the timestamp prefix rather than returning a single file, and have `parseNpmDebugLog` read all chunks of the newest run in ascending chunk order. The zero-padded chunk suffix npm uses sorts lexicographically, so `sort.Strings` over the group is sufficient.

commands/curation/curationaudit.go
```diff
@@ -1862,6 +1912,218 @@ func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, t
...
+
+// nonRegistrySpecifierStatus reports a git URL/local-reference/npm-alias dependency as its
+// own row with "—" Policy/Condition — a third state distinct from clean/blocked, since npm
+// never contacts the registry for these at all and curation's enforcement never applies.
+// Never makes a network call on this package's behalf.
+func nonRegistrySpecifierStatus(entry npmLogEntry, tech techutils.Technology) *PackageStatus {
+	return &PackageStatus{
+		PackageName:    entry.Name,
+		PackageVersion: fmt.Sprintf("%s (not evaluated)", entry.Specifier),
+		Action:         notEvaluated,
```
gineshkumar: **"Not evaluated" rows are still counted and rendered as blocked packages.**

`nonRegistrySpecifierStatus` correctly avoids the misleading `Action`/`BlockingReason` values — `TestNonRegistrySpecifierStatus_BlockingReasonAndActionNotMisleading` locks that in. But neither field reaches the two surfaces the user actually reads:

1. `printResult` prints `Found %v blocked packages` over `len(packagesStatus)`, so a git-URL dependency that curation never contacted is counted as blocked.
2. `PackageStatusTable` has no `Action` or `BlockingReason` column — only `PackageName`, `PackageVersion` (under the header `Blocked Package Version`), `Policy`, `Condition`, `Explanation`, `Recommendation`. In table output the only signal is the `(not evaluated)` text appended to the version string and the `—` in the policy columns.
3. `getBlocked` iterates `pkg.Policy` unconditionally, so the synthetic `{Policy: "—", Condition: "—"}` entry lands in `formats.CuratedPackages.Blocked` and is recorded via `RecordSecurityCommandSummary` as a blocked package under a policy literally named `—`.

Net effect: `jf ca --format=json` distinguishes the third state correctly; the default table and the CI job summary do not. Anything downstream that counts blocked packages from the summary will over-count.

Options: exclude `Action == notEvaluated` rows from the `printResult` count and from `getBlocked`, and report them as warnings alongside the unresolvable-range advisories (they are advisory in nature — no policy was evaluated); or keep them in the table but add an `Action`/status column so the distinction is visible where it is rendered.

commands/curation/npmlogparser.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return nil, nil, logFile, errorutils.CheckErrorf("failed to open npm debug log %q: %v", logFile, err)
+	}
+	defer func() {
+		_ = f.Close()
+	}()
+
+	blockedPackages = map[string]npmBlockedInfo{}
+	// FIFO queue, not a single value: concurrent fetches can leave multiple notices pending at once.
+	var pendingNotices []npmBlockedInfo
```
gineshkumar: **Question on the notice→403 pairing:** what guarantees the FIFO order holds?

The queue and `TestParseNpmDebugLog_MultiplePendingNoticesPairInArrivalOrder` establish that *if* notices and their 403 lines are emitted in the same relative order, attribution is correct. The comment says the queue exists precisely because "concurrent fetches can leave multiple notices pending at once" — but with concurrency, same-relative-order is the assumption, not a guarantee: the notice comes from a response header handler and the `http fetch GET 403` line from the request-completion logger, so two in-flight packument fetches can interleave those two emissions independently.

When they do interleave, the failure is silent and specific: `pkg-a` gets `pkg-b`'s policy/condition, and vice versa. The report is confidently wrong rather than incomplete, and there is nothing in the parse that could detect it.

Is there something in npm's emit path that makes the ordering deterministic per response? If not, worth either (a) documenting the limitation where `pendingNotices` is declared so the next reader knows the attribution is best-effort, or (b) dropping policy/condition to `—` and letting the notice only mark *that* the package is wholly blocked, when more than one notice is pending at the time a 403 is matched — degrading to "blocked, policy unknown" beats naming the wrong policy.

commands/curation/curationaudit.go
```diff
@@ -1862,6 +1912,218 @@ func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, t
 	return nil
 }
 
+// runNpmLogFallback reconstructs a partial curation report from npm's debug log
+// (cacheDir/_logs/) when the npm tree-build install crashes, since buildDeps completes
+// before reify can fail. Falls back to originalErr unchanged if nothing is recoverable.
+// See NPM_CURATION_DEBUGLOG_FALLBACK_DESIGN for the full design.
```
gineshkumar: Two small doc nits in this file:

1. `NPM_CURATION_DEBUGLOG_FALLBACK_DESIGN` doesn't exist anywhere in the repo (checked the full tree at this commit and at `dev`). Either add the doc or point at the Jira ticket / PR instead, so the next reader isn't hunting for a file that was never committed.
2. `findNewestNpmDebugLog`'s comment says it returns "the most recently modified `*-debug-0.log`", but the code filters on `strings.HasSuffix(de.Name(), ".log")` — any `.log` file qualifies. Worth aligning the comment with the code (or the filter with the comment — see the log-chunking comment on this PR, which argues the filter should become timestamp-group-aware anyway).

commands/curation/npmlogparser.go
gauriy-tech: No — nothing in npm's emit path guarantees it. Notices and their matching 403 lines come from separate emit paths (response-handler vs. request-completion logger), so same-relative-order isn't provable under concurrency.

I didn't degrade to "policy unknown" on ambiguity because it would break a real, already-tested case: two packages whole-package-blocked concurrently legitimately queue two notices before either 403 arrives, and pairing them in order is correct — that's indistinguishable in the log from the bad (crossed) case you're worried about. Degrading on ≥2 pending would misattribute the common case for certain, to guard against an edge case we can't confirm is reachable.

Documented the limitation at the pendingNotices declaration instead: best-effort, not a guarantee, with the reason why.

commands/curation/curationaudit.go
```diff
@@ -987,8 +1061,19 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
...
 			return ca.runCvsFallback(cvsErr, tech, results)
 		}
+		// npm's install error is a plain opaque error, nothing to type-match on — trigger the
+		// debug-log fallback unconditionally, unless the config lookup above already failed.
+		if tech == techutils.Npm {
+			if npmConfigLookupErr != nil {
+				return err
+			}
+			return ca.runNpmLogFallback(npmLogsDir, tech, results, err, npmLogsMaxWasZero)
+		}
```
attiasas, line 1081: **Critical: fallback can recover from a stale debug log and return success**

Any `GetTechDependencyTree` failure for npm enters `runNpmLogFallback` unconditionally. Log discovery runs *before* install (`npmLogsDir` at ~1023), but there is no gate that the chosen log was written by *this* invocation.

If install fails before npm writes a new log (missing `npm`, bad `package.json`, resolution-server config, etc.), `findNewestNpmDebugLogChunks` can parse an older run still sitting in `logs-dir` / `<cache>/_logs`. When that log has `placeDep` lines, we store a partial report and `return nil` (`runNpmLogFallback` ~2113), swallowing `originalErr`.

Shared CI caches commonly retain npm logs — this is a false-report / false-success risk.

**Suggested fix:** snapshot log filenames (or newest timestamp) before `GetTechDependencyTree`; only accept chunks newer than that snapshot. If no this-run log exists, return `originalErr`. Prefer also joining/retaining the original install error even when recovery succeeds, and add a regression test: pre-existing log + pre-npm failure → original error, no report written.

commands/curation/npmlogparser.go
```diff
@@ -0,0 +1,398 @@
...
+		return nil, nil, nil, nil
+	}
+
+	blockedPackages = map[string]npmBlockedInfo{}
+	// FIFO queue, not a single value: concurrent fetches can leave multiple notices pending at
+	// once. Best-effort, not a guarantee: notices and their matching 403 lines come from
+	// separate emit paths (a response-header handler vs. a request-completion logger), so
+	// nothing here proves same-relative-order holds under concurrency — if it doesn't, a 403
+	// pairs with the wrong pending notice and reports a confidently wrong policy/condition.
+	var pendingNotices []npmBlockedInfo
```
attiasas: **Important: FIFO notice→403 pairing can report the wrong policy under concurrency**

The comment correctly admits this is best-effort: notices and 403s come from separate emit paths, so same-relative-order isn't guaranteed. When `len(pendingNotices) > 1`, dequeuing `pendingNotices[0]` can attribute package A's policy/condition to package B — reported confidently as fact.

I understand the tradeoff vs degrading to "policy unknown" (breaks the concurrent whole-package-block happy path). Still, for merge I'd prefer degrading when the queue depth is >1 (or when pairing is otherwise ambiguous): mark blocked with unknown/empty policy rather than a guessed pair. Wrong remediation guidance is worse than incomplete guidance here.

commands/curation/curationaudit.go
```diff
@@ -106,10 +108,19 @@ const (
 
 	// hfUnresolvedReportKey is used when an HF scan found only dynamic references — no table, just warnings.
 	hfUnresolvedReportKey = "huggingface (unresolved references)"
+
+	// npmLogPartialReportWarning is shown when npm install crashed mid-resolution (e.g. a
+	// fully-blocked package's packument returned 403) and runNpmLogFallback recovered a
+	// partial report from npm's own debug log instead of aborting with no output.
+	npmLogPartialReportWarning = "npm install could not fully resolve the dependency tree because one or more packages " +
```
attiasas, line 116: **Important: partial-report warning always blames curation**

`npmLogPartialReportWarning` states install failed "because one or more packages are blocked by the curation policy." Real crashes often include ETARGET / git / non-registry failures; curation blocks may only be part of the picture (or absent from the recoverable set).

This misleads operators about root cause. Soften to something like: install failed mid-resolution; report reconstructed from the debug log and may be incomplete — optionally append/surface the underlying npm error.

commands/curation/curationaudit.go
```diff
@@ -1246,7 +1331,7 @@ func printResult(format outFormat.OutputFormat, projectPath string, packagesStat
 	if format == "" {
 		format = outFormat.Table
 	}
-	log.Output(fmt.Sprintf("Found %v blocked packages for project %s", len(packagesStatus), projectPath))
+	log.Output(fmt.Sprintf("Found %v blocked packages for project %s", countBlockedPackages(packagesStatus), projectPath))
```
attiasas, line 1344: **Important: not-evaluated rows still render inside the blocked-packages output**

`countBlockedPackages` correctly excludes `Action != blocked` for the "Found N blocked packages" header — thanks. But `printResult` still passes the full `packagesStatus` slice to JSON and `convertToPackageStatusTable`, so `nonRegistrySpecifierStatus` rows (`Action: notEvaluated`) appear under the blocked-packages table/JSON with no Action column in table mode.

Users can still treat non-registry deps as blocks despite the corrected count. Please either filter `Action != blocked` out of table/JSON, or render them in a clearly separate section/column.

commands/curation/curationaudit.go
```diff
@@ -987,8 +1061,19 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
 		if (tech == techutils.Pip || tech == techutils.Poetry || tech == techutils.Pipenv) && errors.As(err, &cvsErr) {
 			return ca.runCvsFallback(cvsErr, tech, results)
 		}
+		// npm's install error is a plain opaque error, nothing to type-match on — trigger the
+		// debug-log fallback unconditionally, unless the config lookup above already failed.
```
attiasas, line 1075: **Important: every opaque npm tree-build error enters fallback**

Unlike Python's typed `CvsBlockedError`, npm has nothing to type-match on, so *any* install failure triggers debug-log recovery. Combined with unbound newest-log selection, that expands the false-recovery surface well beyond curation crashes.

After binding logs to this run, also require recoverable curation signals (notice/403 and/or placeDep evidence from *this* log) before claiming success; otherwise return the original error.

commands/curation/curationaudit.go
```diff
@@ -1862,6 +1947,237 @@ func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, t
...
+			seenRangeWarnings[rangeKey] = true
+			advisoryWarnings = append(advisoryWarnings, fmt.Sprintf(
+				"Package '%s' (requested as '%s') could not be resolved to a specific version, so its curation status could not be checked. This is not necessarily a curation block — verify the dependency resolves correctly.",
+				entry.Name, entry.Specifier))
+			continue
+		}
+
+		// effectiveGraphVersion avoids collapsing distinct ETARGET/non-registry specifiers onto one key.
+		key, keyErr := npmPackageKey(tech, analyzer.url, analyzer.repo, entry.Name, effectiveGraphVersion(entry, blockedPackages))
+		if keyErr != nil || seenKeys[key] {
```
attiasas: **Important: keyErr is silently skipped**

`keyErr != nil` and `seenKeys[key]` share one `continue` with no advisory. A blocked package can vanish from the report with no warning — same class of problem the probe-error advisories already handle below.

Please split the cases: on `keyErr`, append an advisory (same pattern as probe failures); only silent-continue on true dedup via `seenKeys`.

commands/curation/curationaudit.go
```diff
@@ -479,20 +502,30 @@ func (ca *CurationAuditCommand) Run() (err error) {
...
+		if isWarningsOnlyReport(packagesStatus) && !packagesStatus.hfPartial && !packagesStatus.npmLogPartial {
 			continue
 		}
 		var partialReason string
 		switch {
+		// npmLogPartial is checked before isPartial: an npm-log-fallback report sets both
+		// (isPartial to reuse the existing waiver-skip gate below), so npmLogPartial must
+		// win here or it would incorrectly report "cvs_fallback".
+		case packagesStatus.npmLogPartial:
+			partialReason = "npm_log_fallback"
```
gauriy-tech, line 516: Confirmed via the API — line 515 is exactly partialReason = "npm_log_fallback". This is a clear SAST false positive, not a real credential.

partialReason is a plain categorical tag string used only to populate a JSON summary field describing why a report is partial — it sits alongside sibling values in the very same switch:

switch {
case packagesStatus.npmLogPartial:
    partialReason = "npm_log_fallback"
case packagesStatus.isPartial:
    partialReason = "cvs_fallback"
case packagesStatus.hfPartial:
    partialReason = "hf_unresolved"
}
None of these are usernames, passwords, API keys, or secrets — they're internal enum-like labels, never sent anywhere as an authentication credential, never derived from or resembling a real secret. The go-hardcoded-credentials rule is almost certainly pattern-matching on the shape "string literal assigned inside a case branch" without any actual semantic check on the variable name or value — partialReason doesn't even contain a credential-suggestive substring like "password," "token," or "key."

This is a false positive worth dismissing on the PR (reply/resolve) rather than acting on — there's no code change to make here; changing it to anything else (e.g., an env var, as the rule's generic remediation suggests) would be nonsensical for a value that was never a secret to begin with.

attiasas (approved):
LGTM

PR #739, opened 2026-04-20
Fast sast mode using changed files

attiasas:
## `feat(sast): add changed-files mode support for target-scoped SAST runs`

depends on:
* https://github.com/jfrog/jfrog-client-go/pull/1337

## Summary
This branch introduces SAST changed-files mode wiring so SAST can run only on changed files relevant to each target and skip when no files are in scope. It also expands SAST unit tests for changed-files path handling, refreshes dependencies, and adjusts test assets/expectations.

## Changes
- Added changed-files mode propagation through audit flow into JAS runner (`commands/audit/audit.go`, `commands/audit/auditparams.go`, `jas/runner/jasrunner.go`).
- Refactored SAST scan entrypoint to use a `SastScanParams` struct and pass changed-files metadata (`jas/sast/sastscanner.go`).
- Implemented changed-file normalization/filtering for per-target SAST roots, including safety checks, deduplication, and debug logging (`jas/sast/sastscanner.go`).
- Updated SAST config generation so roots are driven by changed files when mode is enabled (`jas/sast/sastscanner.go`).
- Added/updated tests for changed-files behavior and config roots, plus minor test expectation updates (`jas/sast/sastscanner_test.go`, `scans_test.go`).
- Removed PNPM and UV test data and corresponding CycloneDX audit tests (`tests/testdata/projects/package-managers/...`, `audit_test.go`).
- Updated `jfrog-cli-artifactory`, `jfrog-cli-core`, and `jfrog-client-go` versions (`go.mod`, `go.sum`).

## Testing
- Test coverage changes are included in `jas/sast/sastscanner_test.go` and `scans_test.go`.

PR #670, opened 2026-02-12
Support all attributes in Central Configs

attiasas:
## `feat(audit): config-profile scan targets, include/exclude, and xray-lib working dirs`

**Depends on:** 
* [XRAY-135682](https://jfrog-int.atlassian.net/browse/XRAY-135682) (bug: secrets/CA when multiple roots are passed)
* [XRAY-138915](https://jfrog-int.atlassian.net/browse/XRAY-138915) (improvement: iac does not support multiple roots)

* https://github.com/jfrog/jfrog-client-go/pull/1344

**Analyzer-Manager minimum version**: `1.33.0`
* https://github.com/jfrog/jfrog-cli-security/pull/748

## Summary
Aligns audit, git audit, SCA (Xray-Lib), and JAS scanners with JFrog Platform config profiles: per-target include/exclude patterns, scanner enablement, secret validation, partial results, and custom analyzer download paths. Passes resolved include directories and exclusions into Xray-Lib instead of global ignore patterns, and refactors scan execution around an enriched `ScanTarget` model.

## Changes
- **Audit / scan orchestration** (`commands/audit/`, `commands/scan/`, `commands/git/audit/`): target discovery via `GetTargetsInfo`; populate `ScanTarget` with include/exclude and central-config modules; honor config-profile scanner toggles and `allow_partial_results`; wire custom releases repo for Xray-Lib plugin download.
- **Xray-Lib BOM** (`sca/bom/xrayplugin/`): pass `IncludeDirs` and per-target SCA exclusions to the plugin; remove `WithTotalTargets` / `WithIgnorePatterns`; support `WithCentralRemoteReleasesDetails` for profile-driven dependency downloads.
- **JAS scanners** (`jas/`, `jas/runner/`): run against `results.ScanTarget` (include roots, profile exclusions); consolidate config-profile skip logic; fix SARIF invocation working directory to reflect scan target, not analyzer-manager temp dir; per-target secret validation via `IsSecretValidationActive`.
- **Results model** (`utils/results/`): `ScanTarget` gains `Include`, `Exclude`, `Technologies[]`, `CentralConfigModules`; skippable `GeneralErrors` / `TargetErrors` for partial results; helpers for central-config scan requests and exclusions; expanded unit tests.
- **Path / exclusion utilities** (`utils/paths.go`, `utils/techutils.go`): pattern handling for include/exclude during recursive scans.
[...]

PR #831, opened 2026-08-05
XRAY-157176 - Honor SAST changed files mode in the config profile scan flow

Jordanh1996:
## What

`enable_differential_scanning` (SAST fast diff mode) has been a no-op in the config-profile / Frogbot V3 flow. The flag reaches `SastScanManager` correctly, but the config file handed to the analyzer manager is built with the **full scan target as roots**, so the whole branch tree is analyzed on every PR scan.

`createConfigFileForTarget` never read `changedFilesMode` — only `deprecatedCreateConfigFile` did. Since the V3 flow leaves `DeprecatedAppsConfigModule` nil (it is only populated by `detectScaTargetsFromTechnologies`, the old flow), every config-profile scan took the builder that ignores the mode.

Introduced in #670, which split the single `createConfigFile` into a deprecated and a target-based copy and carried the changed-files branch over to the deprecated one only. The feature itself landed just before it in #739 / #743 / #745.

## Effect on a real customer (XRAY-157176, blocker)

Two runs on the same PR (one changed `.java` file), differential on vs off, produced a **byte-identical** SAST scanner input:

```yaml
scans:
    - roots:
        - /tmp/jfrog.cli.temp.-1785224331-234849826   # whole checkout, both runs
      type: sast
```

2 x ~17m SAST with the flag on vs 2 x ~18m with it off — a ~5% delta that is just noise. What made this hard to spot from logs is that the analyzer manager reports `running diff scan mode; first_scan/second_scan` in **both** cases: Frogbot PR scans always pass `target-result-file` for the post-scan sarif comparison, independently of the profile flag.

## Fix

Both builders now resolve roots through one `getScanRoots`, so the two paths cannot drift apart again. Behavior in changed-files mode is unchanged from the deprecated path: roots become the changed files, and `RunSastScan` still skips the scanner entirely when nothing changed under the target.

## Verified end to end

Reproduction, with both runtimes wired to the same JPD, config profile and PR:
**https://github.com/Jordanh1996/frogbot-sast-differential-repro** ([PR #1](https://github.com/Jordanh1996/frogbot-sast-differential-repro/pull/1))

Same one-file PR, same profile (`enable_sast_scan: true, enable_differential_scanning: true`), 150-class Java corpus:

| | `jfrog/frogbot@v3` (released) | Frogbot v3 built with this branch |
|---|---|---|
| `SAST changed files mode` log | absent | `using 1 paths as scan roots` |
| Scan roots | whole checkout | `.../ReportBuilder.java` |
| Files parsed per scan | `Generated IR for 151 files` | `Generated IR for 1 files` |
| SAST duration per scan | 25.9s / 25.2s | 18.2s / 18.4s |
| New issues reported on the PR | 1 | 1 |

Identical PR verdict for 1/151 of the parsing. On a customer-scale corpus the released runtime took 23m13s where the patched one took 2m24s.

Worth noting for the SAST team: with per-file roots the engine does **not** do full-repository preprocessing — it reports `Generated IR for 1 files` and the fixed scanner startup (~16s) then dominates. So the scoping does what it was designed to do.

## Tests

- `TestCreateConfigFileForTarget_ChangedFilesModeRoots` covers the previously untested target-based builder, including that changed-files mode with nothing changed yields no roots rather than falling back to the whole tree. It fails on `dev` and passes here.
- `TestSastChangedFilesMode` covers the other half of the chain — that a config profile with `enable_differential_scanning` actually turns changed-files mode on. That link was correct but untested, which is why the drop downstream went unnoticed. The derivation moved into `JasRunnerParams.sastChangedFilesMode()` to make it testable, and it now guards against a profile with no modules instead of indexing `Modules[0]` blindly.
- `TestCreateConfigFile_ChangedFilesModeRoots` drives the changed files through the manager field instead of the removed parameter, and both tests share one `readConfigRoots` helper.

## Out of scope, filed separately

`createConfigFileForTarget` also omits `ExcludedRules`, and nothing reads `SastScannerConfig.ExcludeRules` from the config profile at all — so SAST exclude-rules set in Centralized Config are silently ignored. Unlike the bug fixed here that one is not a #670 regression: the `excluded-rules` key has only ever been fed from a `jfrog-apps-config` module, so the profile field appears to have never been wired. Tracked separately in [XRAY-157301](https://jfrog-int.atlassian.net/browse/XRAY-157301) to keep this PR scoped to the blocker.

---

- [ ] Updated the Contributing page / ReadMe page / CI Workflow files if needed.

### DIFF
--- a/jas/sast/sastscanner_test.go
+++ b/jas/sast/sastscanner_test.go
@@ -296,6 +296,20 @@ func TestSastChangedFilesForTarget(t *testing.T) {
 	}
 }
 
+func readConfigRoots(t *testing.T, configFileName string) []string {
+	t.Helper()
+	data, err := os.ReadFile(configFileName)
+	require.NoError(t, err)
+	var cfg struct {
+		Scans []struct {
+			Roots []string `yaml:"roots,omitempty"`
+		} `yaml:"scans,omitempty"`
+	}
+	require.NoError(t, yaml.Unmarshal(data, &cfg))
+	require.Len(t, cfg.Scans, 1)
+	return cfg.Scans[0].Roots
+}
+
 func TestCreateConfigFile_ChangedFilesModeRoots(t *testing.T) {
 	scanner, cleanUp := jas.InitJasTest(t)
 	defer cleanUp()
@@ -318,21 +332,6 @@ func TestCreateConfigFile_ChangedFilesModeRoots(t *testing.T) {
 	changed := []string{"src/a.go", "src/b.go"}
 	ssm, err := newSastScanManager(scanner, scannerTempDir, false, false, "", nil)
 	require.NoError(t, err)
-
-	type yamlCfg struct {
-		Scans []struct {
-			Roots []string `yaml:"roots,omitempty"`
-		} `yaml:"scans,omitempty"`
-	}
-	readConfigRoots := func(t *testing.T) []string {
-		t.Helper()
-		data, err := os.ReadFile(ssm.configFileName)
-		require.NoError(t, err)
-		var cfg yamlCfg
-		require.NoError(t, yaml.Unmarshal(data, &cfg))
-		require.Len(t, cfg.Scans, 1)
-		return cfg.Scans[0].Roots
-	}
 
 	for _, tc := range []struct {
 		name             string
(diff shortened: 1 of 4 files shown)


jas/sast/sastscanner_test.go
attiasas: env_true_… and env_1_… are identical after env parsing went away. Collapse or rename for clarity.

attiasas (approved):
LGTM, see my comment

PR #832, opened 2026-08-05
XRAY-157301 - Apply the SAST exclude rules users configure

Jordanh1996:
https://jfrog-int.atlassian.net/browse/XRAY-157301

# Background

SAST exclude rules have never taken effect — neither those set in Centralized Config nor `excluded_rules` in `.jfrog/jfrog-apps-config.yml`.

# Description

We write `excluded-rules`; the scanner's `config_schema.json` declares `excluded_rules` and ignores unknown keys. The profile value was never read either — `createConfigFileForTarget` emitted no exclusions, `deprecatedCreateConfigFile` used only the apps-config module. Both builders now read them from the scan target via `ScanTarget.GetCentralConfigSastExcludeRules`, central config winning over the module (as `GetJasExcludePatternsForTarget` already does for patterns).

Needs [analyzer-manager#188](https://github.jfrog.info/JFROG/analyzer-manager/pull/188) released and `defaultAnalyzerManagerVersion` bumped, or the corrected key is dropped before the scanner.

# Tests

`TestExcludedRulesFromCentralConfig` over both builders. E2E: this branch in `jf` plus a CI build of #188 against the real scanner — excluding `java-stored-command-injection` in a profile gave 0 findings, without it 1.

---


jas/sast/sastscanner.go
```diff
@@ -201,6 +211,7 @@ func (ssm *SastScanManager) createConfigFileForTarget(target results.ScanTarget)
 				Roots:                  ssm.getScanRoots(jas.GetRootsFromTarget(target)),
 				Output:                 ssm.resultsFileName,
 				PathToResultsToCompare: ssm.resultsToCompareFileName,
+				ExcludedRules:          ssm.centralConfigExcludeRules,
```
attiasas: you can pass `target.GetCentralConfigSastExcludeRules()` here
similar to `target.GetCentralConfigExclusions(utils.SastScan)`

jas/sast/sastscanner.go
```diff
@@ -193,6 +195,14 @@ func (ssm *SastScanManager) getScanRoots(defaultRoots []string) []string {
 	return ssm.sastChangedFiles
 }
 
+// Rules excluded in the centralized config take precedence over the jfrog-apps-config module, as exclude patterns do.
+func (ssm *SastScanManager) getExcludedRules(moduleExcludedRules []string) []string {
+	if len(ssm.centralConfigExcludeRules) > 0 {
+		return ssm.centralConfigExcludeRules
+	}
+	return moduleExcludedRules
+}
```
attiasas, line 203: Do we also want to support in v2 exclude rules?

jas/sast/sastscanner.go
```diff
@@ -31,9 +31,10 @@ const (
 type SastScanManager struct {
 	scanner *jas.JasScanner
 
-	sastChangedFiles   []string
-	signedDescriptions bool
-	sastRules          string
+	sastChangedFiles          []string
+	signedDescriptions        bool
+	sastRules                 string
+	centralConfigExcludeRules []string
```
attiasas:
```suggestion
	sastRules                 string
	excludeRules []string
```

Should be generalized if it is passed to the manager here, usually for flags and not used for cenral config value so IDW if we should add it here if we don't add flag option to pass from CLI as well

attiasas (commented):
Nice work, check out my comments

attiasas (approved):
LGTM, don't forget that you need to update AM version with the sister commit before merging

PR #833, opened 2026-08-05
XRAY-156683 Native support for pip, pipenv and poetry(curation-audit)

Phavya-jfrog:

-----

Add native repository resolution for pip and poetry in `jf ca`, matching the existing pipenv pattern.

**Pip resolution priority:**
1. `.jfrog/projects/pip.yaml` (`jf pip-config`)
2. User `pip.conf` `index-url` — used only if (1) is absent
3. Clear error if neither resolves
Screenshots attached in https://jfrog-int.atlassian.net/browse/XRAY-156683?focusedCommentId=1184012

**Pipenv resolution priority** (no `jf pipenv-config`/yaml — always native):
1. User `pip.conf` `index-url` ("Set me up")
2. Artifactory `[[source]]` entry in `Pipfile` — used only if (1) is absent
3. Clear error if neither resolves
Screenshots attached in https://jfrog-int.atlassian.net/browse/XRAY-156683?focusedCommentId=1197088

**Poetry resolution priority** (no `jf poetry-config`/yaml — always native):
1. Artifactory `[[tool.poetry.source]]` entry in `pyproject.toml` — the only source ("Set me up")
2. Clear error if it doesn't resolve
Screenshots attached in https://jfrog-int.atlassian.net/browse/XRAY-156681?focusedCommentId=1197100

**`--run-native`** has no effect for pip, poetry and pipenv — resolution is already automatic from the priority above, so the flag is accepted as a no-op with a warning, same as pnpm/yarn.


commands/curation/curationaudit.go
```diff
@@ -1591,6 +1645,12 @@ func validateRunNativeForTech(tech techutils.Technology, runNative bool) error {
 				"Run 'jf ca' without --run-native; the repository is resolved automatically from " +
 				"'jf pipenv-config', ~/.pip/pip.conf, or the Artifactory [[source]] entry in your Pipfile.")
 	}
+	if tech == techutils.Pip {
+		return errorutils.CheckErrorf(
+			"--run-native is not supported for 'pip' projects. " +
+				"Run 'jf ca' without --run-native; the repository is resolved automatically from " +
```
gauriy-tech: I don't think we throw error for other package managers like pnpm , pipenv that dont support run-native, please check. It is just logged as no-op.

Phavya-jfrog: Implemented no-op behaviour for pipenv and pip

commands/curation/curationaudit.go
```diff
@@ -1693,10 +1714,112 @@ func (ca *CurationAuditCommand) setRepoFromPipfile() error {
...
+	}
+	if configuredRepo != "" && configuredServer != nil {
+		serverDetails := configuredServer
+		if configuredServer.User == "" && configuredServer.Password == "" && configuredServer.AccessToken == "" {
+			if base, sdErr := ca.ServerDetails(); sdErr == nil && base != nil {
+				copied := *base
+				copied.ArtifactoryUrl = configuredServer.ArtifactoryUrl
+				serverDetails = &copied
+			}
+		}
```
gauriy-tech, line 1730: When `pip.conf`'s `index-url` has no embedded credentials, this copies the configured `jf c` server's real credentials onto whatever host was discovered, with no check that the host (or scheme) matches. `setRepoFromUvToml` right below this (line 1997) already guards this exact case with `sameArtifactoryHost` before reusing `base`'s credentials — a stale or misconfigured local `pip.conf` pointing at a different host, or even a same-host `http://` entry, would send real Artifactory credentials to it.

```suggestion
		if configuredServer.User == "" && configuredServer.Password == "" && configuredServer.AccessToken == "" {
			base, sdErr := ca.ServerDetails()
			if sdErr != nil || base == nil {
				return errorutils.CheckErrorf("pip: no 'jf c' server configured to source credentials for %s", pipConfPath)
			}
			if !sameArtifactoryHost(base.ArtifactoryUrl, configuredServer.ArtifactoryUrl) {
				return errorutils.CheckErrorf(
					"pip: the Artifactory URL declared in %s (%s) does not match the configured 'jf c' "+
						"server (%s) — refusing to send its credentials to a different host. "+
						"Align the two, or select the matching server with --server-id.",
					pipConfPath, configuredServer.ArtifactoryUrl, base.ArtifactoryUrl)
			}
			copied := *base
			copied.ArtifactoryUrl = configuredServer.ArtifactoryUrl
			serverDetails = &copied
		}
```

---

commands/curation/curationaudit.go
```diff
@@ -1693,10 +1714,112 @@ func (ca *CurationAuditCommand) setRepoFromPipfile() error {
...
```
gauriy-tech: Same issue as the pip.conf path above: no host/scheme check before reusing the configured server's credentials for the host discovered in `pyproject.toml`'s `[[tool.poetry.source]]`. This one has a wider blast radius since `pyproject.toml` ships in the repo and can arrive via a fork PR's CI run.

```suggestion
		if configuredServer.User == "" && configuredServer.Password == "" && configuredServer.AccessToken == "" {
			base, sdErr := ca.ServerDetails()
			if sdErr != nil || base == nil {
				return errorutils.CheckErrorf("poetry: no 'jf c' server configured to source credentials for pyproject.toml [[tool.poetry.source]]")
			}
			if !sameArtifactoryHost(base.ArtifactoryUrl, configuredServer.ArtifactoryUrl) {
				return errorutils.CheckErrorf(
					"poetry: the Artifactory URL declared in pyproject.toml's [[tool.poetry.source]] (%s) does not match "+
						"the configured 'jf c' server (%s) — refusing to send its credentials to a different host. "+
						"Align the two, or select the matching server with --server-id.",
					configuredServer.ArtifactoryUrl, base.ArtifactoryUrl)
			}
			copied := *base
			copied.ArtifactoryUrl = configuredServer.ArtifactoryUrl
			serverDetails = &copied
		}
```

Given this same fallback-copy block is now duplicated three ways (uv, pip, poetry), it may be worth factoring into a shared helper once both guards are added.

---

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -1073,6 +1073,45 @@ func extractPoetrySourceNames(v any) []string {
...
+		if !ok {
+			continue
+		}
+		rawURL, _ := m["url"].(string)
+		if strings.TrimSpace(rawURL) == "" {
+			continue
+		}
+		if sd, repo := parseArtifactoryPypiURL(rawURL); sd != nil {
+			return sd, repo, nil
+		}
```
gauriy-tech: A `[[tool.poetry.source]]` URL with a username but no password (e.g. `https://user@host/...`) is accepted as-is here — `parseArtifactoryPypiURL` doesn't validate credential completeness. Its Pipfile sibling (`parsePipfileSource`) explicitly rejects this exact shape with "has incomplete URL credentials" before matching. Since the fallback-copy branch above only triggers when User/Password/AccessToken are *all* empty, a username-only URL skips that path and is used with an empty password — producing a confusing downstream 401 instead of a clear parse-time error.

```suggestion
		if sd, repo := parseArtifactoryPypiURL(rawURL); sd != nil {
			if sd.User != "" && sd.Password == "" {
				name, _ := m["name"].(string)
				return nil, "", fmt.Errorf("poetry: [[tool.poetry.source]] %q has incomplete URL credentials", name)
			}
			return sd, repo, nil
		}
```

---

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -1073,6 +1073,45 @@ func extractPoetrySourceNames(v any) []string {
...
+	v.SetConfigType("toml")
+	v.SetConfigFile(pyprojectToml)
+	if readErr := v.ReadInConfig(); readErr != nil {
+		return nil, "", errorutils.CheckErrorf("failed to read %s: %s", pyprojectToml, readErr)
+	}
+	arr, ok := v.Get("tool.poetry.source").([]any)
+	if !ok {
+		return nil, "", nil
+	}
+	for _, e := range arr {
```
gauriy-tech, line 1107: The doc comment above confirms "first Artifactory match wins" is intentional, but when `pyproject.toml` declares more than one Artifactory-shaped `[[tool.poetry.source]]` entry (e.g. a stale entry left from a prior migration, declared before the real one), this silently resolves against whichever comes first — including ignoring poetry's own `priority`/`default` markers. `pip.conf`'s duplicate `index-url` handling and the Pipfile resolver both reject this kind of ambiguity instead of silently picking one. Worth considering the same treatment here, so a leftover entry can't silently point curation-audit at the wrong repo.

---

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -1073,6 +1073,45 @@ func extractPoetrySourceNames(v any) []string {
 	return names
 }
 
+// ParsePyprojectArtifactorySource scans the current directory's pyproject.toml for a
+// [[tool.poetry.source]] entry whose url points at an Artifactory PyPI repository
+// (e.g. https://<host>/artifactory/api/pypi/<repo>/simple), returning its server + repo
+// name. Entries are checked in declaration order; the first Artifactory match wins.
+// Returns (nil, "", nil) when pyproject.toml is missing or has no such entry.
+func ParsePyprojectArtifactorySource() (serverDetails *config.ServerDetails, repoName string, err error) {
```
gauriy-tech, line 1083: `ParsePyprojectArtifactorySource` has no test that isolates its own parsing behavior (multi-source ordering, malformed entries, incomplete credentials) — its Pipfile and pip.conf analogs each have several dedicated tests for exactly this. It's exercised indirectly through `setRepoFromPyproject`'s tests, but those only cover the same-host success/fallback paths. Worth adding parser-level tests for the not-found, multiple-source, and incomplete-credential cases before merging.

---

gauriy-tech:
**Design Twice observation:** `setRepoFromPipConf` and `setRepoFromPyproject` share an identical 3-branch structure, and their credential-fallback blocks (the ones flagged above) are byte-for-byte identical. Once both add the `sameArtifactoryHost` guard, that block becomes identical to `setRepoFromUvToml`'s equivalent too — the same logic would then exist three times.

Worth considering a shared helper, e.g.:

```go
func (ca *CurationAuditCommand) resolveWithHostGuard(discovered *config.ServerDetails, sourceLabel string) (*config.ServerDetails, error) {
    if discovered.User != "" || discovered.Password != "" || discovered.AccessToken != "" {
        return discovered, nil
    }
    base, err := ca.ServerDetails()
    if err != nil || base == nil {
        return nil, errorutils.CheckErrorf("no 'jf c' server configured to source credentials for %s", sourceLabel)
    }
    if !sameArtifactoryHost(base.ArtifactoryUrl, discovered.ArtifactoryUrl) {
        return nil, errorutils.CheckErrorf(
            "the Artifactory URL declared in %s (%s) does not match the configured 'jf c' server (%s) — "+
                "refusing to send its credentials to a different host. Align the two, or select the matching server with --server-id.",
            sourceLabel, discovered.ArtifactoryUrl, base.ArtifactoryUrl)
    }
    copied := *base
    copied.ArtifactoryUrl = discovered.ArtifactoryUrl
    return &copied, nil
}
```

called from `setRepoFromPipConf`, `setRepoFromPyproject`, and (optionally, as a follow-up) `setRepoFromUvToml`, so the host-check fix only needs to exist and be tested once. Not blocking — flagging for your judgment on whether it's worth the refactor now vs. a follow-up.

---

sca/bom/buildinfo/technologies/python/python.go
```diff
@@ -1073,6 +1073,68 @@ func extractPoetrySourceNames(v any) []string {
...
+			continue
+		}
+		sd, repo := parseArtifactoryPypiURL(rawURL)
+		if sd == nil {
+			continue
+		}
+		if sd.User != "" && sd.Password == "" {
+			return nil, "", fmt.Errorf("poetry: [[tool.poetry.source]] %q has incomplete URL credentials", name)
+		}
+		matches = append(matches, poetrySourceMatch{name: name, sd: sd, repo: repo})
```
gauriy-tech, line 1124: Good fix on the username-without-password case, but the check is one-directional — a password-without-username URL (e.g. `https://:token@host/...`, which Go's `net/url` parses as a valid but incomplete credential) still slips through unrejected.

This file already has the exact symmetric check needed: `serverCredentialState()` (used by `mergePipenvCredentials` for the same purpose) returns `partialCredentials` whenever exactly one of user/password is set. Reusing it here avoids adding a third near-duplicate completeness check to the file.

```suggestion
		sd, repo := parseArtifactoryPypiURL(rawURL)
		if sd == nil {
			continue
		}
		if serverCredentialState(sd) == partialCredentials {
			return nil, "", fmt.Errorf("poetry: [[tool.poetry.source]] %q has incomplete URL credentials", name)
		}
		matches = append(matches, poetrySourceMatch{name: name, sd: sd, repo: repo})
```

---

gauriy-tech:
**[Re: sca/bom/buildinfo/technologies/python/python.go:1763-1769]** (`ParsePipConfigIndexUrl` — not part of this PR's diff, so this can't be attached as an inline comment, but it's a new consumer of this function via `setRepoFromPipConf`)

Unlike the poetry parser (which now at least catches one direction) and the Pipfile parser (which is fully symmetric), `ParsePipConfigIndexUrl` has no credential-completeness check at all — a pip.conf `index-url` with a username but no password (or vice versa) is returned as-is. Since this feeds `setRepoFromPipConf`'s fallback gate (which only triggers when user/password/token are *all* empty), an incomplete embedded credential skips the host-guarded fallback entirely and is used verbatim.

Same fix as the poetry case — reuse `serverCredentialState()`:

```go
	sd, repo := parseArtifactoryPypiURL(rawURL)
	if sd == nil {
		parsed, parseErr := url.Parse(rawURL)
		if parseErr != nil || parsed.Scheme == "" || parsed.Host == "" || strings.Contains(rawURL, "/api/pypi/") {
			return nil, "", "", fmt.Errorf("pip config \"%s\" [global] index-url is not a valid Artifactory PyPI URL", sourcePath)
		}
		return sd, repo, sourcePath, nil
	}
	if serverCredentialState(sd) == partialCredentials {
		return nil, "", "", fmt.Errorf("pip config %q [global] index-url has incomplete URL credentials", sourcePath)
	}
	return sd, repo, sourcePath, nil
```

---

gauriy-tech (approved):

commands/curation/curationaudit_test.go
```diff
@@ -4098,7 +4105,255 @@ url = "https://user:token@acme.jfrog.io/artifactory/api/pypi/repo/simple"
 
 	ca := NewCurationAuditCommand()
 	err := ca.setRepoFromPipfile()
+	require.NoError(t, err, "malformed pipenv.yaml must not break resolution — it's never read")
+	require.NotNil(t, ca.PackageManagerConfig)
+	assert.Equal(t, "repo", ca.PackageManagerConfig.TargetRepo(),
+		"must resolve from the Pipfile [[source]], not the stray pipenv.yaml")
+}
```
gauriy-tech, line 4112: This test proves the trivial case (malformed pipenv.yaml is ignored, no crash). It doesn't cover the case that actually matters for anyone who previously ran `jf pipenv-config`: a *well-formed* pipenv.yaml pointing at one repo, while pip.conf/Pipfile resolve to a different one. Worth adding a sibling test that locks in the intended behavior explicitly, so a future reader can see this was tested, not just assumed:

```suggestion
func TestSetRepoFromPipfile_IgnoresPipenvYaml(t *testing.T) {
	t.Setenv("HOME", t.TempDir())
	t.Setenv("PIP_CONFIG_FILE", "")
	t.Chdir(t.TempDir())
	require.NoError(t, os.MkdirAll(filepath.Join(".jfrog", "projects"), 0755))
	require.NoError(t, os.WriteFile(filepath.Join(".jfrog", "projects", "pipenv.yaml"), []byte("resolver: [\n"), 0600))
	require.NoError(t, os.WriteFile("Pipfile", []byte(`[[source]]
name = "jfrog"
url = "https://user:token@acme.jfrog.io/artifactory/api/pypi/repo/simple"
`), 0600))

	ca := NewCurationAuditCommand()
	err := ca.setRepoFromPipfile()
	require.NoError(t, err, "malformed pipenv.yaml must not break resolution — it's never read")
	require.NotNil(t, ca.PackageManagerConfig)
	assert.Equal(t, "repo", ca.PackageManagerConfig.TargetRepo(),
		"must resolve from the Pipfile [[source]], not the stray pipenv.yaml")
}

// TestSetRepoFromPipfile_ValidYamlDoesNotOverrideConflictingPipConf locks in the same
// ignore-pipenv.yaml behavior for a well-formed config, not just a malformed one: even a
// valid, correctly-configured pipenv.yaml (e.g. from a pre-existing 'jf pipenv-config' setup)
// is deliberately ignored in favor of the Pipfile [[source]] entry.
func TestSetRepoFromPipfile_ValidYamlDoesNotOverrideConflictingPipConf(t *testing.T) {
	t.Setenv("HOME", t.TempDir())
	t.Setenv("PIP_CONFIG_FILE", "")
	t.Chdir(t.TempDir())
	require.NoError(t, os.MkdirAll(filepath.Join(".jfrog", "projects"), 0755))
	require.NoError(t, os.WriteFile(filepath.Join(".jfrog", "projects", "pipenv.yaml"), []byte(`version: 1
type: pipenv
resolver:
    repo: yaml-configured-repo
    serverId: test
`), 0600))
	require.NoError(t, os.WriteFile("Pipfile", []byte(`[[source]]
url = "https://user:token@acme.jfrog.io/artifactory/api/pypi/pipfile-repo/simple"
`), 0600))

	ca := NewCurationAuditCommand()
	err := ca.setRepoFromPipfile()
	require.NoError(t, err)
	assert.Equal(t, "pipfile-repo", ca.PackageManagerConfig.TargetRepo(),
		"pipenv.yaml is deliberately ignored even when valid — resolution always comes from Pipfile/pip.conf")
}
```

---

PR #543, opened 2025-08-26
Static sca auto upload and fetch violations

attiasas:
- [ ] The pull request is targeting the `dev` branch.
- [ ] The code has been validated to compile successfully by running `go vet ./...`.
- [ ] The code has been formatted properly using `go fmt ./...`.
- [ ] All [static analysis checks](https://github.com/jfrog/jfrog-cli-security/actions/workflows/analysis.yml) passed.
- [ ] All [tests](https://github.com/jfrog/jfrog-cli-security/actions/workflows/test.yml) have passed. If this feature is not already covered by the tests, new tests have been added.
- [ ] Updated the [Contributing page](https://github.com/jfrog/jfrog-cli-security/blob/main/CONTRIBUTING.md) / [ReadMe page](https://github.com/jfrog/jfrog-cli-security/blob/main/README.md) / [CI Workflow files](https://github.com/jfrog/jfrog-cli-security/tree/main/.github/workflows) if needed.
- [ ] All changes are detailed at the description. if not already covered at [JFrog Documentation](https://github.com/jfrog/documentation), new documentation have been added.

-----

Depends on:
* https://github.com/jfrog/jfrog-client-go/pull/1186

`Static-SCA` Improvements:

1. Automatic cdx upload of `audit`/`git audit` scan results (you can control which repo by `--upload-rt-repo-path`),
A note is added with a link to the scan results that were uploaded.
2. Wait, fetch, and parse Xray violations using the new API call (generated by JPD for the cdx artifact)
3. upload-cdx: Wait for results to be parsed by Xray before showing the link.

Other Improvements:
1. Add `--sbom` option to `git-audit` cmd
2. `INFO` level Log, extra msg text. and links Improvements & Fixes. (add durations to scans)
3. Little code refactors to remove code duplications
4. Align BOM REF to reflect REF in Xray (`pkg:`/`file:` prefix, `/` as delimiter) 
5. Add `fail build` attribute to `simple-json` format.
6. Refactor convertors interface (remove redundant `target` input for each method)
7. dump final cdx format file to out-dir if requested
8. SARIF format - violations will now be displayed at a different run

PR #835, opened 2026-08-05
Fix CVE-less SCA violations dropped under a Watch

Jordanh1996:
## Problem

Any Xray SCA "Security" violation for a vulnerability that has no assigned CVE (only a native `XRAY-XXXXX` issue ID) is silently dropped when a Watch is used — with Frogbot V3 / `jf audit --static-sca`, the vulnerability never shows up in output, regardless of severity or policy match.

`convertToCveViolations` only builds output for entries in `violation.Cves`. Xray still returns a `CveDetails` entry for these vulnerabilities (CVSS/CWE populated), but with an empty `cve` field — which hits the `cve.Id == ""` guard and skips the violation entirely, with no fallback to the violation's own `IssueId`.

This is not a regression — the CVE-only assumption has been present since this converter was introduced (#543, 2025-11-23); CVE-less SCA vulnerabilities have never worked under a Watch.

## Fix

Fall back to `violation.IssueId` when no `Cves` entry has a usable ID, mirroring the same fallback already used when building the SBOM (`ExtractIssuesInfoForCdx`) — so the pattern is already precedented elsewhere in this codebase, just not applied here.

## Verification

Verified live against a real local Xray (not a unit test): a real Watch + Security policy (min severity Low) bound to a real indexed NuGet artifact carrying a real High-severity CVE-less vulnerability (`XRAY-1034038`, `OpenTelemetry.Resources.Host`).

Before this fix, calling the unmodified `GenerateViolations` against the real persisted violation:
```
[Info] Fetching violations from Xray...
[Warn] Skipping CVE violation with empty CVE ID for violation ID 2085062705793982464
=== SCA (CVE) violations returned: 0 ===
```

After the fix, same real Xray, same real violation:
```
=== SCA (CVE) violations returned: 1 ===
```
with correct severity (`High`), matched policy/watch, resolved impacted component, and fix version.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/policy/enforcer/policyenforcer.go
+++ b/policy/enforcer/policyenforcer.go
@@ -460,31 +460,43 @@ func convertToBasicViolation(violationType violationutils.ViolationIssueType, vi
 
 func convertToCveViolations(cmdResults *results.SecurityCommandResults, violation services.XrayViolation) (cveViolations []violationutils.CveViolation) {
 	resolved, unresolvedCount := resolveInfectedComponents(cmdResults, violation)
-	for _, cve := range violation.Cves {
-		if cve.Id == "" {
-			log.Warn(fmt.Sprintf("Skipping CVE violation with empty CVE ID for violation ID %s", violation.Id))
-			continue
-		}
+	for _, cveId := range getCveIds(violation) {
 		if len(resolved) == 0 {
 			logComponentLessFallback(violation, unresolvedCount)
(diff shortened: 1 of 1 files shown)


attiasas (commented):
This will have the CLI shows violation that there are no vulnerabilities related to them that returned from the scan

i.e if you run the same scan without watches it will not show the `XRAY-...` vulnerabilities
until Catalog will also return those

FYI

PR #837, opened 2026-08-06
XRAY-156919 - CVS fallback for yarn and pnpm

gauriy-tech:

-----

## Summary
Fixes CVS (curation) fallback reporting for pnpm and Yarn V2 when a curation-blocked package prevents a lockfile from being produced, and extracts the shared logic so both package managers use one implementation.

- Extracts Yarn's direct-dependency curation probe (HEAD/GET-checking each declared dependency's tarball) into a new shared package, `sca/bom/buildinfo/technologies/npm/curationprobe.go`, so pnpm can reuse it instead of duplicating it.
- Adds an equivalent no-lockfile fallback to pnpm: when `pnpm install --lockfile-only` aborts because curation blocked a package, `jf ca` now probes declared direct dependencies and reports which ones are blocked, matching Yarn's existing behavior.
- Removes Yarn V2's routing through `api/curation/audit` — that endpoint's redirect is broken for passing packages (missing the `/api/npm/` path segment), which crashed Yarn's JSON parser and aborted the whole install even for clean projects. Enforcement now goes through the direct-dependency probe plus the existing post-resolution HEAD-walker.
- Extends direct-dependency collection to workspace members (previously only the root `package.json` was checked).
- Relaxes pnpm's supported-version floor from "exactly 10" to "10 or newer" (tested through pnpm 12).
- Moves raw install-command output from user-facing error messages to debug-only logging, to avoid duplicate/noisy output.

## Testing
- New unit tests for the pnpm no-lockfile fallback, pnpm workspace-scoping, and the shared probe.
- New end-to-end test, `TestYarnV2CurationAudit`, covering the Yarn V2 fallback path.

-----


gauriy-tech:

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -759,7 +443,7 @@ func reconcileDeclaredDirectDepsAgainstTree(
 		if _, present := resolvedNames[name]; present {
 			continue
 		}
```
Phavya-jfrog: `resolvedNames` is keyed by package **name only** across the entire resolved tree, with zero logging when the skip fires. If a curation-blocked direct dependency shares a name with *any* already-resolved package — a workspace sibling's different version, or an ordinary transitive dependency — it's silently dropped: never synthesized, never HEAD-checked, never reported. This is a security-gating tool, so a full silent bypass here is high-impact.

The per-locator dup check a few lines below (`dependenciesMap[locator]`) already guards against re-inserting a real duplicate, so this name-only pre-check looks like the actual bug rather than a needed safeguard — removing it is the minimal fix (the now-unused `resolvedNames` construction above this hunk can be cleaned up too):

```suggestion
```

---

sca/bom/buildinfo/technologies/npm/curationprobe.go
```diff
@@ -0,0 +1,477 @@
...
+			ParentName:     dep.Name,
+			ParentVersion:  dep.ProbedVersion,
+			PackageName:    dep.Name,
+			PackageVersion: dep.ProbedVersion,
+			DepRelation:    "direct",
+			PkgType:        string(pkgType),
+		}
+		if len(dep.Policies) == 0 {
+			if dep.Reason == "not_found" {
+				row.BlockingReason = "Package not found in curation repository"
```
Phavya-jfrog: This "not found" wording diverges from the real curation walker's `BlockingReasonNotFound` ("Package pending update", `commands/curation/curationaudit.go:70`) — the sibling reasons just above (`"Policy violations"`, `"Blocked by curation (response could not be parsed)"`) do mirror their real counterparts, but this one doesn't. That breaks log correlation between this fallback path and the normal path for the identical condition. It's pre-existing (carried over from the old yarn.go), but this PR is what newly extends it to pnpm, so it's worth fixing while this code is being touched.

```suggestion
			if dep.Reason == "not_found" {
				// mirrors curation.BlockingReasonNotFound — import cycle prevents direct use
				row.BlockingReason = "Package pending update"
```

The same fix applies to the table-row equivalent a bit further down (`buildBlockedDirectDepsTableRows`, `"Package not found in curation repository"` → `"Package pending update"`).

---

sca/bom/buildinfo/technologies/npm/curationprobe.go
```diff
@@ -0,0 +1,477 @@
...
+		memberPI, err := biutils.ReadPackageInfoFromPackageJsonIfExists(wsDir, nil)
+		if err != nil || memberPI == nil {
+			continue
+		}
+		for n, v := range MergeDirectDeps(memberPI) {
+			declared[n] = v
+		}
+	}
+	return declared
+}
```
Phavya-jfrog, line 150: On a name conflict, the last-processed workspace member's version silently overwrites the previous one with no logging — the "last wins" behavior itself is intentional and covered by `TestCollectDeclaredDirectDepsAcrossWorkspaces`, but there's currently no trace when it actually happens. Since this feeds both yarn's and pnpm's total-install-failure fallback, a discarded version here means that package is never probed at all for either package manager.

```suggestion
func CollectDeclaredDirectDeps(curWd string, workspaceDirs []string) map[string]string {
	declared := map[string]string{}
	if rootPI, err := biutils.ReadPackageInfoFromPackageJsonIfExists(curWd, nil); err == nil && rootPI != nil {
		for n, v := range MergeDirectDeps(rootPI) {
			declared[n] = v
		}
	}
	for _, wsDir := range workspaceDirs {
		memberPI, err := biutils.ReadPackageInfoFromPackageJsonIfExists(wsDir, nil)
		if err != nil || memberPI == nil {
			continue
		}
		for n, v := range MergeDirectDeps(memberPI) {
			if existing, dup := declared[n]; dup && existing != v {
				log.Debug(fmt.Sprintf("curation: %s declared at both %s and %s across workspace members; using %s", n, existing, v, v))
			}
			declared[n] = v
		}
	}
	return declared
}
```

---

Phavya-jfrog:
**Missing test coverage** (from full review — no single line applies):

- `TestYarnV2CurationAudit` (`curation_test.go`) is self-flagged by its own comment as not run locally, needs a live platform — worth confirming green in CI before merge.
- No test yet exercises a name collision between a workspace member's declared dep and an already-resolved package elsewhere in the tree (the scenario in the review comment on `reconcileDeclaredDirectDepsAgainstTree`), nor the equivalent collision in `CollectDeclaredDirectDeps`. Both are straightforward to add once those fixes land.
- Pre-existing gap, not introduced by this PR: a non-curation install failure (e.g. a network error) is unconditionally framed as curation-caused once the probe finds no blocked package name — not blocking, but worth a follow-up ticket.

---

Phavya-jfrog (approved):

attiasas (approved):
LGTM, few comments:

* pnpm `resolveLockfileDir` funnel — non-install errors (temp dir / copy / FS) can still get curation framing and trigger the probe; yarn already gates more tightly

* Residual false diagnosis — empty-probe / non-curation failures can still read as “curation caused this”

* Last-wins on same package name across workspace members (can miss a blocked version)

* Stale comment on `TestYarnV2CurationAudit` claiming a live platform is required

* `^`/`~` treated as probeable concrete versions (known limitation)

PR #841, opened 2026-08-10
XRAY-157605 Use the real package name for aliased pnpm dependencies

Phavya-jfrog:

-----

pnpm aliases install a package under a different local name:

"strip-ansi-cjs": "npm:strip-ansi@^6.0.1"
Both pnpm tree builders used the map key — the alias — as the package name, producing a component ID for a package that exists in no registry.

The real name was already available in both paths and simply unused:

curation parses pnpm-lock.yaml, where the ref held the real name and it was discarded audit runs pnpm ls --json, which reports it as "from" — a field read nowhere in the repo. Both now resolve it into From, and the tree builders prefer it, so one change fixes both paths.

Before:
Screenshot 2026-08-10 at 12 36 44 PM

After:
Screenshot 2026-08-10 at 12 37 41 PM

### DIFF
--- a/sca/bom/buildinfo/technologies/pnpm/pnpm.go
+++ b/sca/bom/buildinfo/technologies/pnpm/pnpm.go
@@ -463,7 +463,7 @@ func parsePnpmLSContent(projectInfo []pnpmLsProject) (dependencyTrees []*xrayUti
 		// query Artifactory for a package that doesn't exist (404).
 		for name, dependency := range project.Dependencies {
 			if dependency.Local {
-				_ = uniqueDepsSet.Remove(getDependencyId(name, dependency.Version))
+				_ = uniqueDepsSet.Remove(getDependencyId(dependencyName(name, dependency), dependency.Version))
 			}
 		}
 	}
@@ -475,13 +475,13 @@ func createProjectDependenciesTree(project pnpmLsProject) map[string]xray.DepTre
 	treeMap := make(map[string]xray.DepTreeNode)
 	var directDependencies []string
 	for depName, dependency := range project.Dependencies {
-		directDependency := getDependencyId(depName, dependency.Version)
-		directDependencies = append(directDependencies, directDependency)
+		directDependency := getDependencyId(dependencyName(depName, dependency), dependency.Version)
+		directDependencies = appendUniqueChild(directDependencies, directDependency)
 		appendTransitiveDependencies(directDependency, dependency.Dependencies, &treeMap)
 	}
 	for depName, dependency := range project.DevDependencies {
-		directDependency := getDependencyId(depName, dependency.Version)
-		directDependencies = append(directDependencies, directDependency)
+		directDependency := getDependencyId(dependencyName(depName, dependency), dependency.Version)
+		directDependencies = appendUniqueChild(directDependencies, directDependency)
 		appendTransitiveDependencies(directDependency, dependency.Dependencies, &treeMap)
 	}
 	if len(directDependencies) > 0 {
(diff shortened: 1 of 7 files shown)


sca/bom/buildinfo/technologies/pnpm/pnpm_test.go
```diff
@@ -291,6 +291,66 @@ func TestParsePnpmLSContentNestsWorkspaceMembers(t *testing.T) {
 	assert.Contains(t, uniqueDeps, getDependencyId("cross-spawn", "7.0.6"))
 }
 
+// TestParsePnpmLSContentUsesRealNameForAliases asserts that an aliased dependency is
```
attiasas, line 295: `pnpmlock_test.go` / `pnpm_test.go` (new tests) — Coverage is split (lockfile asserts From; parsePnpmLSContent asserts IDs). There is no single `lockfile-YAML` → `parsePnpmLsContent` → `uniqueDeps `fixture proving the full curation pipeline. Low risk given the thin glue, but a small end-to-end case would catch a future From overwrite.

attiasas (approved):
LGTM

sca/bom/buildinfo/technologies/pnpm/pnpm.go
```diff
@@ -494,9 +494,22 @@ func getDependencyId(depName, version string) string {
...
+	}
+	return key
+}
+
 func appendTransitiveDependencies(parent string, dependencies map[string]pnpmLsDependency, result *map[string]xray.DepTreeNode) {
 	for depName, dependency := range dependencies {
-		dependencyId := getDependencyId(depName, dependency.Version)
+		dependencyId := getDependencyId(dependencyName(depName, dependency), dependency.Version)
 		if node, ok := (*result)[parent]; ok {
 			node.Children = append(node.Children, dependencyId)
```
gauriy-tech: **Duplicate sibling node when an alias collides with another dependency's real name+version (Medium)**

When two aliases resolve to the same real name+version (or an alias collides with a same-name/version direct dependency), the resulting duplicate string is appended twice into the same parent's `Children` here — and at lines 478/483 — with no dedup guard. Traced through to `populateXrayDependencyTree` (`utils/xray/xrayutils.go:49-61`): it iterates `Children` with no dedup either, so a duplicated ID becomes two identical sibling `GraphNode`s under the parent.

The `uniqueDeps` set stays correctly deduped (so vulnerability detection / the Artifactory HEAD-check aren't affected), and growth is bounded by `MaxUniqueAppearances`, but the tree itself ends up with a real duplicate-sibling defect.

Reproduces today:
```go
func TestParsePnpmLSContentDuplicateSiblingOnAliasCollision(t *testing.T) {
	project := pnpmLsProject{
		Name: "alias-collision-project", Version: "1.0.0",
		Dependencies: map[string]pnpmLsDependency{
			"strip-ansi-cjs": {From: "strip-ansi", Version: "6.0.1"},
			"strip-ansi-esm": {From: "strip-ansi", Version: "6.0.1"},
		},
	}
	trees, uniqueDeps := parsePnpmLSContent([]pnpmLsProject{project})
	require.Len(t, trees, 1)
	realID := getDependencyId("strip-ansi", "6.0.1")
	assert.Contains(t, uniqueDeps, realID)
	var childIDs []string
	for _, c := range trees[0].Nodes {
		childIDs = append(childIDs, c.Id)
	}
	assert.Len(t, childIDs, 1, "expected 1 deduped sibling, got: %v", childIDs)
}
```
Fails against the current head with `expected 1 deduped sibling, got: [npm://strip-ansi:6.0.1 npm://strip-ansi:6.0.1]`.

This codebase already has the fix for this exact shape of problem — `appendUniqueChild` in `sca/bom/buildinfo/technologies/npm/npm.go:253-260` — used one package over when building npm's own `treeMap[parent].Children`:
```go
func appendUniqueChild(children []string, candidateDependency string) []string {
	for _, existingChild := range children {
		if existingChild == candidateDependency {
			return children
		}
	}
	return append(children, candidateDependency)
}
```
Suggest adding an equivalent helper in this package (or promoting it to a shared location both `npm.go` and `pnpm.go` import) and using it at the three raw-append sites (lines 478, 483, and here) instead of the plain `append`.

---
🤖 Posted by an automated PR review (xray-pr-review).

gauriy-tech:
**Missing Tests: no round-trip test through real (de)serialization for aliased dependencies**

The new alias tests are solid on their own, but none of them exercise the actual serialization boundary — a mismatch between the assumed struct shape and what pnpm actually emits wouldn't be caught by any of them today:

- `pnpm_test.go` (JSON / `pnpm ls --json` path) — `TestParsePnpmLSContentUsesRealNameForAliases` and `TestParsePnpmLSContentAliasedTransitive` construct `pnpmLsProject`/`pnpmLsDependency` structs directly in Go and pass them straight into `parsePnpmLSContent`, never through `json.Unmarshal`.
- `pnpmlock_test.go` (YAML / lockfile path) — `TestBuildDepsMapResolvesAliases` and `TestWalkSnapshotResolvesAliases` construct `pnpmLockDep`/`pnpmLockSnapshot` maps directly and call `buildDepsMap`/`walkSnapshot` directly, never through the YAML parser.

Since the fix hinges specifically on a field tag (`json:"from"`) and a lockfile ref format, both would silently keep passing even if that tag or ref-parsing were wrong — only a real `json.Unmarshal`/YAML-parse of aliased-dependency input would catch that class of regression.

Suggest adding one test per path that goes through the real boundary:
- JSON: `json.Unmarshal` a literal `pnpm ls --json`-shaped payload (e.g. `"strip-ansi-cjs": {"from": "strip-ansi", "version": "6.0.1"}`) into `[]pnpmLsProject`, then call `parsePnpmLSContent`.
- YAML: parse a literal `pnpm-lock.yaml` snippet containing an aliased dependency through the existing lockfile entry point, rather than hand-building the intermediate structs.

---
🤖 Posted by an automated PR review (xray-pr-review).

gauriy-tech:
Approved

gauriy-tech (approved):

PR #842, opened 2026-08-10
Fix incorrect WD in rollback mechanism for npm and pnpm

eranturgeman:

-----
This PR corrects the rollback mechanism for npm and pnpm. Before we attempt to even regenerate the lock file we cd to the descriptor's dir in withDescriptorWorkingDir, and we cd back only after the rollback attempt (if exists).
So far the code tried to write the backup content to the descriptors root-relative path when we are already in dir where the lock file exists, which caused an error (file not found).
The fix simply fixes the path so the backup will be written to a correct path, and adds a test that can catch this scenario


attiasas (approved):
LGTM

PR #846, opened 2026-08-13
XRAY-157755 - Fix insecure tls flag traversal for maven projects

gauriy-tech:
## What & why

`curation-audit`/`audit` spawn a real `mvn` subprocess to build the dependency tree. Two bugs made
this fail silently for customers behind a TLS-intercepting proxy:

- `--insecure-tls` never reached that subprocess — `DepTreeParams` had no field for it.
- The real Maven error was swallowed behind `log.Verbose` (above `DEBUG`); only `exit status 1` ever surfaced.

## Changes

- Thread `InsecureTls` through `DepTreeParams` → `MavenDepTreeManager` → the spawned `mvn` command.
- Append both Wagon and native-resolver TLS-bypass properties (`-Dmaven.wagon.http.ssl.*` and
  `-Daether.connector.https.securityMode=insecure`), since Maven 3.9+ ignores the former.
- Embed captured Maven output in the returned error, bounded to 8KB, rune-boundary-safe.

## Testing

New unit tests, offline via a fake `mvnw` script (see `mvn_test.go`). Also verified end-to-end
against a real Artifactory instance with `curation-audit` — no regressions.

Related: XRAY-157755

Screenshot 2026-08-14 at 4 02 05 PM

-----


sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -232,12 +244,31 @@ func (mdt *MavenDepTreeManager) RunMvnCmd(goals []string) (cmdOutput []byte, err
 		if msg := technologies.GetMsgToUserForCurationBlock(mdt.isCurationCmd, techutils.Maven, stringOutput); msg != "" {
 			err = fmt.Errorf("failed running command '%s %s'\n\n%s", execPath, strings.Join(goals, " "), msg)
 		} else {
-			err = fmt.Errorf("failed running command '%s %s': %s", execPath, strings.Join(goals, " "), err.Error())
+			errMsg := fmt.Sprintf("failed running command '%s %s': %s", execPath, strings.Join(goals, " "), err.Error())
+			if stringOutput != "" {
+				errMsg += "\n" + truncateForError(stringOutput)
```
Phavya-jfrog: Captured Maven output is now embedded verbatim into the returned error with no redaction. Previously this content only reached the log at `Verbose` level (rarely seen); now it flows into the error shown to the user (and anything that logs or forwards it). If the underlying `mvn`/wrapper process ever echoes a credential or token in its stdout/stderr on failure — a proxy-auth header, a token in a URL, a plugin diagnostic dump — it would leak here.

`sca/bom/buildinfo/technologies/uv/uv.go`'s `maskPassword` (lines 709-726, used at 574/674) already solves exactly this problem for the `uv` provider. Worth reusing or adapting that pattern on `stringOutput` before it's embedded, rather than introducing a third, inconsistent variant of "does this technology redact captured output" (uv: yes, gradle: no, and now maven: no).

---

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -232,12 +244,31 @@ func (mdt *MavenDepTreeManager) RunMvnCmd(goals []string) (cmdOutput []byte, err
...
+			}
+			err = errors.New(errMsg)
```
Phavya-jfrog: `errors.New(errMsg)` discards the underlying `*exec.ExitError`, so `errors.As`/`errors.Is` can't recover it from this error going forward. Checked every `errors.As` + `exec.ExitError` pairing in the repo — none touch this call chain today, so this isn't an active gap for any current caller, just a minor forward-looking observation. If exit-code introspection is ever needed here, wrapping with `%w` (e.g. `fmt.Errorf("%s: %w", errMsg, err)`) would preserve it instead of flattening to a string.

---

sca/bom/buildinfo/buildinfobom.go
```diff
@@ -274,6 +274,7 @@ func GetTechDependencyTree(params technologies.BuildInfoBomGeneratorParams, arti
 		depTreeResult.FullDepTrees, uniqDepsNodes, err = java.BuildDependencyTree(java.DepTreeParams{
 			Server:                        artifactoryServerDetails,
 			DepsRepo:                      params.DependenciesRepository,
+			InsecureTls:                   params.InsecureTls,
```
Phavya-jfrog: This is the only line wiring `params.InsecureTls` into `java.DepTreeParams.InsecureTls` for both Maven and Gradle, but no test asserts it. The new coverage in this PR (`TestNewMavenDepTreeManagerPreservesAllParams`) only covers the next hop downstream (`DepTreeParams` → `MavenDepTreeManager`), not this call site. Consider a small test on `GetTechDependencyTree` confirming the field actually reaches `DepTreeParams`.

---

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -232,12 +244,31 @@ func (mdt *MavenDepTreeManager) RunMvnCmd(goals []string) (cmdOutput []byte, err
 		if msg := technologies.GetMsgToUserForCurationBlock(mdt.isCurationCmd, techutils.Maven, stringOutput); msg != "" {
```
Phavya-jfrog, line 245: None of the new tests exercise this branch (the curation-block message path) together with the new truncated-output embedding added just below in the `else` branch — nothing confirms `GetMsgToUserForCurationBlock`'s message is unaffected by the surrounding change. Low priority since this branch's own logic is unchanged, but worth a quick test given the code around it just changed.

---

Phavya-jfrog:
**[Re: sca/bom/buildinfo/technologies/java/gradle.go:61-65]** Gradle shares `DepTreeParams`/`BuildDependencyTree` with Maven — both route through the same `switch tech { case techutils.Maven, techutils.Gradle: ... }` in `buildinfobom.go`, which now threads `InsecureTls` into the shared struct. But `gradleDepTreeManager` has no `insecureTls` field, and `execGradleDepTree` never reads `InsecureTls` or adds any TLS-bypass argument to the spawned `gradle`/`gradlew` process. A customer behind the same TLS-intercepting proxy running a Gradle project would hit the identical silent failure this PR fixes for Maven, using the same `--insecure-tls` flag they'd reasonably expect to just work for either build tool.

Out of scope here since the PR title explicitly scopes to Maven — flagging as a candidate for a follow-up ticket.

---

gauriy-tech:
@Phavya-jfrog There will be improvement ticket for gradle https://jfrog-int.atlassian.net/browse/XRAY-140549?focusedCommentId=1199336

sca/bom/buildinfo/technologies/java/mvn.go
```diff
@@ -220,24 +224,68 @@ func (mdt *MavenDepTreeManager) RunMvnCmd(goals []string) (cmdOutput []byte, err
...
+func maskCredentials(output string, server *config.ServerDetails) string {
+	if server == nil {
+		return output
+	}
+	username, password, err := server.GetAuthenticationCredentials()
+	if err != nil {
+		return output
+	}
+	if password != "" {
+		output = strings.ReplaceAll(output, password, "***")
```
Phavya-jfrog, line 267: `maskCredentials` doesn't cover percent-encoded credential forms — verified with a repro: a password like `p@ss!` appearing percent-encoded as `p%40ss%21` in captured output survives masking today. `uv.go`'s `maskPassword` (which this mirrors) explicitly handles both the raw and percent-encoded forms of the password.

Mitigating factor: this repo's own settings.xml writer never embeds credentials in a URL (always separate `<username>`/`<password>` elements), so the concrete trigger would have to be Maven's own wagon/aether transport echoing a userinfo-embedded URL on failure — plausible but not something I could confirm or rule out here.

If you do close this gap, note that `url.QueryEscape` isn't a guaranteed match for whatever encoding Maven's transport layer might actually use (unlike `uv.go`, which parses an actual URL object it constructed itself) — worth validating against real Maven-observed output rather than assuming byte-for-byte parity with the `uv.go` approach.

---

sca/bom/buildinfo/technologies/java/mvn_test.go
```diff
@@ -720,6 +724,170 @@ func TestNewMavenDepTreeManagerPreservesAllParams(t *testing.T) {
...
+	require.NoError(t, os.WriteFile("mvnw", []byte("#!/bin/sh\nexit 1\n"), 0644))
+
+	manager := NewMavenDepTreeManager(&DepTreeParams{UseWrapper: true}, Tree)
+	_, err = manager.RunMvnCmd([]string{"some-goal"})
+	require.Error(t, err)
+	assert.False(t, strings.HasSuffix(err.Error(), "\n"), "error must not have a dangling trailing newline when there is no captured output")
+}
+
+// TestRunMvnCmdMasksCredentialsInError: server credentials must never appear verbatim in the error.
+func TestRunMvnCmdMasksCredentialsInError(t *testing.T) {
```
Phavya-jfrog, line 844: No test locks in credential masking for an `AccessToken`-only `ServerDetails` (common in CI/service-account setups) — this test covers `User`+`Password` only. Verified the current code already redacts an access token correctly (`GetAuthenticationCredentials()` returns `AccessToken` as the "password" value), so this isn't a live bug — but it's exactly the property this fix guarantees, so worth a sibling test to lock it in:

```go
func TestRunMvnCmdMasksAccessTokenInError(t *testing.T) {
	tmpDir := t.TempDir()
	currentDir, err := os.Getwd()
	require.NoError(t, err)
	restoreDir := tests.ChangeDirWithCallback(t, currentDir, tmpDir)
	defer restoreDir()

	script := "#!/bin/sh\necho 'auth failed, token eyJhbGciOiJSUzI1NiJ9.super-secret-token-body.sig rejected' >&2\nexit 1\n"
	require.NoError(t, os.WriteFile("mvnw", []byte(script), 0644))

	server := &config.ServerDetails{AccessToken: "eyJhbGciOiJSUzI1NiJ9.super-secret-token-body.sig"}
	manager := NewMavenDepTreeManager(&DepTreeParams{UseWrapper: true, Server: server}, Tree)
	_, err = manager.RunMvnCmd([]string{"some-goal"})
	require.Error(t, err)
	assert.NotContains(t, err.Error(), "super-secret-token-body", "access token must never appear verbatim in the returned error")
	assert.Contains(t, err.Error(), "***")
}
```

---

Phavya-jfrog (approved):

attiasas (approved):
LGTM

PR #849, opened 2026-08-13
XRAY-157127 - Keep each package manager resolving against its own Artifactory repo

Jordanh1996:
https://jfrog-int.atlassian.net/browse/XRAY-157127

# Background
Polyglot CI audits with distinct Go/npm (and other) resolver configs intermittently resolved one technology against another's Artifactory repository, producing flaky 404s depending on which technology ran first in the same `jf audit` process.

# Description
`BuildInfoBomGenerator` reused one shared `params` across technologies; `SetResolutionRepoInParamsIfExists` wrote the first tech's repo into that field, so later techs skipped their `projects/<tech>.yaml`. Each tech now resolves through `resolveTechParams` against a per-call copy, leaving the generator params clean while `--deps-repo` / `IgnoreConfigFile` still short-circuit the lookup.

# Tests
Reproduced and verified locally before and after the fix.

-----

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sca/bom/buildinfo/buildinfobom.go
+++ b/sca/bom/buildinfo/buildinfobom.go
@@ -157,12 +157,12 @@ func (b *BuildInfoBomGenerator) buildDependencyTree(scan results.ScanTarget) (*D
 	hasAnyTree := false
 	for _, tech := range techs {
 		log.Debug(fmt.Sprintf("Generating '%s' dependency tree for '%s'...", tech.ToFormal(), scan.Target))
-		serverDetails, err := SetResolutionRepoInParamsIfExists(&b.params, tech)
+		techParams, serverDetails, err := b.resolveTechParams(tech)
 		if err != nil {
 			buildErr = errors.Join(buildErr, fmt.Errorf("failed to set resolution repo in params: %w", err))
 			continue
 		}
-		treeResult, err := GetTechDependencyTree(b.params, serverDetails, tech)
+		treeResult, err := GetTechDependencyTree(techParams, serverDetails, tech)
 		if err != nil {
 			buildErr = errors.Join(buildErr, fmt.Errorf("failed while building '%s' dependency tree: %w", tech, err))
 			continue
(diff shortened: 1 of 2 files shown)


sca/bom/buildinfo/buildinfobom_test.go
```diff
@@ -2,16 +2,116 @@ package buildinfo
...
+	}}))
+	projectRoot := t.TempDir()
+	originalCwd, err := os.Getwd()
+	require.NoError(t, err)
+	require.NoError(t, os.Chdir(projectRoot))
+	t.Cleanup(func() { require.NoError(t, os.Chdir(originalCwd)) })
+	return projectRoot
+}
+
+func TestResolveTechParamsIsolation(t *testing.T) {
```
eranturgeman: please rename test to be Test<name of function> as the name of the func should be explanatory for its actions

sca/bom/buildinfo/buildinfobom.go
```diff
@@ -183,6 +183,12 @@ func (b *BuildInfoBomGenerator) buildDependencyTree(scan results.ScanTarget) (*D
 	return merged, buildErr
 }
 
+func (b *BuildInfoBomGenerator) resolveTechParams(tech techutils.Technology) (techParams technologies.BuildInfoBomGeneratorParams, serverDetails *config.ServerDetails, err error) {
+	techParams = b.params
```
eranturgeman, line 187: the copy is shallow — ServerDetails is a pointer field, so it still points at the same shared object. Nuget's curation path (technologies/nuget/nuget.go:146) mutates that shared ServerDetails.ArtifactoryUrl in place, which can leak across techs in a polyglot curation run — the exact bug class this PR is meant to fix, just via a different field. 
pnpm.go already guards against this with an explicit copy-before-mutate (with a comment calling out the same hazard); nuget doesn't.

sca/bom/buildinfo/buildinfobom_test.go
```diff
@@ -2,16 +2,116 @@ package buildinfo
...
+
+	generator := NewBuildInfoBomGenerator()
+	_, _ = generator.buildDependencyTree(results.ScanTarget{
+		Target:       projectRoot,
+		Technologies: []techutils.Technology{techutils.Go},
+	})
+
+	assert.Empty(t, generator.params.DependenciesRepository)
+}
+
```
eranturgeman: if possible, try combining all those test funcs into one with testcases and explicit name for each test case. this is easier to maintain where we have 1to1 relation between unit test to a func (as much as possible)

eranturgeman (approved):
LGTM! see my comments

PR #854, opened 2026-08-19
XRAY-156682 - Yarn native support

Phavya-jfrog:

-----

Add native repository resolution for yarn v2 and v3 in jf ca.

Yarn v2, v3 resolution priority:
1. `npmRegistryServer` entry in `.yarnrc.yml` — the only source ("Set me up")
2. Clear error if it doesn't resolve

Screenshots attached in https://jfrog-int.atlassian.net/browse/XRAY-156682?focusedCommentId=1210543


sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -546,25 +545,24 @@ func resolveCurationLockfileDir(
...
-		return runYarnInstallAccordingToVersion(curWd, yarnExecPath, params.InstallCommandArgs, params.IsCurationCmd)
-	}
 
-	executableYarnVersion, err := bibuildutils.GetVersion(yarnExecPath, curWd)
-	if err != nil {
-		return err
+	// Skip credential injection when no repo was resolved, or for curation (native
+	// .yarnrc.yml resolution already has it, for V2/V3/V4 alike). Only non-curation
+	// V2/V3 with a repo from --deps-repo or 'jf yarn-config' still needs it below.
+	useNativeInstall := depsRepo == "" || params.IsCurationCmd
```
gauriy-tech, line 578: Curation's fallback credential (`credentialFallbackServerDetails`, computed a few lines above in `setRepoFromYarnrc`) never reaches the actual `yarn install` / `yarn jfrog-yarn-resolve-lockfile` subprocess — `params.IsCurationCmd` short-circuits `useNativeInstall` to `true` before the credential-injection branch (`GetYarnAuthDetails` + `ModifyYarnConfigurations`) can run, for V2/V3/V4 alike. Any project whose `.yarnrc.yml` carries no inline token and relies on the `jf c` server as fallback (a case this same PR added tests for) will now hit the registry anonymously and fail auth — previously working for V2/V3 via `jf yarn-config` injection.

Consider injecting the resolved token into the subprocess's env (e.g. `YARN_NPM_AUTH_TOKEN`) immediately before `runYarnInstallAccordingToVersion`/`runYarnResolveOnlyLockfile` when curation resolved a fallback credential, without touching `YARN_NPM_REGISTRY_SERVER` (which must keep coming from `.yarnrc.yml` per this PR's design).

---

gauriy-tech:
**[Re: sca/bom/buildinfo/technologies/yarn/yarn.go:298]** `curationNoLockfileError` unconditionally frames any pre-lockfile install failure as "curation is blocking manifests, not just tarballs," even when `installErr` is unrelated (e.g. a plugin-install permission error). Since `probeBlockedDirectDeps` authenticates independently via `params.ServerDetails`, an unrelated failure here (including the fallback-auth gap above) would still report zero blocked packages while the top-level message blames curation — pointing engineers toward removing/waiving packages that curation never actually evaluated.

Consider branching on whether `installErr` actually carries a curation-block signal before choosing this wording, falling back to a neutral "install failed, see debug log" message otherwise.

---

gauriy-tech:
**[Re: commands/curation/curationaudit.go:975]** This log line still says "(V4 native mode)" — after this PR, `.yarnrc.yml`-native resolution applies to V2/V3 too, so the label is now inconsistent with the rest of this change (not introduced by this PR, but adjacent to it). Consider dropping the version-specific qualifier, e.g. "...treating project as yarn (native .yarnrc.yml mode)."

---

gauriy-tech:
Missing test coverage for the two behavioral changes flagged above:

- No test exercises `configureYarnResolutionServerAndRunInstall` / `runYarnResolveOnlyLockfile` end-to-end with `IsCurationCmd: true` and a fallback-only credential (no token in `.yarnrc.yml`) to confirm the spawned yarn subprocess actually authenticates. `yarn_test.go` is untouched by this PR, and `TestSetRepoFromYarnrcAcceptsMatchingHost` (added here) only asserts `setRepoFromYarnrc` populates `ca.PackageManagerConfig` — it stops short of driving that credential through to the subprocess.
- No test on `curationNoLockfileError` covers a non-curation-shaped `installErr` (e.g. a plugin-install permission error) to confirm the message doesn't misattribute the failure to curation policy.

Both are reproducible with a fake `yarn` executable/script standing in for the real binary — happy to share a sketch if useful.

---

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -546,25 +571,38 @@ func resolveCurationLockfileDir(
...
+	if !useNativeInstall {
+		executableYarnVersion, versionErr := bibuildutils.GetVersion(yarnExecPath, curWd)
+		if versionErr != nil {
+			return versionErr
+		}
+		useNativeInstall = version.NewVersion(executableYarnVersion).Compare(YarnV4Version) <= 0
+	}
+	if useNativeInstall {
+		if params.IsCurationCmd && depsRepo != "" {
```
gauriy-tech: Nice fix for the previous credential-propagation gap — one follow-up: `injectCurationFallbackAuthEnv` now fires whenever `params.IsCurationCmd && depsRepo != ""`, with no way to distinguish "`.yarnrc.yml` already has its own token" from "no token, needs the `jf c` fallback" (`setRepoFromYarnrc` sets `ServerDetails.AccessToken` non-empty in both cases). That means every successful yarn curation run now makes an extra, non-optional `GetYarnAuthDetails` → `ValidateRepoExists` HTTP call to Artifactory — confirmed this with a mock server returning 500: an install that previously needed no such call, and would have succeeded purely from `.yarnrc.yml`'s own token, now fails.

Consider threading a flag (e.g. `YarnCredentialsFromFallback`, set in `setRepoFromYarnrc` based on `registryConfig.AuthToken == ""`) through to here, and only injecting when the credential actually came from the fallback.

---

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -296,13 +303,31 @@ func lockfileMtime(yarnLockPath string) time.Time {
...
 
+// installErrCarriesCurationBlockSignal reports whether installErr looks like a curation
+// block (HTTP 403), as opposed to an unrelated failure (e.g. an auth error). Yarn echoes
+// curation's HTTP response verbatim, e.g. "YN0035: ... Response Code: 403 (Forbidden)".
+func installErrCarriesCurationBlockSignal(installErr error) bool {
+	if installErr == nil {
+		return false
+	}
+	errText := strings.ToLower(installErr.Error())
+	return strings.Contains(errText, "403") || strings.Contains(errText, "forbidden")
```
gauriy-tech: The curation-block detection here is a bare substring match with no word boundary — `strings.Contains(errText, "403")` would also match an unrelated error whose text happens to contain "403" (e.g. inside a hostname or port), sending it down the curation-blaming message path this fix was meant to avoid.

```suggestion
	return strings.Contains(errText, "response code: 403") || strings.Contains(errText, "forbidden")
```

---

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -47,7 +48,14 @@ const (
...
-	yarnV2Version   = "2.0.0"
+	// Env vars yarn reads for npm auth, used to inject curation's fallback credential
+	// (see injectCurationFallbackAuthEnv) without touching YARN_NPM_REGISTRY_SERVER.
+	//#nosec G101
+	yarnNpmAuthIdentEnv = "YARN_NPM_AUTH_IDENT"
+	//#nosec G101
+	yarnNpmAuthTokenEnv  = "YARN_NPM_AUTH_TOKEN"
+	yarnNpmAlwaysAuthEnv = "YARN_NPM_ALWAYS_AUTH"
+	yarnV2Version        = "2.0.0"
 	yarnV3Version   = "3.0.0"
```
gauriy-tech: Minor: `yarnV3Version` didn't pick up the realignment when the longer `yarnNpmAuth*Env` names were added to this block — `gofmt -w` will fix it.

```suggestion
	yarnV3Version        = "3.0.0"
```

---

commands/curation/curationaudit_test.go
```diff
@@ -4078,6 +4079,169 @@ url = "http://configured-server.example.com/artifactory/api/pypi/uv-test-repo/si
 	assert.Nil(t, ca.PackageManagerConfig, "credentials must not be downgraded to a cleartext http URL on the same host")
 }
 
+// yarnBerryExecutableForTest returns a "yarn" wrapper that resolves via Corepack to the
+// Berry version pinned in the target project's package.json. This avoids relying on
+// whatever "yarn" happens to be on PATH (which may be Classic/V1). Skips if corepack is
+// missing.
+func yarnBerryExecutableForTest(t *testing.T) string {
```
gauriy-tech, line 4086: Low/informational: these Berry tests shell out to a real `corepack`, which does a live fetch to `registry.npmjs.org` to resolve the pinned yarn version — no proxy restriction is configured in this repo's `.github/workflows/test.yml`, so it should work fine on the actual CI runners, but it does add a live third-party network dependency to unit tests across all 3 OS legs (rate limits/registry hiccups could cause occasional flakes unrelated to the code under test). Not blocking — just flagging in case you'd rather vendor/pin the Berry binary or mock the corepack resolution down the line.

---

sca/bom/buildinfo/technologies/yarn/yarn.go
```diff
@@ -296,13 +303,31 @@ func lockfileMtime(yarnLockPath string) time.Time {
...
 
+// installErrCarriesCurationBlockSignal reports whether installErr looks like a curation
+// block (HTTP 403), as opposed to an unrelated failure (e.g. an auth error). Yarn echoes
+// curation's HTTP response verbatim, e.g. "YN0035: ... Response Code: 403 (Forbidden)".
+func installErrCarriesCurationBlockSignal(installErr error) bool {
+	if installErr == nil {
+		return false
+	}
+	errText := strings.ToLower(installErr.Error())
+	return strings.Contains(errText, "response code: 403") || strings.Contains(errText, "forbidden")
```
gauriy-tech: The `"response code: 403"` anchor is solid (empirically, Yarn always renders `"Response Code: 403 (Forbidden)"`, so it has full recall for real curation blocks on its own) — but the leftover `|| strings.Contains(errText, "forbidden")` still allows a false positive: an unrelated failure containing the word "forbidden" (e.g. a git-dependency clone rejected with `"remote: Forbidden"`) would still get classified as a curation block.

Since `"response code: 403"` alone already catches every real case, this disjunct doesn't add coverage — only false-positive risk.

```suggestion
	return strings.Contains(errText, "response code: 403")
```

---

gauriy-tech (approved):

PR #856, opened 2026-08-23
XRAY-145930 - Onboard cargo for jf ca

gauriy-tech:
## What
Onboards Cargo (Rust) support for `jf curation-audit`: resolves the project's dependency graph via `cargo generate-lockfile` against an isolated, curation-redirected `CARGO_HOME`, then reports it like every other supported package manager.

## Design notes
- **Isolation**: `jf ca` never touches the real project or its ambient Cargo config. Resolution runs in a temp copy (`.cargo` excluded from the copy) with `CARGO_HOME` redirected to a self-contained, always-freshly-written config pointing at Artifactory — so a project's own config, or the developer's global one, can never override the curation redirect.
- **Registry discovery**: there's no `jf cargo-config` command; the Artifactory URL/repo is read directly from the project's or the machine's existing `.cargo/config.toml`, matching what Artifactory's "Set Me Up" instructions already produce. A host/scheme match against the configured `jf c` server is enforced before any credentials are attached.
- **Workspaces**: a member of a Cargo workspace is resolved together with the rest of the workspace (for a consistent lock), but the reported dependency tree stays scoped to the directory actually audited — not the whole workspace and not the workspace root.
- **Lock reuse**: if a committed `Cargo.lock` already matches `Cargo.toml`'s declared dependencies (including a hand-rolled check that a caret/default version requirement is still satisfied by the locked version), `cargo generate-lockfile` is skipped. Any requirement shape that can't be verified this way (tilde, wildcard, comparison operators, path/git deps) always forces regeneration.

## Testing
Unit tests cover the registry-discovery strategies, the credential host-check, workspace scoping/dedup, the lock-reuse decision (including the caret-range matching, verified against Cargo's full documented semantics), and curation-isolation regressions (project-local config override, global ambient config override) — the latter two are integration-style tests that invoke the real `cargo` binary when available.

Screenshot 2026-08-24 at 10 27 19 AM

-----


sca/bom/buildinfo/technologies/cargo/cargo.go
```diff
@@ -0,0 +1,789 @@
...
+		if parent == dir || (home != "" && dir == home) {
+			return ""
+		}
+		dir = parent
+	}
+}
+
+// BuildDependencyTree uses 'cargo generate-lockfile', not 'cargo metadata', since metadata fails outright when any package is curation-blocked at download time.
+func BuildDependencyTree(params technologies.BuildInfoBomGeneratorParams) (dependencyTrees []*xrayUtils.GraphNode, uniqueDeps []string, err error) {
+	if !params.IsCurationCmd {
```
Phavya-jfrog, line 536: This guard combined with `Cargo.toml`/`Cargo.lock` being registered as generic auto-detection indicators (`utils/techutils/techutils.go:316`) means plain `jf audit`/`jf sbom` hard-fails on any target containing a Cargo project — including mixed-tech targets. A directory with both `Cargo.toml` and `go.mod` will have its Go tree fully resolved, but `buildDependencyTree`'s joined error still propagates to `GenerateSbom`, which discards the entire (valid) result rather than returning the partial tree.

Docker/HuggingFaceML avoid this by simply having no `indicators` field at all, so they're never auto-detected outside their own CLI flow. Consider the same approach for Cargo, or make the shared tree-building loop tolerate a per-tech failure when other techs succeeded.

---

commands/curation/curationaudit.go
```diff
@@ -1105,6 +1112,12 @@ func (ca *CurationAuditCommand) auditTree(tech techutils.Technology, results map
 			return err
 		}
 	}
+	// Cargo has no 'jf cargo-config' file; resolve its repo from .cargo/config.toml instead.
+	if tech == techutils.Cargo && ca.PackageManagerConfig == nil {
```
Phavya-jfrog: Unlike Uv/Pipenv/Pip/Poetry (which populate `PackageManagerConfig` via `GetAuth` in `supportedTech` before this point), Cargo's `supportedTech` entry always returns `true, nil` without ever calling `GetAuth`. So a non-nil `PackageManagerConfig` here can only be stale leftover state — e.g. from an earlier `--working-dirs` entry whose `doCurateAudit` error path returned without resetting it. That stale config silently gets reused, skipping this call (and the host/scheme check inside it) entirely.

The same `== nil` gate pattern exists for the other techs too and shares this root cause (no reset on the generic error path in `doCurateAudit`/`Run()`), but Cargo is unconditionally exposed on every run since it never populates the config through `GetAuth` first.

```suggestion
	// Cargo has no 'jf cargo-config' file, and never pre-populates PackageManagerConfig via GetAuth
	// (unlike Uv/Pipenv/Pip/Poetry) — so a non-nil value here can only be stale leftover state from
	// an earlier working-dir. Always resolve fresh from .cargo/config.toml.
	if tech == techutils.Cargo {
		if err := ca.setRepoFromCargoConfig(); err != nil {
			return err
		}
	}
```

---

sca/bom/buildinfo/technologies/cargo/cargo.go
```diff
@@ -0,0 +1,789 @@
...
+	}
+	// Always self-contained: the ambient config's own crates-io replacement is invisible once CARGO_HOME is redirected below.
+	sourceBlock := fmt.Sprintf("\n[source.%s-remote]\nregistry = %q\n\n[source.crates-io]\nreplace-with = %q\n",
+		cargoRegistryName, sparseUrl, cargoRegistryName+"-remote")
+
+	configContent := fmt.Sprintf(`[registry]
+default = %q
+global-credential-providers = ["cargo:token"]
+
+[registries.%s]
```
Phavya-jfrog, line 793: `registriesName` here (and again at line 755 below) comes from `AmbientRegistriesName` — a TOML table key parsed straight out of the audited project's own or the machine's ambient `.cargo/config.toml`. That's exactly the untrusted input this isolated, "always-freshly-written" `CARGO_HOME` config exists to not be influenced by, yet it's spliced in via unescaped `%s` into a TOML table header (`[registries.%s]`), while the same value is correctly quoted via `%q` two lines above for `default = %q`.

Any ordinary name with a space or other TOML-special character produces invalid TOML in the file that's supposed to be authoritative, breaking `cargo generate-lockfile` for that project. Recommend validating `registriesName` against Cargo's allowed bare-key charset (e.g. `^[A-Za-z0-9_-]+$`) and falling back to the canonical `cargoRegistryName` constant when it doesn't match, rather than reusing the raw ambient value unescaped.

---

sca/bom/buildinfo/technologies/cargo/cargo.go
```diff
@@ -0,0 +1,789 @@
...
+			return false
+		}
+		for depName, req := range declaredReqs {
+			edge, ok := lockedEdges[depName]
+			if !ok {
+				return false
+			}
+			trimmed := strings.TrimSpace(req)
+			resolved := resolveCargoDependency(byName, edge)
+			if exactVersion, ok := strings.CutPrefix(trimmed, "="); ok {
```
Phavya-jfrog, line 434: Cargo accepts a space after `=` in an exact-pin requirement (`"= 1.2.3"`). `CutPrefix` here doesn't re-trim the remainder, so `exactVersion` keeps the leading space and never equals `resolved[0].Version`. This forces an unnecessary (but harmless) `cargo generate-lockfile` regeneration for any dependency pinned this way, even when the checked-in lock already matches.

```suggestion
			if exactVersion, ok := strings.CutPrefix(trimmed, "="); ok {
				exactVersion = strings.TrimSpace(exactVersion)
				if len(resolved) != 1 || resolved[0].Version != exactVersion {
```

---

commands/curation/curationaudit.go
```diff
@@ -1405,19 +1436,18 @@ func (ca *CurationAuditCommand) sendWaiverRequests(pkgs []*PackageStatus, msg st
...
+		resp := &ErrorsResp{}
+		if err := json.Unmarshal(body, resp); err != nil {
 			return nil, fmt.Errorf("failed decoding waiver request status %v", err)
 		}
 
 		if len(resp.Errors) != 1 {
 			return nil, fmt.Errorf("got unexpected response structure while sending waiver request: %s", body)
 		}
+		// Cargo's curation-block body uses "detail" instead of "message" (see ErrorResp.Detail).
+		if resp.Errors[0].Message == "" && resp.Errors[0].Detail != "" {
```
Phavya-jfrog: No test exercises this fallback (same gap applies to the identical fallback in `getBlockedPackageDetails` around line 2825). Confirmed via mutation testing: removing these three lines still leaves the existing `TestSendWaiverRequests` suite fully passing, since none of its mock responses populate `detail` with `message` left empty.

Since this is the parsing path specifically added for Cargo's Artifactory error body shape, worth adding a subtest case (mock response with `"detail"` set and `"message"` omitted) to `TestSendWaiverRequests` and an equivalent case for `getBlockedPackageDetails`'s coverage.

---

sca/bom/buildinfo/technologies/cargo/cargo.go
```diff
@@ -0,0 +1,789 @@
...
+		err = tempErr
+		return
+	}
+	defer func() {
+		if rmErr := fileutils.RemoveTempDir(tempDir); rmErr != nil {
+			log.Warn(fmt.Sprintf("cargo: could not remove temp dir %s: %v", tempDir, rmErr))
+		}
+	}()
+	// Exclude .cargo: a project-shipped config.toml would otherwise take precedence over the isolated CARGO_HOME below.
+	if err = biutils.CopyDir(copyRoot, tempDir, true, []string{technologies.DotVsRepoSuffix, cargoTargetDirName, cargoConfigDirName}); err != nil {
```
Phavya-jfrog, line 586: Worth flagging: the vendored `build-info-go` `CopyDir` passes `nil` for the exclude list on its recursive calls, so this exclude list (`.cargo`, `target`, etc.) only filters the immediate children of `copyRoot`, not nested subdirectories. Not currently exploitable — `cargo generate-lockfile` always runs with `cmd.Dir` at the copy root, and Cargo's own config discovery only walks upward — but a future refactor that runs cargo from a workspace-member subdirectory would silently reintroduce the isolation bypass this exclude list is meant to prevent. A short comment here noting that assumption would help the next person who touches this.

---

sca/bom/buildinfo/technologies/cargo/cargo_test.go
```diff
@@ -0,0 +1,866 @@
+package cargo
```
Phavya-jfrog, line 1: All the tests in this file call `cargo.BuildDependencyTree` directly — none drive the real `techutils.DetectTechnologiesDescriptors → buildDependencyTree → GenerateSbom` chain with a Cargo-containing target. That's exactly the path where the `jf audit` regression flagged elsewhere in this review lives (a Cargo-only or mixed Cargo+Go target currently hard-fails `GenerateSbom` instead of skipping or partially succeeding). An integration-style test building a `ScanTarget` with `Cargo.toml` (and ideally `go.mod` alongside it) and asserting `GenerateSbom`'s behavior with `IsCurationCmd: false` would catch this class of regression directly, rather than relying on unit tests of `BuildDependencyTree` in isolation.

---

Phavya-jfrog (approved):

Jordanh1996:
## dependency-tree cost

`appendCargoChildren` enumerates every *path* through `Cargo.lock` rather than every package, so
heavily shared crates (`syn`/`quote`/`proc-macro2`) get their whole subtree re-expanded once per
path. `NodeHasLoop()` stops cycles but not re-descent.

Real cargo-generated lockfiles, `jf` built from this branch:

| crates in lock | as-is | patched |
|---|---|---|
| 258 | 0.64 s / 58 MB | 0.31 s / 50 MB |
| 525 | 65 s / 10.6 GB | 0.48 s / 51 MB |
| 627 | 95 s / 12.0 GB | 0.70 s / 51 MB |

Identical output both ways (626 unique packages checked). Nothing crashes — this is cost, not
breakage. Crate count doesn't predict it; graph depth does. Adding `sqlx` is free; adding
`aws-sdk-s3`/`tonic`/`polars` costs 10 GB.

Fix: expand each package once, same guard as `cocoapods.go:246`. It has to be keyed on the package
(`name:version`), not on `node.Id` — keying on the node breaks
`TestBuildCargoDepTree/virtual_workspace_unions_every_local_package_under_a_synthetic_root`, since
that loop passes the same node with different packages.

Note `appendUvChildren` (`uv.go:1055`) is the same code with the same missing guard — probably worth
its own ticket.

Jordanh1996:
## config discovery

`jf ca` fails on any project whose `.cargo/config.toml` doesn't mention registries, even though
`cargo` itself resolves fine.

Real repo — `firecracker-microvm/firecracker` @ `81b38b9`. Its committed `.cargo/config.toml` has
only `[build]`, `[net]`, `[env]`. Artifactory redirect in `$CARGO_HOME`, as Set Me Up produces:

$ cargo generate-lockfile
    Updating artifactory index                                    <- cargo finds it

$ jf ca
[Error] cargo: no Artifactory registry found in .../firecracker/.cargo/config.toml
exit=1

`readEffectiveCargoConfig` returns the *first* config file it finds. Cargo instead merges every
ancestor `.cargo/` plus `$CARGO_HOME` key by key, so a file that says nothing about registries
hides nothing.

This is the common case, not an edge case — 8 of 9 well-known Rust repos I checked ship a
registry-silent `.cargo/config.toml`: rust-analyzer, deno, zed, nushell, astral-sh/uv, firecracker,
paradedb, and rust-lang/cargo itself. (polars was the only one with registry keys.)

Fix: walk the candidates in Cargo's precedence order until one names a registry. All existing cargo
tests pass.

Jordanh1996 (approved):

PR #860, opened 2026-08-26
XRAY-156690 - Rubygems native support

Phavya-jfrog:

-----

**Add native repository resolution for RubyGems in jf ca.**

Ruby (RubyGems) resolution priority:
1. .jfrog/projects/ruby.yaml (jf ruby-config)
2. ~/.gemrc's Artifactory-shaped entry in :sources: ("Set me up") — used only if (1) is absent
3. Clear error if neither resolves

--run-native has no effect for rubygems — resolution is already automatic from the priority above, so the flag is accepted as a no-op with a warning.

Screenshots attached in https://jfrog-int.atlassian.net/browse/XRAY-156690?focusedCommentId=1213539


sca/bom/buildinfo/technologies/gem/gem.go
```diff
@@ -467,3 +481,149 @@ func calculateUniqueDependencies(trees []*xrayUtils.GraphNode) []string {
...
+		separator := ":"
+		if runtime.GOOS == "windows" {
+			separator = ";"
+		}
+		for _, path := range strings.Split(envGemrc, separator) {
+			if path = strings.TrimSpace(path); path != "" {
+				paths = append(paths, path)
+			}
+		}
+	}
```
gauriy-tech, line 542: `GEMRC` only splits on one separator per OS (`:` on non-Windows, `;` on Windows). Real RubyGems accepts both on every platform (`ENV['GEMRC'].split(/[:;]/)` in `rubygems/config_file.rb`), so a `;`-separated `GEMRC` value on non-Windows collapses into a single non-existent path instead of being split — all listed gemrc files are silently dropped.

```suggestion
	if envGemrc := os.Getenv("GEMRC"); envGemrc != "" {
		isSeparator := func(r rune) bool { return r == ';' || (r == ':' && runtime.GOOS != "windows") }
		for _, path := range strings.FieldsFunc(envGemrc, isSeparator) {
			if path = strings.TrimSpace(path); path != "" {
				paths = append(paths, path)
			}
		}
	}
```

---

sca/bom/buildinfo/technologies/gem/gem.go
```diff
@@ -467,3 +481,149 @@ func calculateUniqueDependencies(trees []*xrayUtils.GraphNode) []string {
...
+type NotConfiguredError struct{ msg string }
+
+func (e *NotConfiguredError) Error() string { return e.msg }
+
+func newNotConfiguredError(format string, args ...any) error {
+	return errorutils.CheckError(&NotConfiguredError{msg: fmt.Sprintf(format, args...)})
+}
+
+// DefaultGemrcPaths returns gemrc file candidates in RubyGems' load order:
+// ~/.gemrc, followed by any files listed in GEMRC. A later file's ':sources:' overrides an earlier one.
```
gauriy-tech: This doc comment claims "RubyGems' load order" but real RubyGems also merges a system-wide config (e.g. `/etc/gemrc`) before `~/.gemrc`, which this function never consults. That's fine — the PR's stated design is that `~/.gemrc`'s Artifactory-shaped entry is "the only source" — but the comment overstates fidelity to Ruby's actual load order and could mislead a future maintainer.

```suggestion
// DefaultGemrcPaths returns the gemrc file candidates this feature consults:
// ~/.gemrc, followed by any files listed in GEMRC. A later file's ':sources:' overrides an earlier one.
// Note: unlike real RubyGems, this intentionally does not read a system-wide gemrc (e.g. /etc/gemrc) —
// per this feature's scope, ~/.gemrc's Artifactory-shaped source is the only supported source.
```

---

sca/bom/buildinfo/technologies/gem/gem.go
```diff
@@ -27,6 +31,7 @@ const (
 	stateSearchGEM        = iota
 	stateSearchSpecsKeyword
 	stateInSpecsSection
+	artifactoryApiGemsPath = "/api/gems/"
 )
```
gauriy-tech, line 34: `artifactoryApiGemsPath` is a string constant declared inside an `iota`-based int const block. No current bug, but it's a footgun: a future bare identifier appended after it in this block would silently inherit the string `"/api/gems/"` instead of continuing the `iota` sequence.

```suggestion
	stateSearchGEM        = iota
	stateSearchSpecsKeyword
	stateInSpecsSection
)

const artifactoryApiGemsPath = "/api/gems/"
```

---

commands/curation/curationaudit_test.go
```diff
@@ -538,7 +538,23 @@ func createTempHomeDirWithConfig(t *testing.T, basePathToTests string, testCase
...
 	WriteServerDetailsConfigFileBytes(t, config.ArtifactoryUrl, tempHomeDirPath, testCase.createServerWithoutCreds)
+
+	// Every Gem test case gets a ~/.gemrc with an Artifactory-shaped source, whether or not the
+	// fixture project also has a ruby.yaml. ruby.yaml takes priority when present (see
+	// setRepoFromGemrc), so this also exercises the both-present-yaml-wins case for fixtures
+	// that ship one, e.g. curation-project.
+	var callbackGemrc func()
+	if testCase.tech == techutils.Gem {
+		gemrcPath := filepath.Join(tempHomeDirPath, "gemrc")
+		gemSource := strings.TrimSuffix(config.ArtifactoryUrl, "/") + "/api/gems/ruby-remote/"
```
gauriy-tech, line 549: The `"gem tree - one blocked package"` integration test can't actually verify "ruby.yaml takes priority over ~/.gemrc" — `createTempHomeDirWithConfig` hardcodes every Gem test's `~/.gemrc` to repo `ruby-remote` (line ~549), and this fixture's `ruby.yaml` also declares `repo: ruby-remote`. Since both sources resolve to the identical repo name, reversing the precedence (gemrc winning over yaml) would still pass this test unchanged.

The existing unit test `TestSetRepoFromGemrc_RubyYamlPresent_TakesPriorityOverGemrc` already does this correctly with distinct repo names (`yaml-repo` vs `gemrc-repo`) — worth mirroring that here, e.g. by giving this fixture's `ruby.yaml` a repo name distinct from the hardcoded gemrc source and asserting the report reflects the yaml's repo.

---

commands/curation/curationaudit.go
```diff
@@ -1779,9 +1807,8 @@ func (ca *CurationAuditCommand) setRepoFromPipConf() error {
 		return nil
 	}
 
-	return errorutils.CheckErrorf(
-		"curation-audit for pip requires an Artifactory PyPI resolver. " +
-			"Either run 'jf pip-config' or configure index-url in your user pip.conf via Artifactory 'Set me up'.")
+	_, noConfigErr := ca.getRepoParams(projectType)
+	return noConfigErr
```
gauriy-tech, line 1811: This replaces a correct, actionable error (`"Either run 'jf pip-config'..."`) with `getRepoParams`'s generic fallback, which formats `'jf %s c'` — e.g. `"jf pip c"`, not a real command. This diff also newly routes Gem's "nothing configured" case through the same broken message (`curationaudit.go:2096-2097`).

The fix is one line in `getRepoParams` (`curationaudit.go:2164-2172`): use the same `'jf %s-config'` pattern `validateRunNativeForTech` already uses, instead of `'jf %s c'`. That resolves it for Pip (this regression), Gem (new code), and the pre-existing techs that already hit this (Maven/Go/Nuget/Gradle/Npm-fallback) all at once.

---

Phavya-jfrog: Deferring this — the fix touches getRepoParams, which is shared across Maven/Go/Nuget/Gradle/Npm/Yarn, not just Pip/Gem. It's the right fix in principle, but Maven's real command is jf mvn-config (not jf maven-config), so here would still be wrong for Maven.

gauriy-tech: Fair point, and checked it — you're right about `mvn-config` vs `maven-config`. For the record: that mismatch is Maven-specific. Confirmed the other techs currently routed through this message (`Go`, `Nuget`, `Gradle`, plus this diff's `Pip`/`Gem`) all have `tech.String()` matching their real command name exactly (`go-config`, `nuget-config`, `gradle-config`, `pip-config`, `ruby-config`) — only Maven needs a special case.

Deferring is reasonable given that — a correct fix needs a small tech→command-name exception map (or similar) rather than a one-line format change, and that's more scope than this PR should carry. Leaving this open as a follow-up rather than blocking on it.

commands/curation/curationaudit_test.go
gauriy-tech: Confirmed fixed — `ruby.yaml`'s `repo` is now `ruby-yaml-repo`, distinct from the hardcoded `~/.gemrc` source's `ruby-remote`, and the test's expected `BlockedPackageUrl`s were updated to match. This now actually exercises precedence (ran `TestSetRepoFromGemrc_RubyYamlPresent_TakesPriorityOverGemrc` and the full `TestSetRepoFromGemrc*` suite — all pass). No further action needed here.

gauriy-tech (approved):

PR #861, opened 2026-08-26
Fix SCA descriptor paths losing subdirectory prefix in SARIF output

orto17:

-----

When a project's descriptor lives in a subdirectory (e.g. Server/package.json), Frogbot's SARIF output — and consequently GitHub's Security → Code scanning alerts dashboard — shows the location as just "package.json", dropping the subfolder. The path is correct in the Xray UI, confirming the bug is in SARIF generation, not the underlying scan.

**Root cause**: TargetResults.GetDescriptors() computed each descriptor's path relative to the scan target's own working directory (sr.Target, e.g. /repo/Server) instead of the repo root. Since the working directory already includes the subfolder, filepath.Rel cancels it out, leaving only the filename. SAST/JAS findings don't have this bug because they're already anchored to the audit root directory.

**Fix**: GetDescriptors() now takes a rootDir parameter and resolves descriptor paths relative to it, falling back to sr.Target when empty (preserving existing behavior for any caller that doesn't have a root dir). Its one call site (commands/audit/audit.go) now passes params.RootDir(), which is already populated by Frogbot for every scan.

image

Concretely, a repo with both Server/package.json and Client/package.json today shows two GitHub code-scanning alerts that both just say package.json — indistinguishable. After the fix, each alert shows its full path and is unambiguous.


eranturgeman (approved):
LGTM!

PR #864, opened 2026-08-27
Add support in new scan results CDX upload via Xray

eranturgeman:

-----
This PR adds support in performing scan-cdx upload via Xray instead of a direct 'rt upload' call.
The existing mechanism uploads and creates the designated repo if doesnt exist. This requires an Artifactory ADMIN token.
This feature allows the user to keep providing only an Xray ADMIN token (like we used to provide in V2) and the entire process happens internally in Xray


go.sum
```diff
@@ -79,6 +79,8 @@ github.com/elazarl/goproxy v1.7.2 h1:Y2o6urb7Eule09PjlhQRGNsqRfPmYI3KKQLFpCAV3+o
 github.com/elazarl/goproxy v1.7.2/go.mod h1:82vkLNir0ALaW14Rc399OTTjyNREgmdL2cVoIbS6XaE=
 github.com/emirpasic/gods v1.18.1 h1:FXtiHYKDGKCW2KzwZKx0iC0PQmdlorYgdFG9jPXJ1Bc=
 github.com/emirpasic/gods v1.18.1/go.mod h1:8tpGGwCnJ5H4r6BWwaV6OrWmMoPhUl5jm/FMNAnJvWQ=
+github.com/eranturgeman/jfrog-client-go v0.0.0-20260901103048-89ef37f81c88 h1:WDgJM765Li+eX4GyjVw46WxckEe5kF7N0gT/cdb50YY=
```
Jordanh1996: placeholder: don't forget to revert

utils/utils.go
```diff
@@ -43,6 +43,7 @@ const (
 	GitRepoKeyAnalyticsMinXrayVersion = "3.114.0"
 	ExternalAnalyticsMinXrayVersion   = "3.152.3"
 	StaticScanMinVersion              = "3.133.0"
+	XrayCdxUploadMinVersion           = "3.154.0" // TODO eran: confirm this is the correct final version once the companion
```
Jordanh1996: placeholder: don't forget to clean before merge

utils/results/output/reportartifact.go
```diff
@@ -29,12 +33,60 @@ func UploadCommandResults(serverDetails *config.ServerDetails, rtResultRepositor
...
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed uploading the scan results via xray: %w", err)
+	}
+	return resp.Path, nil
+}
+
+// Used for direct upload only through upload-cdx command.
+// This flow still required an access token with necessary permission to upload an artifact directly to Artifactory and permission to create a repo in Artifactory if doeant exist.
+func uploadViaArtifactoryDirect(serverDetails *config.ServerDetails, rtResultRepository, artifactFinalRepoPath string, cmdType utils.CommandType, projectKey string, cdxResults *cdxutils.FullBOM) (artifactPath string, err error) {
```
Jordanh1996, line 83: On the comments above I'd write boldly this is deprecated and kept as backward compatibility for old xray versions

utils/results/output/reportartifact.go
```diff
@@ -29,12 +33,60 @@ func UploadCommandResults(serverDetails *config.ServerDetails, rtResultRepositor
...
+		Bom:      string(bomBytes),
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed uploading the scan results via xray: %w", err)
+	}
+	return resp.Path, nil
+}
+
+// Used for direct upload only through upload-cdx command.
+// This flow still required an access token with necessary permission to upload an artifact directly to Artifactory and permission to create a repo in Artifactory if doeant exist.
```
Jordanh1996:
```suggestion
// This flow still required an access token with necessary permission to upload an artifact directly to Artifactory and permission to create a repo in Artifactory if doesn't exist.
```

Jordanh1996:
Heads up: this came out of an automated review pass in my harness, not from me reading the diff line by line — and I don't fully follow it myself, so please sanity-check it rather than take it at face value.

The claim is about `RepoName: rtResultRepository` in `uploadViaXrayApi`:

`--upload-repo-path` is documented as *"Artifactory repository **name or path**"* (`cli/docs/flags.go:355`), so `cli-scan-results/myteam` is valid input. The old flow handled that: `reportartifact.go` joined repo+path into one string, and `createRepositoryIfNeededAndUploadFile` split the repo name back out with `strings.Split(scanResultsRepository, "/")[0]` — repo `cli-scan-results`, artifact under `myteam/...`.

The new call passes the value straight into `RepoName`, and the server rejects it (`repo_name must not contain '/'` in `verifyBodyFields`) → 400 → `AddGeneralError` fails the whole scan. Also, `myteam` isn't moved into `RepoPath`, so even if the server accepted it the artifact would land somewhere else than before.

Separately: the server whitelists only `frogbot` / `cli-scan-results` (+ project prefix), so any custom repo name that works today (the CLI creates the repo if missing) also 400s on 3.154+.

Frogbot itself is fine — it hardcodes `"frogbot"` — and the flag is hidden, so this is only `jf audit --static-sca` / `jf git audit` users who set it. Either split like the old code did, or narrow the flag's contract to "name only, one of the two allowed". Right now the code and the help text disagree.

utils/results/output/reportartifact.go
```diff
@@ -29,12 +33,60 @@ func UploadCommandResults(serverDetails *config.ServerDetails, rtResultRepositor
...
+	if bomBytes, err = sarifutils.StripUnsetIndexes(bomBytes); err != nil {
+		return "", fmt.Errorf("failed to sanitize CycloneDx SARIF indexes: %w", err)
+	}
+	xrayManager, err := securityxray.CreateXrayServiceManager(serverDetails, securityxray.WithScopedProjectKey(projectKey))
+	if err != nil {
+		return "", fmt.Errorf("failed creating xray service manager: %w", err)
+	}
+	resp, err := xrayManager.Xsc().UploadScanCdx(services.UploadScanCdxParams{
+		RepoName: rtResultRepository,
+		RepoPath: artifactFinalRepoPath,
```
Jordanh1996, line 71: Same disclaimer as my other comment: this came from an automated review pass in my harness, so please verify rather than trust.

`artifactFinalRepoPath` is sent verbatim, but the server validates every `/`-separated segment of `repo_path` against `^[A-Za-z0-9_\-.]+$` (plus non-empty, non-absolute, no `..`). That string was never validated before - it was just joined onto the repo name and used as an Artifactory deploy path, and Artifactory's path rules are much laxer.

I copied `getLocalArtifactPath` / `extractBaseGitPath` from this file and `verifyBodyFields` from the companion backend PR into a scratch file and fed realistic inputs through:

| input | repo_path produced | server |
|---|---|---|
| local audit, user `admin` | `admin` | ok |
| local audit, SaaS user (email username) | `jordanh@jfrog.com` | 400 invalid characters |
| local audit, reference token / API key | `""` | 400 repo_path is required |
| branch `main` | `github.com/jfrog/frogbot/main/commits` | ok |
| branch `feature/XRAY-1234-fix` | `.../feature/XRAY-1234-fix/commits` | ok |
| branch `fix(auth)` | `.../fix(auth)/commits` | 400 invalid characters |
| non-ascii branch `fix-café` | `.../fix-café/commits` | 400 invalid characters |
| self-hosted GitLab on a port | `gitlab.mycorp.com:8443/team/app/main/commits` | 400 invalid characters |
| Azure project with a space | `dev.azure.com/myorg/My Project/_git/app/main/commits` | 400 invalid characters |

The local-audit case is the sharp one: `getLocalArtifactPath` returns the token's username, and on SaaS that's an email address. So a plain `jf audit --static-sca --upload-cdx-results` with no git context 400s, and since `processScanResults` does `return cmdResults.AddGeneralError(...)` the whole scan fails and violations are never fetched. Same command on Xray < 3.154.0 succeeds via the Artifactory path.

I haven't run this end to end against an Xray carrying the backend PR - only client-builder into a copy of the validator - so worth confirming before acting on it. If it holds, the fix is either sanitizing the segments client-side or loosening the server regex to what these paths actually contain.

Jordanh1996 (approved):

utils/results/output/reportartifact.go
eranturgeman: @Jordanh1996 This is not deprecated.
This is used for the older Xray versions indeed but also used for the upload-cdx command that is still kept as is and is using direct upload and required Artifactory token

eranturgeman:
> Heads up: this came out of an automated review pass in my harness, not from me reading the diff line by line — and I don't fully follow it myself, so please sanity-check it rather than take it at face value.
> 
> The claim is about `RepoName: rtResultRepository` in `uploadViaXrayApi`:
> 
> `--upload-repo-path` is documented as _"Artifactory repository **name or path**"_ (`cli/docs/flags.go:355`), so `cli-scan-results/myteam` is valid input. The old flow handled that: `reportartifact.go` joined repo+path into one string, and `createRepositoryIfNeededAndUploadFile` split the repo name back out with `strings.Split(scanResultsRepository, "/")[0]` — repo `cli-scan-results`, artifact under `myteam/...`.
> 
> The new call passes the value straight into `RepoName`, and the server rejects it (`repo_name must not contain '/'` in `verifyBodyFields`) → 400 → `AddGeneralError` fails the whole scan. Also, `myteam` isn't moved into `RepoPath`, so even if the server accepted it the artifact would land somewhere else than before.
> 
> Separately: the server whitelists only `frogbot` / `cli-scan-results` (+ project prefix), so any custom repo name that works today (the CLI creates the repo if missing) also 400s on 3.154+.
> 
> Frogbot itself is fine — it hardcodes `"frogbot"` — and the flag is hidden, so this is only `jf audit --static-sca` / `jf git audit` users who set it. Either split like the old code did, or narrow the flag's contract to "name only, one of the two allowed". Right now the code and the help text disagree.

@Jordanh1996 Due to security hardening considerations and the fact that this is an implementation details and not for the user to control we dont allow any custom names in this flow anymore.
The users provide EXACTLY one out of 4 options: frogbot, <projectKey>-frogbot, cli-scan-results, <projectKey>-cli-scan-results.
I will block any other option here and the --upload-repo-path is removed completely

utils/results/output/reportartifact.go
eranturgeman: valid note
fixed - loosened the check and now we are blocking using a blacklist instead of allowing via whitelist.
also enhanced paths checks

PR #865, opened 2026-08-28
XRAY-156685 - Nuget native support

Phavya-jfrog:

-----
Add native repository resolution for NuGet in jf ca

NuGet/.NET resolution priority: 

1. .jfrog/projects/nuget.yaml (jf nuget-config)
2. Native NuGet/.NET CLI configured source ("Set me up") — used only if (1) is absent.
3. Clear error if neither resolves.

--run-native has no effect for NuGet/.NET — resolution is already automatic from the priority above, so the flag is accepted as a no-op with a warning.

Screenshots attached in https://jfrog-int.atlassian.net/browse/XRAY-156685?focusedCommentId=1223917


commands/curation/curationaudit.go
```diff
@@ -1890,6 +1904,61 @@ func (ca *CurationAuditCommand) setRepoFromNpmrc() error {
...
+	}
+	if serverDetails == nil || serverDetails.GetArtifactoryUrl() == "" {
+		return errorutils.CheckErrorf("curation-audit for NuGet requires a configured Artifactory server. Run 'jf c add' to configure a server")
+	}
+
+	registryConfig, err := nugettech.GetNativeNuGetRegistryConfig(serverDetails)
+	if err != nil {
+		log.Debug(fmt.Sprintf("NuGet: failed to read Artifactory details from native NuGet/.NET CLI sources: %s", err.Error()))
+		_, noConfigErr := ca.getRepoParams(projectType)
+		return noConfigErr
```
gauriy-tech: The native-source diagnostic from `GetNativeNuGetRegistryConfig` (e.g. "could not find a NuGet source configured...") is only logged at Debug and discarded here — the user sees just the generic "no config file was found" error, which never mentions the native-source path. `setRepoFromPipConf`'s native fallback returns a tailored error instead; `setRepoFromNpmrc` uses the `%w`-wrap pattern below.

```suggestion
		return fmt.Errorf("%w; additionally, native NuGet/.NET CLI source detection failed: %w", noConfigErr, err)
```

---

commands/curation/curationaudit_test.go
```diff
@@ -5690,6 +5810,80 @@ func TestHasCargoProject(t *testing.T) {
 	})
 }
 
+// TestDoCurationAudit_Nuget is a full integration test: it runs a real 'dotnet restore' against
+// the mock Artifactory server. NuGet has no 'jf nuget-config'; the registry is discovered by
+// matching the mock server's dynamic URL against a real, generated NuGet.Config (own harness,
+// like TestDoCurationAudit_Cargo, since the URL/port isn't known until the mock server starts).
+func TestDoCurationAudit_Nuget(t *testing.T) {
+	if err := exec.Command("dotnet", "--version").Run(); err != nil {
```
gauriy-tech: This skip guard only checks that `dotnet` is on PATH, not whether the pinned `net6.0` target framework pack (used by the testdata `.csproj`) is actually installed. On machines/CI images with only newer SDKs installed, this test hard-fails with `NU1102: Unable to find package Microsoft.NETCore.App.Ref with version (= 6.0.36)` instead of skipping — reproduced directly against this branch. Consider checking `dotnet --list-runtimes` for a 6.x pack, or dropping the hardcoded `net6.0` pin in the test fixture.

---

gauriy-tech:
The deleted `nuget.yaml`-fixture test case was the only end-to-end coverage for the explicit `jf nuget-config` priority path (yaml → resolved repo, full report pipeline). The new `TestDoCurationAudit_Nuget` only exercises the native-source fallback. Consider restoring an equivalent full end-to-end case for the yaml route, with an updated fixture set.

Also, `tests/testdata/projects/package-managers/dotnet/dotnet-curation/.jfrog/projects/dotnet.yaml` looks orphaned now — nothing reads it for NuGet's config path (`GetProjectConfFilePath` looks for `nuget.yaml`), so it may be safe to remove.

---

sca/bom/buildinfo/technologies/nuget/nuget.go
```diff
@@ -327,6 +327,173 @@ func validateInputCommand(completeCommandArgs []string) (string, []string, error
...
+		sourceHost, hostErr := schemeAndHostOf(source.url)
+		if hostErr != nil || !strings.EqualFold(sourceHost, artifactoryHost) {
+			continue
+		}
+		rtBaseUrl, repoName, parseErr := parseArtifactoryNugetSourceUrl(source.url)
+		if parseErr != nil {
+			return nil, fmt.Errorf("NuGet source %q (%s) matches the configured Artifactory host %q but is not a recognizable Artifactory NuGet repository URL: %w", source.name, source.url, artifactoryHost, parseErr)
+		}
+		return &NuGetRegistrySourceConfig{SourceName: source.name, ArtifactoryUrl: rtBaseUrl, RepoName: repoName}, nil
+	}
```
gauriy-tech, line 451: This picks the first native source matching the configured Artifactory host with no signal if more than one matches. For a curation tool, that means the wrong repository's curation policy could be enforced silently — e.g. a leftover "Set me up" source plus a newly added one, both pointing at the same host. Consider collecting all matches and logging a `Warn` naming them, while still keeping the first match for backward compatibility:

```suggestion
	var matched []nugetSource
	for _, source := range sources {
		sourceHost, hostErr := schemeAndHostOf(source.url)
		if hostErr != nil || !strings.EqualFold(sourceHost, artifactoryHost) {
			continue
		}
		matched = append(matched, source)
	}
	if len(matched) > 1 {
		names := make([]string, len(matched))
		for i, source := range matched {
			names[i] = source.name
		}
		log.Warn(fmt.Sprintf("Multiple configured NuGet sources match the Artifactory host %q: %s. Using %q; remove or rename the unused sources to avoid ambiguity.",
			artifactoryHost, strings.Join(names, ", "), matched[0].name))
	}
	if len(matched) > 0 {
		rtBaseUrl, repoName, parseErr := parseArtifactoryNugetSourceUrl(matched[0].url)
		if parseErr != nil {
			return nil, fmt.Errorf("NuGet source %q (%s) matches the configured Artifactory host %q but is not a recognizable Artifactory NuGet repository URL: %w", matched[0].name, matched[0].url, artifactoryHost, parseErr)
		}
		return &NuGetRegistrySourceConfig{SourceName: matched[0].name, ArtifactoryUrl: rtBaseUrl, RepoName: repoName}, nil
	}
```

---

commands/curation/curationaudit.go
```diff
@@ -778,6 +779,7 @@ func (ca *CurationAuditCommand) techsToAudit() []string {
 		techs := promotePnpmWorkspaceMember(techutils.DetectedTechnologiesListForCurationAudit())
 		techs = promoteYarnWorkspaceMember(techs)
 		techs = promotePipToUv(techs)
```
gauriy-tech, line 786: Missing test coverage:

- `selectMatchingNuGetSource` has no test with 2+ sources matching the same Artifactory host (ties to the ambiguity comment on this function).
- The new `Dotnet`-exclusion dedup (`slices.DeleteFunc(techs, ...)` a few lines below) has no dedicated unit test — it's only indirectly covered by the integration tests, which would fail if the dedup broke, but a small direct unit test would pin the behavior explicitly.

---

gauriy-tech (approved):

PR #873, opened 2026-09-06
Fix Audit SCA parallel runs

attiasas:
## Summary
`--threads` was supposed to run SCA per target in parallel, but `ResultsMu` was held for the entire scan-graph/enrich call. Threads were created, yet SCA work stayed sequential (JAS already scaled because it only locks when writing results).

This PR:
- Locks `ResultsMu` only while attaching SCA results, matching JAS.
- Clones scan-graph params per target so concurrent scans do not share `DependenciesGraph` / technology.
- Adds tests that overlapping SCA tasks actually run together (`-race` clean).

Dependency-tree / SBOM generation is still sequential (`os.Chdir` per target). That is out of scope here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved scan isolation to prevent concurrent SCA scans from overwriting each other’s dependency or technology data.
  - Ensured scan results are written safely when multiple targets are processed at the same time.
  - Improved consistency of results across both legacy scan-graph and newer SBOM enrichment flows.

- **Performance**
  - SCA scans now process multiple targets in parallel, reducing overall scan completion time.

### DIFF
--- a/utils/xray/scangraph/params.go
+++ b/utils/xray/scangraph/params.go
@@ -59,8 +59,6 @@ func (sgp *ScanGraphParams) Technology() techutils.Technology {
 	return sgp.technology
 }
 
-// Clone returns a copy that can be mutated per scan without racing other concurrent scans
-// that share the original params (DependenciesGraph and Technology are set per target).
 func (sgp ScanGraphParams) Clone() ScanGraphParams {
 	cloned := sgp
 	if sgp.xrayGraphScanParams != nil {
--- a/utils/xray/scangraph/params.go
+++ b/utils/xray/scangraph/params.go
@@ -58,3 +58,14 @@ func (sgp *ScanGraphParams) SetTechnology(technology techutils.Technology) *Scan
 func (sgp *ScanGraphParams) Technology() techutils.Technology {
 	return sgp.technology
 }
+
+// Clone returns a copy that can be mutated per scan without racing other concurrent scans
+// that share the original params (DependenciesGraph and Technology are set per target).
+func (sgp ScanGraphParams) Clone() ScanGraphParams {
+	cloned := sgp
+	if sgp.xrayGraphScanParams != nil {
+		paramsCopy := *sgp.xrayGraphScanParams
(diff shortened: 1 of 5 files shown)


utils/xray/scangraph/params.go
```diff
@@ -58,3 +58,14 @@ func (sgp *ScanGraphParams) SetTechnology(technology techutils.Technology) *Scan
 func (sgp *ScanGraphParams) Technology() techutils.Technology {
 	return sgp.technology
 }
+
+// Clone returns a copy that can be mutated per scan without racing other concurrent scans
+// that share the original params (DependenciesGraph and Technology are set per target).
```
Jordanh1996: i'd skip the comment above. "clone" speaks for itself I believe.

utils/xray/scangraph/params.go
```diff
@@ -58,3 +58,14 @@ func (sgp *ScanGraphParams) SetTechnology(technology techutils.Technology) *Scan
...
+func (sgp ScanGraphParams) Clone() ScanGraphParams {
```
Jordanh1996, line 62: Clone() is one level deep: Watches, BinaryGraph, and serverDetails stay aliased.
intentional?

Jordanh1996 (approved):

utils/xray/scangraph/params.go
attiasas: Removed. The other exported setters in this file are undocumented too.

utils/xray/scangraph/params.go
attiasas: Yes. Clone copies ScanGraphParams and XrayGraphScanParams by value so this path can set DependenciesGraph and Technology per target without racing.

Watches, BinaryGraph, and serverDetails stay aliased on purpose: they are shared read-only config here. Watches is only ranged into query params, BinaryGraph is unused on the source SCA path (we always set DependenciesGraph), and serverDetails is only read. Deep-copying those would not change isolation unless we started mutating them in place.

PR #874, opened 2026-09-06
Add NuGet Package Updater For plain PackageReference Fixes

orto17:

-----

Adds a NugetPackageUpdater (remediation/sca/packageupdaters/nugetpackageupdater.go) implementing the PackageUpdater interface for NuGet/.NET fix PRs.

Covers two use cases:
1. Plain PackageReference, no lock file — patches the Version in place (byte-level regex, no XML re-serialization, matching the existing Maven updater's approach), supporting both Include=/Update= attributes and both attribute-form and child-element-form Version.
2. Plain PackageReference, with packages.lock.json present — additionally regenerates the lock file via dotnet restore --force-evaluate --no-dependencies (only if the lock file is git-tracked), rolling back both the .csproj and the lock file to their pre-fix content on any restore failure.

--no-dependencies flag keeps a fix scoped to the touched project's own lock file rather than also restoring (and diffing) every project it references via ProjectReference.

A. Deliberately out of scope:

1. packages.config - legacy NuGet CLI projects.
2. Indirect (transitive) dependencies — IndirectDependencyFixNotSupported, same convention as the existing Maven/npm/Go updaters.

B. Detection-side blind spots (not fixable by the updater alone):

1. Central Package Management (Directory.Packages.props) — detected and reported distinctly via CentralPackageManagementFixNotSupported.
2. Directory.Build.props/.targets-injected PackageReference — not considered at all; a package declared only there currently reports a plain "not found" not a distinct diagnosis.
3. A CPM or Directory.Build.props-managed dependency with no packages.lock.json anywhere in the repo — Xray itself can't resolve a concrete version for it, so the fix flow is never invoked regardless of updater logic.

C. Syntax the regex-based approach can't resolve:

1. Include="$(SomeMSBuildProperty)" (property-valued, not a literal package name) — exact-string match fails, reports "not found"; would need real MSBuild evaluation to resolve.
2. A vulnerable package with no existing <PackageReference> at all (e.g. an SDK-implicit package) — fixing this requires inserting a new pinning element, not patching an existing one; not supported.
3. Include="A;B;C" (semicolon-delimited multi-package Include) — legal but rare MSBuild syntax; reports "not found" rather than resolving the individual package (safe failure, not a crash).

## Summary by CodeRabbit

- **New Features**
  - Added NuGet remediation support for vulnerable dependencies in C#, F#, and VB.NET project files.
  - Updates package versions in project references and regenerates existing NuGet lock files when needed.
  - Supports multiple projects while preserving rollback behavior when restoration fails.
  - Provides clear handling for indirect dependencies and references without inline versions.

- **Bug Fixes**
  - NuGet projects are now correctly recognized across supported project file types.


remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil
+	}
+	if fixedAny {
+		// At least one descriptor was fixed - don't fail the whole vulnerability just because a
+		// sibling descriptor (e.g. one governed by Central Package Management) couldn't be fixed.
+		// A caller treating any error as "nothing happened" would otherwise discard an
+		// already-applied, successful fix.
+		log.Warn(fmt.Sprintf("Partially fixed '%s': could not fix descriptor(s) [%s]: %s", fixDetails.ImpactedDependencyName, strings.Join(failingDescriptors, ", "), fixErrors.Error()))
+		return nil
+	}
```
attiasas, line 102: If one project is patched and a sibling fails (CPM / restore), this returns nil. Maven/npm/Go still return the joined error in that situation.

I agree we shouldn’t throw away an already-applied fix, but nil usually means “this issue is handled” to Frogbot. Can we confirm the caller still commits the successful files and does not mark the CVE fully resolved when `failingDescriptors` is non-empty?

If this is the intended contract, a short comment at the `PackageUpdater` / Frogbot call site would help the next person. Returning the error while keeping the successful writes (Maven’s model) is the other option — then the caller decides whether to keep the diff.

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -0,0 +1,264 @@
...
+
+	lockFileTracked, checkErr := IsFileTrackedByGit(lockFilePath, originalWd)
+	if checkErr != nil {
+		log.Debug(fmt.Sprintf("Failed to check if lock file is tracked in git: %s. Proceeding with lock file regeneration.", checkErr.Error()))
+		lockFileTracked = true
+	}
+	if !lockFileTracked {
+		log.Debug(fmt.Sprintf("Lock file '%s' is not tracked in git, skipping lock file regeneration", lockFilePath))
+		return nil
+	}
```
attiasas, line 162: `IsFileTrackedByGit(lockFilePath, originalWd)` uses `filepath.Rel(repoRoot, filePath)`. If evidence paths are relative and cwd is absolute, `Rel` often errors; we then set `lockFileTracked = true` and run restore anyway.

That’s the same fail-open as npm, but there’s no test that an untracked `packages.lock.json` is left untouched (tests aren’t a git repo, so they always take the fail-open path).

Suggestion: `filepath.Abs` the lock path (and/or join with `originalWd` when not absolute) before the git check, and add a test that restore is skipped when the file is genuinely untracked.

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -0,0 +1,264 @@
...
+	originalLockFile, err := os.ReadFile(lockFilePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// No lock file for this project - nothing further to regenerate.
+			return nil
+		}
+		return rollbackProjectFile(projectFilePath, originalProjectFile, fmt.Errorf("failed to read %s: %w", lockFilePath, err))
+	}
+
+	lockFileTracked, checkErr := IsFileTrackedByGit(lockFilePath, originalWd)
```
attiasas, line 154: Not introduced here, but now that NuGet is a fifth caller of `IsFileTrackedByGit` (`commonpackageupdater.go`): this belongs in `utils/scm`, not in packageupdaters.

Caveat: don’t put it on today’s `GitManager` — `NewGitManager` requires a remote, and this check only needs HEAD’s tree. A `scm.IsFileTracked(repoRoot, path)` that `PlainOpen`s and looks up the tree (and `filepath.Abs`’s relative evidence paths) would fix the Rel fail-open once for npm/pnpm/yarn/go/nuget.

Happy for that to be a follow-up rather than this PR.

remediation/sca/packageupdaters/nugetpackageupdater_test.go
```diff
@@ -0,0 +1,821 @@
...
+	biutils "github.com/jfrog/build-info-go/utils"
+	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/jfrog/jfrog-cli-security/tests/utils/integration"
+	"github.com/jfrog/jfrog-cli-security/utils/formats"
+	"github.com/jfrog/jfrog-cli-security/utils/techutils"
+)
+
+func writeFakeDotnetRestore(t *testing.T, dir string, exitCode int, lockFileContent string, createObjDir bool) {
```
attiasas, line 21: The fake `dotnet` is a good way to assert flags, rollback, and isolation on Windows. It never validates that `dotnet restore --force-evaluate --no-dependencies` actually rewrites a real lock file.

Not a blocker if CI images don’t have an SDK. If they do, even one opt-in test against a tiny `net8.0` project would catch a wrong flag. If they don’t, calling that out in the PR description is enough.

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -0,0 +1,264 @@
...
+
+	"github.com/jfrog/jfrog-client-go/utils/log"
+)
+
+// nugetProjectFileSuffixes are matched case-insensitively, since MSBuild project file suffixes
+// aren't guaranteed to be written in any particular casing.
+var nugetProjectFileSuffixes = []string{".csproj", ".fsproj", ".vbproj"}
+
+const (
+	nugetPackageReferenceElementPattern = `(?s)<PackageReference\b[^>]*/>|<PackageReference\b[^>]*[^/]>.*?</PackageReference>`
```
attiasas, line 21: Nit: project suffixes are matched case-insensitively, but `<PackageReference` is case-sensitive. MSBuild element names aren’t. Rare, but `<packagereference Include="Foo" Version="1.0" />` would report “not found”. `(?i)` on the element pattern (or documenting the limitation next to the Include="$(X)" notes) would be enough.

attiasas (commented):
Nice Job, check out my comments

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: UpdateDependency now always returns the joined error when any descriptor fails — matching Maven's and npm's existing contract

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: Before the git check, the lock file path is now made absolute (joined with originalWd when not already absolute), so filepath.Rel inside IsFileTrackedByGit no longer errors and fails open. Added TestNugetUpdateDependencySkipsRestoreForUntrackedLockFile.

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: Left as-is —  as this is a cross-ecosystem change (npm/pnpm/go/yarn all share the same helper), not NuGet-specific.

remediation/sca/packageupdaters/nugetpackageupdater_test.go
orto17: Added a NuGet case to the existing real-CLI TestUpdateDependency suite in commonpackageupdater_test.go. It runs the actual dotnet restore --force-evaluate --no-dependencies against a real fixture and confirms the lock file is genuinely rewritten.

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: Added (?i) to nugetPackageReferenceElementPattern (MSBuild element names are case-insensitive).

attiasas (approved):
LGTM

PR #875, opened 2026-09-08
Add Yarn V3/4 Package Updater 

orto17:
Regenerates yarn.lock only via 'yarn install --mode=update-lockfile', rejecting Yarn 1.x/2.x and missing/broken yarn executables as unsupported rather than failing, and rolling back package.json on install failure. Runs from the workspace root so a fix in a member updates the single root yarn.lock, and cleans up install-time artifacts (.yarn/install-state.gz, .pnp.cjs, .yarn/cache) that '--mode=update-lockfile' can still leave behind.

-----

- Adds YarnPackageUpdater: fixes a direct-dependency violation by editing package.json and regenerating yarn.lock only, via yarn install --mode=update-lockfile — no full install, no node_modules, no lifecycle scripts.
- Supports Yarn 3.x/4.x only. Yarn 1.x/2.x, indirect-dependency violations, and a runner with no usable Yarn are reported as unsupported (ErrUnsupportedFix) rather than failing the run or being silently skipped.
- Runs from the workspace root, not the descriptor's own directory — a fix in a workspace member updates that member's package.json and the single root yarn.lock.
- Rolls back package.json if the install fails, leaving the working tree untouched.
- Cleans up install-time bookkeeping (.yarn/install-state.gz, .pnp.cjs, .yarn/cache) that --mode=update-lockfile can still leave behind, when not already present before the fix.
- Adds techutils.Yarn to SupportedFixTechnologies.


remediation/sca/packageupdaters/yarnpackageupdater.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+	}
+
+	lockFilePath := filepath.Join(rootDir, yarnLockFileName)
+	lockFileTracked, checkErr := IsFileTrackedByGit(lockFilePath, originalWd)
+	if checkErr != nil {
+		log.Debug(fmt.Sprintf("Failed to check if lock file is tracked in git: %s. Proceeding with lock file regeneration.", checkErr.Error()))
+		lockFileTracked = true
+	}
+	if !lockFileTracked {
```
Jordanh1996, line 126: I believe if lock file is not tracked in git. but package.json is. then a fix would give a "false positive" and not produce the fix(because UpdatePackageJSONDescriptor is only in line 130 and you return here)

remediation_test.go
```diff
@@ -122,13 +128,143 @@ func TestRemediationGo(t *testing.T) {
...
+	assert.NoDirExists(t, filepath.Join(tmpDir, "node_modules"))
+	assert.NoFileExists(t, filepath.Join(tmpDir, ".pnp.cjs"))
+	assert.NoFileExists(t, filepath.Join(tmpDir, ".yarn", "install-state.gz"))
+}
+
+// TestRemediationYarnV4 verifies the fix also works against a real Yarn 4 project.
+func TestRemediationYarnV4(t *testing.T) {
+	integrationUtils.InitRemediationTest(t)
+
+	tmpDir, cleanup := testutils.CreateTestProjectEnvAndChdir(t, remediationProjectDir("yarn", "remediation-v4"))
```
Jordanh1996: `remediation-v4` is not shipped in this PR? yet a test for it exists here?

remediation_test.go
```diff
@@ -122,13 +128,143 @@ func TestRemediationGo(t *testing.T) {
...
+	descriptorAfter, err := os.ReadFile("package.json")
+	require.NoError(t, err)
+	assert.Equal(t, descriptorBefore, descriptorAfter, "package.json should be rolled back to its original state")
+}
+
+// TestRemediationYarnUnsupportedVersion verifies that a Yarn 2.x project returns ErrUnsupportedFix.
+func TestRemediationYarnUnsupportedVersion(t *testing.T) {
+	integrationUtils.InitRemediationTest(t)
+
+	_, cleanup := testutils.CreateTestProjectEnvAndChdir(t, remediationProjectDir("yarn", "yarn-v2"))
```
Jordanh1996: doesn't reach the version gate. The yarn-v2 fixture has no yarn.lock, so findYarnLockfileRoot errors first and the ErrorAs assertion fails on no yarn.lock found in '.' or any parent directory.

remediation_test.go
```diff
@@ -22,6 +23,11 @@ func remediationProjectDir(tech, subdir string) string {
 	return filepath.Join(filepath.FromSlash(securityTests.GetTestResourcesPath()), "projects", "package-managers", tech, subdir)
 }
 
+func copyYarnRelease(t *testing.T, destDir, canonicalFixture string) {
```
Jordanh1996: The test needs a Yarn 3 binary. It looks for it in the wrong folder, finds nothing, copies nothing, and doesn't complain. So Yarn is missing and every test dies immediately.

The wrong folder comes from one thing: the path is built from the current directory, and the current directory was changed a line earlier.

line 1:  CreateTestProjectEnvAndChdir(...)   →  cwd is now /tmp/xxx/remediation
line 2:  copyYarnRelease(...)                →  builds path from cwd

         looks in:  /tmp/xxx/remediation/tests/testdata/.../yarn-v3/.yarn    (empty, nothing there)
         should be: /repo/tests/testdata/.../yarn-v3/.yarn                   (the binary is here)

Nothing gets copied. The copy function returns success anyway on a missing folder, so the test carries on. Then Yarn can't start, and the updater bails with "no usable yarn" before doing anything.

remediation/sca/packageupdaters/yarnpackageupdater.go
orto17: Good catch. The skip now happens after `UpdatePackageJSONDescriptor`, matching npm/pnpm: a tracked `package.json` is still rewritten when `yarn.lock` is untracked, and only lockfile regeneration is skipped. Covered by `TestRemediationYarnUntrackedLockfile`.

remediation_test.go
orto17: The `remediation-v4` fixture was local-only. It is now included in the PR (`tests/testdata/projects/package-managers/yarn/remediation-v4/`).

remediation_test.go
orto17: Agreed — `yarn-v2` has no `yarn.lock`, so the version gate was never reached. The test now uses a dedicated `remediation-v2` fixture that includes `yarn.lock` and copies the Yarn 2 binary from `yarn-v2`.

remediation_test.go
orto17: Fixed. The Yarn release path is resolved with `filepath.Abs` before `CreateTestProjectEnvAndChdir`, and `require.DirExists` fails the test if the binary fixture is missing instead of copying an empty tree.

remediation_test.go
orto17: yarn-v2 has no yarn.lock, so the test failed in findYarnLockfileRoot. It now uses a dedicated remediation-v2 fixture that includes a lockfile.

Jordanh1996 (approved):

PR #878, opened 2026-09-09
XRAY-159447 - cvs fallback for pypi

Phavya-jfrog:

-----
## Summary
Fixes CVS (curation) fallback reporting for PyPI (pip/poetry/pipenv/uv) so pinned, ranged, and
transitive whole-package blocks show consistent, npm-parity policy detail.
- Adds a whole-package check before probing a specific version: if the package is blocked at
  every version, the row renders as `All versions blocked` (mirrors npm), instead of resolving
  to one version and reporting on just that one.
- Normalizes `Explanation`/`Recommendation` to npm's generic text when a policy matches the
  whole-package block, but keeps the real per-version detail when it's a genuinely different
  policy (e.g. `immature-30` on one version vs. an `openssf` block on the whole package).
- Adds a loose "depends on" fallback to recover the real parent for transitive blockers across
  pip/poetry/pipenv/uv when the child has no exact version in the tool's error text.
  
## Testing
- 9 new unit tests in `curationaudit_test.go` covering pinned/ranged/transitive whole-package
  blocks and same-policy vs. different-policy normalization.
- Verified end-to-end against a live D2C environment across pip, poetry, pipenv, and uv.
Screenshot 2026-09-09 at 10 38 48 PM
Screenshot 2026-09-09 at 10 39 05 PM

### DIFF
--- a/commands/curation/curationaudit.go
+++ b/commands/curation/curationaudit.go
@@ -84,6 +84,11 @@ const (
 
 	extractPoliciesRegexTemplate = "({.*?})"
 
+	// allVersionsBlockedMsgPattern matches Artifactory's whole-package-block message, e.g.
+	// "All versions blocked - {policy:openssf-jfca,condition:openssf}". Unlike the per-version
+	// download-block message (gated by BlockMessageKey), it only carries policy + condition.
+	allVersionsBlockedMsgPattern = `(?i)all versions blocked\s*-\s*\{\s*policy\s*:\s*([^,}]+?)\s*,\s*condition\s*:\s*([^}]+?)\s*\}`
+
 	errorTemplateHeadRequest = "failed sending HEAD request to %s for package '%s:%s'. Status-code: %v. Cause: %v"
 
 	errorTemplateUnsupportedTech = "It looks like this project uses '%s' to download its dependencies. " +
@@ -102,13 +107,12 @@ const (
 	MinXrayPassThroughSupport = "3.92.0"
 	MinArtiGradleGemSupport   = "7.63.5"
 
-	// cvsPartialReportWarning is shown when pip or poetry resolution failed because CVS
-	// stripped a required version from the simple index, but the metadata-API
-	// fallback succeeded in recovering at least one policy violation.
-	cvsPartialReportWarning = "The curation audit was unable to fully resolve the dependency tree because one or more pinned package versions " +
-		"are blocked by the curation policy. Details of the policy violations are shown in the table below.\n" +
-		"Dependency analysis cannot proceed until these issues are addressed.\n" +
-		"Once you switch to an approved version and re-run the audit, additional results will be available."
+	// cvsPartialReportWarningTemplate is shown when a Python package manager's resolution
+	// failed due to a curation block, but the fallback still recovered a partial report.
+	// %s is the technology name (pip/poetry/pipenv/uv), filled in at print time.
+	cvsPartialReportWarningTemplate = "%s could not fully resolve the dependency tree because one or more packages " +
+		"are blocked by the curation policy, so this report may be incomplete. Dependencies of blocked packages " +
+		"could not be analyzed because their metadata was unavailable."
 
 	// hfUnresolvedReportKey is used when an HF scan found only dynamic references — no table, just warnings.
 	hfUnresolvedReportKey = "huggingface (unresolved references)"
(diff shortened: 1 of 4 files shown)


commands/curation/curationaudit.go
```diff
@@ -2474,6 +2479,15 @@ func (ca *CurationAuditCommand) runCvsFallback(cvsErr *python.CvsBlockedError, t
 	return nil
 }
 
+// cvsFallbackTechName returns the package-manager name to show in
+// cvsPartialReportWarningTemplate, taken from the first row's PkgType.
+func cvsFallbackTechName(report *CurationReport) string {
+	if len(report.packagesStatus) > 0 && report.packagesStatus[0].PkgType != "" {
+		return report.packagesStatus[0].PkgType
+	}
+	return "the package manager"
```
gauriy-tech, line 2490: `"the package manager"` fallback is currently unreachable: `runCvsFallback` never produces a report with an empty `packagesStatus`, and every `PackageStatus` on this path always sets a non-empty `PkgType`. The other `isPartial` producer (npm-log fallback) can hit this branch, but its result is discarded immediately after (the `npmLogPartial` check overwrites `warningText` on the next line). Harmless — flagging as a cleanup opportunity, not a required fix.

---

commands/curation/curationaudit_test.go
```diff
@@ -5402,6 +5408,759 @@ func TestFetchCvsBlockedStatusUvTransitive(t *testing.T) {
...
+	require.NoError(t, err)
+	rtAuth := rtManager.GetConfig().GetServiceDetails()
+	analyzer := treeAnalyzer{
+		rtManager:            rtManager,
+		extractPoliciesRegex: regexp.MustCompile(extractPoliciesRegexTemplate),
+		rtAuth:               rtAuth,
+		httpClientDetails:    rtAuth.CreateHttpClientDetails(),
+		url:                  rtAuth.GetUrl(),
+		repo:                 repo,
+		tech:                 techutils.Pip,
```
gauriy-tech, line 5467: This pinned-exact whole-package-block enrichment path (Steps 1-3 with `wholePackageBlocked=true`) is only exercised for `techutils.Pip` — the cross-tech table test (`TestFetchCvsBlockedStatusWholePackageBlockedAcrossTechs`) only covers the ranged short-circuit case for Poetry/Pipenv/Uv, not a pinned-exact entry. The HEAD/GET probing is gated by the same tech conditional for all four, so this is provably tech-agnostic today — but worth a test case per tech to guard against a future regression that special-cases pip.

---

sca/bom/buildinfo/technologies/python/python_cvs_fallback.go
```diff
@@ -107,6 +116,20 @@ var uvDependsOnParentRegex = regexp.MustCompile(
 	`([A-Za-z0-9][A-Za-z0-9._-]*)(?:\[[^\]]*\])?==([0-9][0-9A-Za-z._+\-]*)\s+depends on\s+` +
 		`([A-Za-z0-9][A-Za-z0-9._-]*)(?:\[[^\]]*\])?==([0-9][0-9A-Za-z._+\-]*)`)
 
+// looseDependsOnParentRegex is a fallback for "parent==version depends on child" clauses
+// where the child has no exact version (constrained by a range, not a pin). Applied last,
+// after tool-specific parsing, to attribute otherwise self-attributed blockers.
+var looseDependsOnParentRegex = regexp.MustCompile(
```
gauriy-tech, line 122: `looseDependsOnParentRegex`/`pipDependsOnParentRegex` reattribute via loose, whole-output, first-match-wins matching keyed only by child name (mirrors the existing, unchanged `poetryParentByChild` pattern, now extended to pip/pipenv). This is a real — not just theoretical — risk under pip's classic diamond-dependency/version-conflict output: two independent chains that both depend on the same child name could get the wrong parent named in the remediation text. It only affects the user-facing "is a transitive dependency of X" wording, not the block/waiver decision. Recommend adding one case to the existing table-driven `TestParseCvsFailedPackages` with two independent chains colliding on the same child name, to document current first-match-wins behavior.

---

commands/curation/curationaudit.go
Phavya-jfrog: Fixed — call moved so it's only computed when actually used

commands/curation/curationaudit_test.go
Phavya-jfrog: Added test that covers pip, poetry, pipenv, uv.

sca/bom/buildinfo/technologies/python/python_cvs_fallback.go
Phavya-jfrog: Test added — two packages depending on the same blocked package; confirms we pick the first as "parent" in the message and locks that behavior in.

gauriy-tech (approved):

commands/curation/curationaudit.go
```diff
@@ -126,6 +128,7 @@ const (
 var CurationOutputFormats = []string{string(outFormat.Table), string(outFormat.Json)}
 var osGetwd = os.Getwd
 var npmGetConfigValue = npmtech.GetNpmConfigValue
+var allVersionsBlockedMsgRegex = regexp.MustCompile(allVersionsBlockedMsgPattern)
```
Phavya-jfrog, line 131: False positive — Go's regexp can't backtrack, so ReDoS isn't possible here.

PR #880, opened 2026-09-10
Resolve Pip/Uv detection ambiguity in the shared detector

orto17:

-----

A pyproject.toml with no [tool.poetry] / hatchling / flit / pdm build-backend section defaults to Pip detection (pyProjectTomlIndicatorContent). This means a uv-managed project — which has pyproject.toml + uv.lock and typically none of those build-backend markers — was detected as both Pip and Uv for the same scan target.

### Impact of the bug
When a target has more than one detected technology sharing the same Xray ecosystem (pypi), techutils.ResolveIssueTechnology cannot disambiguate and collapses the resolved Technology to NoTech (empty/unknown) for every vulnerability found there. Verified live end-to-end against jf audit --static-sca:

- Real uv project (pyproject.toml + uv.lock): detected as [pip, uv] → resolved technology for every vulnerability came back empty/unknown.
- Equivalent minimal pip, pipenv, and poetry fixtures: each detected as a single, correct technology, unaffected.

### Fix
`jf curation-audit` already solves this exact ambiguity via `promotePipToUv`: a pip-exclusive file (requirements.txt, setup.py, setup.cfg, Pipfile, poetry.lock) wins over a stray uv.lock; otherwise, any uv signal (uv.lock, pyproject.toml `[tool.uv]`/`[[tool.uv.index]]`, or `~/.config/uv/uv.toml`) wins and Pip is dropped. That logic only ran for curation-audit, not for the general `jf audit`/`jf scan` flow (and therefore not for Frogbot).

An earlier version of this PR fixed the ambiguity by excluding `uv.lock` from Pip's indicators in `utils/techutils/techutils.go`. That approach was reverted: it mutated the shared base detector used by every command, which broke curation-audit's own `promotePipToUv` rule for the "pip project with a stray leftover uv.lock" case (`requirements.txt` + `uv.lock` should still resolve to pip, not uv) — that rule never got a chance to run since Pip was already stripped from the candidate list upstream.

Instead, this PR extracts `promotePipToUv`'s rule into a shared, exported `techutils.PromotePipToUv(techs []Technology, dir string)`, and wires it into both places:
- `commands/curation/curationaudit.go` — now delegates to the shared function; behavior is unchanged, existing tests pass as-is.
- `commands/audit/audit.go` — added to both detection flows (`detectTechnologiesInTarget` for the new BOM-generator flow, `detectScaTargetsFromTechnologies` for the old flow Frogbot currently uses), which previously had no disambiguation at all for this ambiguity.

### Verification
Re-ran detection (`techutils.DetectTechnologiesDescriptors` + `techutils.PromotePipToUv`) against one fixture per manager:

#### Detection results per package manager

| Package manager | Fixture files | Raw detected | After promotion |
|---|---|---|---|
| pip    | `requirements.txt`                          | `[pip]`      | `[pip]`      |
| pipenv | `Pipfile`, `Pipfile.lock`                    | `[pipenv]`   | `[pipenv]`   |
| poetry | `pyproject.toml` (`[tool.poetry]`), `poetry.lock` | `[poetry]` | `[poetry]`   |
| uv     | `pyproject.toml`, `uv.lock`                  | `[pip, uv]` ❌ | `[uv]` ✅  |
| pip + stray `uv.lock` | `requirements.txt`, `uv.lock` (leftover) | `[pip, uv]` | `[pip]` ✅ |

The last row is the regression case `TestPipWinsOverStrayUvLock` already guards: a genuine pip project with a stray leftover `uv.lock` must still resolve to pip, not uv. Also confirmed `commands/curation`'s existing `TestPipWinsOverStrayUvLock`, `TestPureUvProjectNotReportedAsPip`, and `TestTechsToAuditQueuesPep723HintForDeferredLogging` all still pass unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Python project detection so UV-managed projects are identified as UV instead of Pip when both indicators are present.
  - Prevented duplicate scanning of the same project as both Pip and UV.
  - Kept Pip detection when only Pip-specific files are present.
  - Ignored global UV configuration when determining a project’s technology.
  - Applied consistent Pip/UV detection across audit, scan, and curation workflows.

### DIFF
--- a/utils/techutils/techutils.go
+++ b/utils/techutils/techutils.go
@@ -225,7 +225,7 @@ var technologiesData = map[Technology]TechData{
 		indicators:         []string{"pyproject.toml", "setup.py", "requirements.txt"},
 		validators:         map[string]ContentValidator{"pyproject.toml": pyProjectTomlIndicatorContent(Pip)},
 		packageDescriptors: []string{"setup.py", "requirements.txt", "pyproject.toml"},
-		exclude:            []string{"Pipfile", "Pipfile.lock", "poetry.lock"},
+		exclude:            []string{"Pipfile", "Pipfile.lock", "poetry.lock", "uv.lock"},
 		projectType:        project.Pip,
 		language:           Python,
 	},


commands/audit/audit.go
```diff
@@ -678,7 +678,11 @@ func detectTechnologiesInTarget(target results.ScanTarget, otherParams *AuditPar
 			log.Warn(fmt.Sprintf("Couldn't detect technologies in '%s' directory: %s", included, err.Error()))
 			continue
 		}
+		dirTechs := make([]techutils.Technology, 0, len(techToWorkingDirs))
 		for tech := range techToWorkingDirs {
+			dirTechs = append(dirTechs, tech)
+		}
+		for _, tech := range techutils.PromotePipToUv(dirTechs, included) {
```
attiasas: PromotePipToUv is called with `included` (the scan/include root), after unioning every tech from recursive DetectTechnologiesDescriptors.

For a default new-flow target, `included` is the repo root and recursion is on (`included == target.Target`). A uv tree under e.g. `pkg/python/` (pyproject.toml + uv.lock) still contributes both Pip and Uv, while the root has neither a pip-exclusive file nor uv.lock — so promotion is a no-op and the original dual-tech → NoTech bug remains for that target.

Can we run PromotePipToUv per inner working directory in `techToWorkingDirs` (same granularity as filterAmbiguousPipUvTargets), instead of once against the aggregate root?

utils/techutils/techutils.go
```diff
@@ -876,6 +876,60 @@ func promoteYarnWorkspaceMembers(technologiesDetected map[Technology]map[string]
...
+	if _, statErr := os.Stat(filepath.Join(dir, "uv.lock")); statErr == nil {
+		uvSignal = "uv.lock detected"
+	} else if data, readErr := os.ReadFile(filepath.Join(dir, "pyproject.toml")); readErr == nil &&
+		(strings.Contains(string(data), "[tool.uv]") || strings.Contains(string(data), "[[tool.uv.index]]")) {
+		uvSignal = "pyproject.toml has uv configuration ([tool.uv] or [[tool.uv.index]])"
+	} else if home, homeErr := os.UserHomeDir(); homeErr == nil {
+		if _, statErr := os.Stat(filepath.Join(home, ".config", "uv", "uv.toml")); statErr == nil {
+			uvSignal = "~/.config/uv/uv.toml detected"
+		}
+	}
```
attiasas: `~/.config/uv/uv.toml` is a machine-global signal, but this is now on the shared detector used by jf audit / jf scan.

The new flow always calls PromotePipToUv whenever Pip is in the candidate list, even if Uv was never detected. Any pip-only PEP 621 project (pyproject.toml, no requirements.txt / setup.py / …) on a machine with uv installed will be rewritten to Uv and scanned with the wrong manager.

The old flow mostly avoids this: filterAmbiguousPipUvTargets returns immediately unless both techs exist, and Uv’s detector indicator is uv.lock, so this global file is almost never consulted. Expanding curation’s heuristic to every audit target looks like a regression vs previous audit behavior.

Can we drop the global uv.toml check from shared detection (keep project-local uv.lock / [tool.uv]), or at least not apply it unless Uv was already detected in *that* dir?

commands/audit/audit.go
```diff
@@ -703,6 +707,38 @@ func matchCentralConfigModulesForOldFlow(cmdResults *results.SecurityCommandResu
 	}
 }
 
+// filterAmbiguousPipUvTargets drops, per working directory, whichever of Pip/Uv
+// techutils.PromotePipToUv rejects for that directory - the same ambiguity
+// detectTechnologiesInTarget resolves for the new BOM-generator flow. Without this, a
+// uv-managed working directory that also matches Pip's generic pyproject.toml indicator
+// would be scanned twice, once mislabeled.
+func filterAmbiguousPipUvTargets(techToWorkingDirs map[techutils.Technology]map[string][]string) map[techutils.Technology]map[string][]string {
```
attiasas, line 733: There are no tests for PromotePipToUv in techutils, filterAmbiguousPipUvTargets, or detectTechnologiesInTarget.

Curation tests cover cwd + os.Getwd() wrapping, not explicit `dir`, nested WDs, in-place map deletes, or the new-flow union. The audit/Frogbot wiring is the reason this PR exists.

Could we add:
- temp-dir unit tests on techutils.PromotePipToUv(techs, dir) (no Chdir)
- table tests for this filter (overlap, pip-only, uv-only, empty after delete)
- one nested-root case for detectTechnologiesInTarget

utils/techutils/techutils.go
```diff
@@ -876,6 +876,60 @@ func promoteYarnWorkspaceMembers(technologiesDetected map[Technology]map[string]
...
+			return removeTechnology(techs, Uv)
+		}
+	}
+
+	uvSignal := ""
+	if _, statErr := os.Stat(filepath.Join(dir, "uv.lock")); statErr == nil {
+		uvSignal = "uv.lock detected"
+	} else if data, readErr := os.ReadFile(filepath.Join(dir, "pyproject.toml")); readErr == nil &&
+		(strings.Contains(string(data), "[tool.uv]") || strings.Contains(string(data), "[[tool.uv.index]]")) {
+		uvSignal = "pyproject.toml has uv configuration ([tool.uv] or [[tool.uv.index]])"
```
attiasas: Nit: strings.Contains for `[tool.uv]` / `[[tool.uv.index]]` is coarser than Poetry’s `^\[tool\.poetry\]` regex — comments/strings can false-positive, and `[tool.uv.sources]` without a bare `[tool.uv]` table can false-negative. Not a blocker; a line-anchored TOML table match would match pyProjectTomlIndicatorContent.

utils/techutils/techutils.go
```diff
@@ -876,6 +876,60 @@ func promoteYarnWorkspaceMembers(technologiesDetected map[Technology]map[string]
...
+	} else if home, homeErr := os.UserHomeDir(); homeErr == nil {
+		if _, statErr := os.Stat(filepath.Join(home, ".config", "uv", "uv.toml")); statErr == nil {
+			uvSignal = "~/.config/uv/uv.toml detected"
+		}
+	}
+	if uvSignal == "" {
+		return techs
+	}
+	log.Info(uvSignal + " — treating project as uv.")
```
attiasas: Nit: log.Info on every promotion will be noisy for multi-module audits. Yarn’s similar detector promotion uses Debug; Debug (or once per target) would fit better here.

commands/curation/curationaudit.go
```diff
@@ -709,45 +709,19 @@ func promoteYarnWorkspaceMember(techs []string) []string {
 	}
 }
 
-// Rule, in order:
-//  1. Pip-exclusive file present (requirements.txt, setup.py, setup.cfg, Pipfile,
-//     poetry.lock) → Pip wins, Uv is dropped.
```
attiasas, line 714: Nit: the wrapper still uses os.Getwd(), not ca.OriginPath. Behavior matches the old function and existing tests, but the new “pass dir explicitly” contract is only honored after Getwd. Fine for jf ca as tested; just inconsistent with multi-target audit.

attiasas (commented):
Nice job, check out my comments

commands/curation/curationaudit.go
```diff
@@ -709,45 +709,26 @@ func promoteYarnWorkspaceMember(techs []string) []string {
...
-	dir, err := os.Getwd()
-	if err != nil {
-		return techs
-	}
-	for _, pipOnlyFile := range []string{"requirements.txt", "setup.py", "setup.cfg", "Pipfile", "poetry.lock"} {
-		if _, statErr := os.Stat(filepath.Join(dir, pipOnlyFile)); statErr == nil {
-			return removeTech(techs, techutils.Uv.String())
-		}
-	}
+	return promotePipToUvIn(techs, "")
```
attiasas: Do we still need `promotePipToUv` here? 

Nit (same as before, still accurate): the wrapper still `Getwd()`s when `dir == ""`. Production `techsToAudit` passes `ca.OriginPath`, so `jf ca` is fine.

Easy to reuse later and silently scan the process cwd. Point tests at `promotePipToUvIn(techs, projectDir)` and drop the empty-dir `Getwd` path, or keep this wrapper test-only.

commands/audit/audit.go
```diff
@@ -703,6 +725,38 @@ func matchCentralConfigModulesForOldFlow(cmdResults *results.SecurityCommandResu
...
+	pipDirs, hasPip := techToWorkingDirs[techutils.Pip]
+	uvDirs, hasUv := techToWorkingDirs[techutils.Uv]
+	if !hasPip || !hasUv {
```
attiasas: Nit / follow-up: this early return means the old/Frogbot path only promotes when **both** Pip and Uv are already in the map. Uv's detector indicator is `uv.lock`, so `pyproject.toml` + `[tool.uv]` with no lockfile stays Pip here.

The new flow (`technologiesAfterPipUvPromotion`) always calls `PromotePipToUv` per dir and can rewrite that case to Uv. Not the original same-dir `[pip, uv]` collapse, but the comment above says both flows resolve the same rule.

If we want them identical: for each Pip working dir, call `PromotePipToUv` even when Uv is absent, then add/remove map entries from the result (including creating a Uv entry). Fine as a follow-up.

commands/audit/audit.go
```diff
@@ -678,13 +678,35 @@ func detectTechnologiesInTarget(target results.ScanTarget, otherParams *AuditPar
 			log.Warn(fmt.Sprintf("Couldn't detect technologies in '%s' directory: %s", included, err.Error()))
 			continue
 		}
-		for tech := range techToWorkingDirs {
+		for _, tech := range technologiesAfterPipUvPromotion(techToWorkingDirs) {
```
attiasas, line 681: Nit / pre-existing: after per-dir promotion, techs are still unioned onto a single new-flow target. A repo with a pip tree and a **different** uv tree still becomes `[pip, uv]`, and `ResolveIssueTechnology` can still return `NoTech` for PyPI.

This PR correctly fixes same-dir uv. Mixed Python monorepos on `XrayLibBomGenerator` would need split targets (or one Python manager per target) — out of this slice.

attiasas (approved):
LGTM, check out my last comments

PR #883, opened 2026-09-15
Add uv fix support to the Python package updater

orto17:

-----

Stacks on top of [jfrog/jfrog-cli-security#880](https://github.com/jfrog/jfrog-cli-security/pull/880) (Pip/Uv detection ambiguity fix) — uv is now detected correctly, but there was no fix implementation for it: GetCompatiblePackageUpdater had no case for techutils.Uv, so any fix attempt hit the unsupported-technology path.

Adds a handleUv method to PythonPackageUpdater that rewrites the pinned dependency directly in pyproject.toml — wherever it appears (main dependencies, optional extras, or a PEP 735 dependency-groups table) — then runs uv lock --upgrade-package <name> to refresh uv.lock.

### Why not uv add ###

The obvious approach, mirroring Poetry's handler (poetry add pkg==version + poetry update), doesn't generalize: uv add pkg==version requires an explicit --group/--optional flag to target anything outside the main dependency list, and hard-fails with a resolver conflict if you omit it while the package lives in a dependency-groups table. There's no evidence data available (from Xray or the SBOM) telling us which group a dependency came from, so that flag can never be chosen reliably.

Instead this reuses the same mechanism the existing Pip handler already uses (handlePip): a text substitution of the exact pinned line, regardless of which manifest table it's in, followed by a lock refresh. It's also simpler than Poetry's handler — uv add already keeps pyproject.toml and uv.lock in sync in one command, so the equivalent here is one substitution + one scoped lock refresh, no two-step CLI dance.

### Verification ###

Added test fixtures under tests/testdata/projects/package-managers/uv/ (plain project, indirect-dependency, dependency-groups, and a two-package workspace) and four new subtests in TestUpdateDependency:

Indirect dependency → ErrUnsupportedFix, matching pip/poetry/pipenv.
Plain direct dependency → pyjwt==1.7.1 becomes 2.4.0 in pyproject.toml; uv.lock refreshed.
Dependency-groups ([dependency-groups].dev) → same fix succeeds, no --group flag needed, unlike a naive uv add.
Workspace (dependency owned by a member, not the root) → fails safely with "impacted package not found, fix failed", the same error Pip's handler already returns for an unmatched package; root's pyproject.toml is left untouched.
### Limitations ###
* Workspaces aren't fixed, only safely rejected. If the vulnerable dependency is declared in a workspace member's pyproject.toml rather than the scanned target's own file, this handler can't locate it — it only ever reads/writes the target directory's own manifest. The result is a clean failure (no wrong-file edits, no corruption), not a fix. Solving this needs the vulnerability's evidence to identify which workspace member owns it, which isn't available today.
* Indirect/transitive dependencies are unsupported, same as pip/poetry/pipenv — uv does have a resolver-driven option (uv lock --upgrade-package without editing the manifest first) that pip fundamentally lacks, but that's a possible future enhancement, not attempted here, to keep behavior consistent with the other Python managers.
* Requires the uv binary and network access wherever the fix runs (CI/build environment), same precondition Poetry and Pipenv already have.
* Static, single-pin assumption: like handlePip, this assumes the impacted version is statically pinned with a simple operator (==, >=, etc.) directly in the manifest — it doesn't handle version ranges spanning multiple constraints in unusual ways, dynamic versions, or a dependency declared via path/git/url sources instead of a version pin.


remediation/sca/packageupdaters/pythonpackageupdater.go
```diff
@@ -77,6 +79,28 @@ func (py *PythonPackageUpdater) handlePip(fixDetails *FixDetails) (err error) {
 	return
 }
 
+func (py *PythonPackageUpdater) handleUv(fixDetails *FixDetails) (err error) {
```
Jordanh1996, line 87: other package-updaters have "rollback" for manifest files. I see here you didn't include?

remediation/sca/packageupdaters/pythonpackageupdater.go
```diff
@@ -77,6 +79,28 @@ func (py *PythonPackageUpdater) handlePip(fixDetails *FixDetails) (err error) {
...
+	const pyprojectFile = "pyproject.toml"
+	var fixedFile string
+	fixedPackage := fixDetails.ImpactedDependencyName + "==" + fixDetails.SuggestedFixedVersion
+	currentFile, err := py.tryReadRequirementFile(pyprojectFile)
+	if err != nil {
+		return errors.New("failed to read pyproject.toml: " + err.Error())
+	}
+	re := regexp.MustCompile(PythonPackageRegexPrefix + "(" + fixDetails.ImpactedDependencyName + "|" + strings.ToLower(fixDetails.ImpactedDependencyName) + ")" + PythonPackageRegexSuffix)
```
Jordanh1996: regex not strong enough which might cause "fixing" the wrong package

`The package name is interpolated into the regex raw — no regexp.QuoteMeta, no left-hand boundary — so it also matches as a suffix of a longer dependency name: with dependencies = ["cattrs==23.2.3", "attrs==23.2.0"] and a fix of attrs→24.1.0, handleUv returns nil after rewriting cattrs to 24.1.0 and leaving the vulnerable attrs==23.2.0 untouched in both pyproject.toml and uv.lock. Because it succeeds silently, Frogbot opens a PR that bumps an unrelated package and reports the CVE as fixed — CommonPackageUpdater.BuildPackageDependencyLineRegex (commonpackageupdater.go:108) already does the QuoteMeta, so reusing it plus a (^|[\s"'\[,]) prefix closes both halves.`

reproduction:
dependencies = ["cattrs==23.2.3", "attrs==23.2.0"]   fix: attrs → 24.1.0
→ handleUv returns nil
→ pyproject.toml: "cattrs==24.1.0", "attrs==23.2.0"
→ uv.lock:        cattrs 24.1.0,    attrs 23.2.0

remediation/sca/packageupdaters/pythonpackageupdater.go
```diff
@@ -77,6 +79,28 @@ func (py *PythonPackageUpdater) handlePip(fixDetails *FixDetails) (err error) {
...
+	if packageToReplace := re.FindString(currentFile); packageToReplace != "" {
+		fixedFile = strings.Replace(currentFile, packageToReplace, strings.ToLower(fixedPackage), 1)
```
Jordanh1996: FindString + Replace(..., 1) only ever fixes the first declaration, so a package listed in both [project].dependencies and a [dependency-groups] table is half-edited and uv lock then fails unsatisfiable

example:
dependencies = ["pyjwt==1.7.1"]   [dependency-groups] dev = ["pyjwt==1.7.1"]   fix: pyjwt → 2.4.0
→ pyproject.toml: dependencies "pyjwt==2.4.0", dev "pyjwt==1.7.1"
→ uv lock fails: pyjwt==2.4.0 vs demo:dev pyjwt==1.7.1 unsatisfiable
→ no uv.lock, manifest left half-edited

utils/techutils/techutils.go
```diff
@@ -253,14 +253,16 @@ var technologiesData = map[Technology]TechData{
...
-		projectType:        project.UV,
-		language:           Python,
+		formal:                     "uv",
+		packageType:                Pypi,
+		xrayPackageType:            Pypi,
+		indicators:                 []string{"uv.lock"},
+		packageDescriptors:         []string{"pyproject.toml"},
+		execCommand:                "uv",
+		packageInstallationCommand: "add",
+		packageVersionOperator:     "==",
```
Jordanh1996: `packageInstallationCommand` and `packageVersionOperator` are unused

Jordanh1996 (changes requested):
From frogbot side - don't forget to whitelist yarn and uv(and dotnet once its ready)

remediation/sca/packageupdaters/pythonpackageupdater.go
orto17: if uv lock --upgrade-package fails after the manifest edit, pyproject.toml is now restored to its original content — matching the existing convention in npm/pnpm/yarn/go's updaters.

remediation/sca/packageupdaters/pythonpackageupdater.go
orto17: switched from FindString + Replace(..., 1) to ReplaceAllString, so a package declared in both [project].dependencies and a [dependency-groups] table gets updated in both places instead of being left half-edited and unsatisfiable.

utils/techutils/techutils.go
orto17: removed

Jordanh1996 (approved):

PR #888, opened 2026-09-17
Nuget Package Updater - CPM Support

orto17:

-----

Adds Central Package Management (CPM) support to the NuGet fix updater, on top of the plain-PackageReference support merged in [jfrog/jfrog-cli-security#874](https://github.com/jfrog/jfrog-cli-security/pull/874). Goal: keep "fixable by our updater" equal to "detected by the scanner" for NuGet — if Xray can find a version string to bump, this updater should be able to bump it.

* Directory.Packages.props resolution — when a PackageReference has no inline version, walks up from the project directory toward the repo root for the nearest Directory.Packages.props (NuGet's own "closest wins" resolution), and patches the matching <PackageVersion> entry there instead of the evidenced file. Handles nested/multiple CPM files, <Import Project="..."> chains (including the $(MSBuildThisFileDirectory) macro), and explicit <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally> opt-outs.
* VersionOverride support — a local VersionOverride attribute or child element wins over CPM entirely, since that's how NuGet itself resolves precedence.
* Directory.Build.props / .targets fallback — packages injected purely via a repo-wide auto-imported props/targets file (no local PackageReference at all) are now also reachable, tried after CPM comes back unresolved.
* Lock-file evidence resolution — Xray's evidence for a CPM- or Directory.Build.props-resolved package only ever cites packages.lock.json (the version text never appears in the .csproj at all). packages.lock.json is now recognized as a NuGet package descriptor, and its evidence is resolved back to the sibling project file in the same directory.
* Mixed inline/non-inline reference guard — a project with two PackageReference entries for the same package, one inline-versioned and one not, is now treated as unsupported rather than silently falling through to CPM (where it doesn't belong).
* Evidence preference fix (utils/results/common.go) — CdxEvidencesToPreferredLocation now prefers a real descriptor (.csproj) over a lock file when both are present, instead of picking whichever occurrence happened to come first.
### Out of scope ###
* CPM/Directory.Build.props fixes still require a lock file to safely regenerate — same limitation as the rest of the NuGet updater.
* Sibling projects that share the same governing Directory.Packages.props but weren't themselves evidenced are not proactively restored; their lock files can go stale relative to the bumped central version until they're separately evidenced and fixed. (In practice this is narrower than it sounds — Xray merges same-name/same-version SBOM components' evidence across projects, so most genuinely-affected siblings are already evidenced on their own.)


remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -258,3 +432,181 @@ func updatePackageReferenceVersion(content []byte, packageName, fixedVersion str
 	}
 	return nil, fmt.Errorf("dependency %s not found", packageName)
 }
+
+func updatePackageVersionEntry(content []byte, packageName, fixedVersion string) ([]byte, bool) {
```
Jordanh1996, line 424: This matches <PackageVersion> elements with no XML-comment awareness, and searchVersionUpdate returns as soon as fixedAny is true (line 534), so a commented-out pin both gets rewritten and ends the <Import> descent before the real entry is reached.

see this reproduction:
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/pull/1/changes
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/tree/frogbot-Newtonsoft.Json-f94909a31152380ca531fb055d6189b6

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -240,12 +401,25 @@ func updatePackageReferenceVersion(content []byte, packageName, fixedVersion str
...
+		case versionOverrideElement.Match(match):
+			fixedAny = true
+			return versionOverrideElement.ReplaceAll(match, []byte("${1}"+fixedVersion+"${2}"))
 		default:
 			foundWithoutVersion = true
 			return match
 		}
 	})
 
+	if foundWithoutVersion && fixedAny {
```
Jordanh1996: A commented-out <PackageReference ... Version="..."/> sets fixedAny, and a <PackageReference Update="..."/> metadata row sets foundWithoutVersion, so both get classified as "mixed". fixVulnerabilityAndRestore (line 181) returns on this error type with no fallback, so the fix is abandoned even when it's unambiguous — in the second case the project isn't using CPM at all.

see these reproductions (Frogbot logged skipping fix of vulnerable package 'Newtonsoft.Json' version '13.0.1' - package has both inline and versionless PackageReference entries for both, no fix PR):
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/tree/s2-mixed
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/tree/s6-metadata

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -258,3 +432,181 @@ func updatePackageReferenceVersion(content []byte, packageName, fixedVersion str
...
+		if strings.HasPrefix(projectRef, nugetMSBuildThisFileDirectoryMacro) {
+			suffix := filepath.FromSlash(strings.TrimPrefix(projectRef, nugetMSBuildThisFileDirectoryMacro))
+			projectRef = filepath.Join(filepath.Dir(absPath), suffix)
+		}
+		if strings.Contains(projectRef, "$") {
+			continue
+		}
+		nextPath := projectRef
+		if !filepath.IsAbs(projectRef) {
+			nextPath = filepath.Join(filepath.Dir(absPath), filepath.FromSlash(projectRef))
```
Jordanh1996: nuget projects might include "\"(forward slash) which is the convention on windows. which would break in the current implementation.

relevant in both lines 550 and 542.

AI:
--
filepath.FromSlash is a no-op outside Windows, so <Import Project="Shared\Packages.props"/> stays one filename component, os.ReadFile gets ENOENT, and the import is skipped. Both call sites have it — line 542 breaks $(MSBuildThisFileDirectory)Shared\Packages.props, line 550 breaks the plain relative form. MSBuild treats \ as a separator on every platform and resolves both fine on macOS/Linux, where Frogbot runs.

A helper used at both sites covers it:
```go
func msbuildPathToFilePath(msbuildPath string) string {
    return filepath.FromSlash(strings.ReplaceAll(msbuildPath, `\`, "/"))
}
```

see these reproductions — before, no fix PR and could not resolve a version to fix (no inline attribute, VersionOverride, or matching Directory.Packages.props / Directory.Build.props entry found):
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/tree/s3-backslash
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/tree/s7-macro

and after applying the helper, both fix correctly:
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/pull/4/changes
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/pull/5/changes

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -27,8 +28,20 @@ const (
...
+	nugetVersionOverrideAttrPattern    = `(?is)(\bVersionOverride\s*=\s*["'])[^"']*(["'])`
+	nugetVersionOverrideElementPattern = `(?is)(<VersionOverride>)[^<]*(</VersionOverride>)`
+
+	nugetPackageVersionElementPattern  = `(?is)<PackageVersion\b[^>]*/>|<PackageVersion\b[^>]*[^/]>.*?</PackageVersion>`
+	nugetPackageVersionKeyAttrPattern  = `(?i)\b(?:Include|Update)\s*=\s*["']%s["']`
+	nugetDirectoryPackagesPropsName    = "Directory.Packages.props"
+	nugetDirectoryBuildPropsName       = "Directory.Build.props"
+	nugetDirectoryBuildTargetsName     = "Directory.Build.targets"
+	nugetImportProjectAttrPattern      = `(?i)<Import\b[^>]*\bProject\s*=\s*["']([^"']+)["']`
+	nugetManageCpmFalsePattern         = `(?is)<ManagePackageVersionsCentrally>\s*false\s*</ManagePackageVersionsCentrally>`
```
Jordanh1996: The regex only matches a bare <ManagePackageVersionsCentrally>. Add an attribute and it stops matching — so a repo that turns CPM off conditionally, <ManagePackageVersionsCentrally Condition="...">false</...>, is read as CPM on.

We then bump <PackageVersion> in Directory.Packages.props. With CPM off, that element does nothing. The version that actually applies sits in Directory.Build.props and stays vulnerable. Frogbot opens a PR that changes nothing.

Fix: allow attributes on the tag — \b[^>]*>.

see this reproduction — the fix branch still resolves 9.0.1 and still warns NU1903:
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/pull/2/changes
- https://github.com/Jordanh1996/frogbot-nuget-cpm-repro/tree/s4-condition

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -27,8 +28,20 @@ const (
...
 	// it; Maven's updater handles the analogous ${property} case by updating the definition instead.
-	nugetVersionAttrPattern    = `(?is)(\bVersion\s*=\s*["'])[^"']*(["'])`
-	nugetVersionElementPattern = `(?is)(<Version>)[^<]*(</Version>)`
+	nugetVersionAttrPattern            = `(?is)(\bVersion\s*=\s*["'])[^"']*(["'])`
+	nugetVersionElementPattern         = `(?is)(<Version>)[^<]*(</Version>)`
+	nugetVersionOverrideAttrPattern    = `(?is)(\bVersionOverride\s*=\s*["'])[^"']*(["'])`
+	nugetVersionOverrideElementPattern = `(?is)(<VersionOverride>)[^<]*(</VersionOverride>)`
+
+	nugetPackageVersionElementPattern  = `(?is)<PackageVersion\b[^>]*/>|<PackageVersion\b[^>]*[^/]>.*?</PackageVersion>`
+	nugetPackageVersionKeyAttrPattern  = `(?i)\b(?:Include|Update)\s*=\s*["']%s["']`
```
Jordanh1996: identical to the other constant `nugetKeyAttrPattern`

remediation/sca/packageupdaters/nugetpackageupdater.go
```diff
@@ -258,3 +432,181 @@ func updatePackageReferenceVersion(content []byte, packageName, fixedVersion str
 	}
 	return nil, fmt.Errorf("dependency %s not found", packageName)
 }
+
+func updatePackageVersionEntry(content []byte, packageName, fixedVersion string) ([]byte, bool) {
+	element := regexp.MustCompile(nugetPackageVersionElementPattern)
+	keyAttr := regexp.MustCompile(fmt.Sprintf(nugetPackageVersionKeyAttrPattern, regexp.QuoteMeta(packageName)))
+	versionAttr := regexp.MustCompile(nugetVersionAttrPattern)
+	versionElement := regexp.MustCompile(nugetVersionElementPattern)
```
Jordanh1996: 11 regexp.MustCompile calls, all at call time, and only 2 depend on the package name. The rest are constant and should be package-level vars

Jordanh1996 (changes requested):

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: added maskXMLComments/replaceMatchingElements, used by both updatePackageReferenceVersion and updatePackageVersionEntry, so a commented-out <PackageVersion>/<PackageReference> is never mistaken for a live one, and searchVersionUpdate no longer stops descending an Import chain because of it. New test: TestNugetUpdateDependencyIgnoresCommentedOutPackageVersion.

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: Backslash-separated Import paths — added msbuildPathToFilePath (converts \→/ before filepath.FromSlash) and used it at both call sites in searchVersionUpdate. New tests: TestNugetUpdateDependencyFollowsBackslashSeparatedImport, TestNugetUpdateDependencyFollowsMSBuildThisFileDirectoryImportWithBackslashSubpath.

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: the regexes now allow \b[^>]*> on the opening tag instead of requiring a bare tag, so a conditional false/true is read correctly instead of defaulting to "on". New test: TestNugetUpdateDependencyRecognizesManagePackageVersionsCentrallyFalseWithAttributes.

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: removed

remediation/sca/packageupdaters/nugetpackageupdater.go
orto17: all 9 constant patterns are now package-level regexp.MustCompile vars instead of being recompiled on every call

Jordanh1996 (approved):

